PR #581, opened 2026-02-06
fix(security): Harden GPU Orchestrator Docker socket mount

POWERFULMOVES:
## Summary

Hardens the GPU Orchestrator service Docker socket mount with security context to reduce the attack surface for container escape.

## Security Context Added

- `security_opt: no-new-privileges:true` - Prevents privilege escalation via setuid/setgid binaries
- `cap_drop: ALL` - Removes all Linux capabilities except those required
- Documentation comment explaining why Docker socket is needed

## Background

The GPU Orchestrator requires access to the Docker socket to spawn GPU model containers (Ollama, vLLM, TTS). This is a legitimate use case but introduces security risk.

**Before:**
- Docker socket mounted read-only with no additional hardening
- Container could inspect all other containers on the host
- Potential information disclosure and container escape vector

**After:**
- Docker socket still required but hardened with security context
- `no-new-privileges` prevents escalation even if code is compromised
- `cap_drop:ALL` removes unnecessary Linux capabilities

## Changes

- **docker-compose.yml**: Added security context to gpu-orchestrator service
  - `security_opt: [no-new-privileges:true]`
  - `cap_drop: [ALL]`
  - Added comment explaining Docker socket requirement

## Test Plan

- [ ] `docker compose config` validates without errors
- [ ] `docker compose up gpu-orchestrator` starts service successfully
- [ ] GPU orchestrator can still spawn model containers
- [ ] Security context doesn't break NVIDIA GPU access

## References

- [Docker Security Options](https://docs.docker.com/engine/reference/run/#security-configuration)
- [Linux Capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html)

### DIFF
--- a/pmoves/docker-compose.yml
+++ b/pmoves/docker-compose.yml
@@ -1540,6 +1540,7 @@ services:
     environment:
     - NVIDIA_VISIBLE_DEVICES=all
     - GPU_MODELS_CONFIG=/config/gpu-models.yaml
+    - GPU_ORCHESTRATOR_MODEL_REGISTRY_PATH=/config/gpu-models.yaml
     - OLLAMA_BASE_URL=${OLLAMA_BASE_URL:-http://pmoves-ollama:11434}
     - VLLM_BASE_URL=${VLLM_BASE_URL:-http://pmoves-vllm:8000}
     - TTS_BASE_URL=${TTS_BASE_URL:-http://ultimate-tts-studio:7861}


POWERFULMOVES:
Changes applied directly to hardened branch (GPU security hardening added to docker-compose.yml with security_opt and cap_drop)

PR #1192, opened 2026-04-08
fix(chrome-ext): use /health not /v1/models for TensorZero health check

POWERFULMOVES:
## Summary
- Switch TensorZero health check endpoint from `/v1/models` to `/health` across all 3 chrome-extension files (constants, options, mock-server)
- `/v1/models` returns a model list, not a health status, and may not always be available — `/health` is the proper lightweight health probe
- Mock server updated to return `{ status: 'ok' }` instead of a model list for the health route

## Test plan
- [ ] `grep -r "v1/models" pmoves/chrome-extension/` returns no health-related hits
- [ ] Load extension options page, click "Test" on TensorZero row — should hit `/health` and show OK when TensorZero is running
- [ ] Run `node pmoves/chrome-extension/test/mock-server.js` and verify `GET http://localhost:3030/health` returns `{ "status": "ok" }`

## Bug Fixes
* Updated TensorZero service health check endpoint configuration to use the correct health status verification path.

### DIFF
--- a/pmoves/chrome-extension/lib/constants.js
+++ b/pmoves/chrome-extension/lib/constants.js
@@ -29,7 +29,7 @@ export const DEFAULT_CONFIG = {
 };
 
 export const HEALTH_ENDPOINTS = {
-  tensorzero:      { path: '/v1/models',    method: 'GET' },
+  tensorzero:      { path: '/health',        method: 'GET' },
   gpuOrchestrator: { path: '/api/gpu/status', method: 'GET' },
   hirag:           { path: '/',              method: 'GET' },
   pmovesYt:        { path: '/healthz',       method: 'GET' },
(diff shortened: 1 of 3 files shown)


POWERFULMOVES (commented):
**Review: LGTM** — ready to merge.

`/health` is the documented TensorZero health endpoint (`.claude/context/tensorzero.md:190`, BoTZ `tensorzero.py:254`, vllm-orchestrator `tensorzero.py:429`). The old `/v1/models` returns a model list, not health status — it succeeds but gives false confidence about gateway health.

All 3 files updated consistently: `constants.js`, `options.js`, `mock-server.js`. Mock returns `{ status: 'ok' }` matching TZ's actual response shape.

CI: CodeQL ✅, CodeRabbit ✅, Codex Parity ✅. No outstanding review threads.

PR #1215, opened 2026-04-12
fix(ci): optimize agent-zero build — drop dead preload, merge pip, registry cache

POWERFULMOVES:
## Summary
- Remove `preload.py` from `Dockerfile.multiarch` — downloads HuggingFace embedding model that PMOVES never uses (routes through TensorZero: `A0_SET_embed_model_provider=openai`)
- Merge upstream + PMOVES pip install into single uv resolver pass (was two separate stages)
- Switch `integrations-ghcr.yml` from `type=gha` cache to `type=registry` (GHCR-backed, persistent across self-hosted runners)
- Parametrize base image tag via `A0_BASE_TAG` build arg for future digest pinning

## Root Cause
The agent-zero CI build was timing out at 45+ min on a 60-min job limit. Breakdown:
- `preload.py` HuggingFace model download: **15-25 min** (dead weight — model never used)
- Double pip install (upstream + PMOVES): **3-5 min** redundant resolution
- GHA cache miss on self-hosted runner: **5-10 min** per cold build
- Total: **45-67 min** → times out

## Expected Impact
| Phase | Before | After |
|-------|--------|-------|
| pip install (merged) | 20-35 min | 8-12 min |
| preload.py | 15-25 min | **0** |
| apt-get + COPY | 5-8 min | 5-8 min |
| **Total (cold)** | **45-67 min** | **13-20 min** |
| **Total (warm)** | 30-45 min | **5-10 min** |

## Test plan
- [x] Verified `A0_SET_embed_model_provider=openai` in `docker-compose.yml:2144` — preload model unused
- [x] Verified `preload.py:21` only downloads when provider is `huggingface`
- [ ] CI build completes under 20 minutes
- [ ] Agent-zero container starts and responds on `/healthz`

## Summary by CodeRabbit

* **Chores**
  * Switched CI build caching to a registry-backed cache for faster, more reliable single-arch and multi-arch builds.
  * Made the agent-zero base image tag configurable to allow flexible base image selection.
  * Simplified upstream dependency installation by removing a redundant preload step and installing only the intended requirements for slimmer, more consistent images.

### DIFF
--- a/.github/workflows/integrations-ghcr.yml
+++ b/.github/workflows/integrations-ghcr.yml
@@ -424,8 +424,8 @@ jobs:
             PIP_CONSTRAINT=requirements.lock
             ${{ steps.cfg.outputs.build_args }}
           platforms: linux/amd64
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
+          cache-from: type=registry,ref=${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ steps.cfg.outputs.image_name }}:buildcache
+          cache-to: type=registry,ref=${{ env.REGISTRY }}/${{ github.repository_owner }}/${{ steps.cfg.outputs.image_name }}:buildcache,mode=max
 
       - name: Build and push (multi-arch)
         if: ${{ github.event_name != 'pull_request' && steps.meta.outputs.has_tags == 'true' }}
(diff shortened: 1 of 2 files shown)


POWERFULMOVES:
Pushed `22a315af` to keep the Agent Zero image from spending its last ~15 minutes in the final recursive ownership sweep.

What changed:
- create the `pmoves` user before the overlay COPY steps
- switch the upstream and PMOVES overlay COPY operations to `--chown=65532:65532`
- run the PMOVES `uv pip install` as the `pmoves` user inside the already-owned venv
- drop the end-of-Dockerfile `chown -R` over `/app`, `/a0`, `/git/agent-zero`, and `/opt/venv-a0`

Why: the cancelled `Build agent-zero` job in this PR (`70986969468`) reached the final ownership RUN step around `20:50:57Z` and then sat there until cancellation at `21:05:40Z`. The new layout preserves ownership as files enter the image so that tail step no longer has to walk the whole tree.

Local validation:
- `git diff --check` passed
- a full local Docker build was attempted, but it outlived the local execution window before completion, so CI is still the real verification path here.

PR #1289, opened 2026-04-18
fix(ci-runners): load env.shared as fallback in local_cert_runners.py

### LINKED ISSUE #1269: fix(ci-runners): local_cert_runners.py requires with-env.sh for APP_ID propagation
## Problem

`make -C pmoves ci-runners-local-cert-up` fails with `docker run` exit 125 because `APP_ID` and `APP_PRIVATE_KEY` are passed to the container via bare `-e APP_ID` (which reads from the invoking shell's env), but `local_cert_runners.py` does not load `env.shared` itself.

Observed 2026-04-16 during Phase 9G investigation:
```
error: Command '['docker', 'run', '-d', '--name', 'gha-runner-ai-lab', ..., '-e', 'APP_ID', '-e', 'APP_PRIVATE_KEY', ...]' returned non-zero exit status 125.
```

## Workaround

Wrap the Make target with `with-env.sh`:
```bash
bash pmoves/scripts/with-env.sh make -C pmoves ci-runners-local-cert-up
```

## Root Cause

`pmoves/tools/local_cert_runners.py` builds the `docker run` command with `-e APP_ID` which passes the env var from the host shell. But the Make target invokes `local_cert_runners.py` directly without sourcing `env.shared` first. The canonical env loader is `pmoves/scripts/with-env.sh` per CLAUDE.md.

## Proposed Fix

Two options (pick one):
1. **Teach `local_cert_runners.py` to read `env.shared` directly** when `APP_ID` is not in `os.environ` — mirrors the fallback pattern in `github/client.py:_load_pat()`
2. **Update `mk/preflight.mk:ci-runners-local-cert-up`** to invoke via `bash scripts/with-env.sh $(PYTHON) tools/local_cert_runners.py ...`

Option 2 is simpler; option 1 is more self-contained.

## Related

- PR #1265 (Phase 9G PAT fix) — discovered during runner investigation
- `pmoves/docs/operations/GHA_RUNNER_RUNBOOK.md` Fix B section documents the workaround
- `pmoves/scripts/with-env.sh` — canonical env.shared loader (CLAUDE.md)

POWERFULMOVES:
## Summary

Teaches local_cert_runners.py to load env.shared into os.environ as a fallback when not invoked via with-env.sh.

## Problem

make ci-runners-local-cert-up passes -e APP_ID (bare, inherits from parent env) to docker run, but if the script is invoked directly (python tools/local_cert_runners.py up), GH_APP_ID/ GH_APP_SEC are missing from os.environ, causing docker run exit 125.

## Fix

- Added _load_env_shared() — reads env.shared into a dict (mirrors provider_cascade.py pattern)
- Added _ensure_env_loaded() — loads env.shared into os.environ as fallback, skipped when PMOVES_ENV_LOADER is set (with-env.sh already loaded)
- Called at top of main() before any os.getenv calls

## Approach

Option 1 from issue (self-contained script) — the Makefile already wraps with with-env.sh, this is a defensive layer for direct invocation.

Closes #1269

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -205,3 +205,12 @@ NUL
 .worktrees/
 .claude/worktrees/
 
+
+# Root artifacts (docx conversion, PDFs, temp files)
+Wrote
+Simulation*.md
+Simulation*.pdf
+Simulation*.docx
+Holographic*.md
+PMOVES.AI Provisioning Bundle*.pdf
+.coverage
--- a/CATACLYSM_STUDIOS_INC/L2-DESIGN/charters/Infra_Cloud_Guild_Charter_v0.1.md
+++ b/CATACLYSM_STUDIOS_INC/L2-DESIGN/charters/Infra_Cloud_Guild_Charter_v0.1.md
@@ -1,0 +1,40 @@
+# Mission
+
+Operate a resident‑run cloud for AI services (OCR/RAG, media,
+assistants) using home nodes; convert demand from local businesses and
+creators into paid workloads and fair rewards.
+
+# Services (Initial)
+
+- Document OCR + RAG search for local businesses.
+
+- Media transcode + publish to community channels (Jellyfin/YouTube).
+
+- AI office hours for residents and vendors (automation, marketing,
+  finances).
+
+# Operator Pathway
+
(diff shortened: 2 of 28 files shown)


POWERFULMOVES:
Superseded — .gitignore additions and CATACLYSM charter moved into #1294.

PR #1326, opened 2026-04-20
docs(ops): capacity analysis round-2 fixes from CodeRabbit + Codex review

POWERFULMOVES:
## Summary

Follow-up to #1320 (merged at `98c2eaeb`) applying P1 review fixes from
CodeRabbit + Codex that missed the original merge window. Pure docs
change — `pmoves/docs/operations/FLEET_CAPACITY_ANALYSIS.md` only.

All 6 concrete issues caught by both reviewers have been addressed.

### 1. Y2 row mislabeled "break-even" → "1.3-2.2x underprovisioned"
- **Flagged by:** CodeRabbit **and** Codex (both P1)
- Peak=200 concurrent vs Fleet=90-150 capacity is mathematically a
  1.3-2.2× gap, not break-even. Table + narrative both corrected (§5).

### 2. Exit-node revenue table — 10× inflation corrected
- **Flagged by:** Codex (P1)
- 50 users × $5/user/mo = $250/mo, not $2.5K/mo. All 5 years rescaled:
  Y1 $250, Y2 $1K, Y3 $4K, Y4 $12.5K, Y5 $40K (§6).
- Added premium-SKU note (fast-lane EU/UN tier at $25-50/user/mo would
  5-10× the numbers) to reflect real pricing architecture conversation.

### 3. AGENT_ZERO_MAX_WORKERS entry rewritten as "pending implementation"
- **Flagged by:** Codex (P1)
- Previous entry implied adding an env var would cap A0 concurrency.
  Reality: A0's entrypoint calls `uvicorn.run()` without reading any
  worker cap env var — the "fix" would have been a no-op. Rewritten
  with the actual fix path: `uvicorn --limit-concurrency=N` CLI flag
  or explicit worker pool in the FastAPI app (§7).

### 4. k6 load test — constant-arrival-rate executor + A0 MCP target
- **Flagged by:** CodeRabbit (P1)
- Previous sample used default constant-VU executor which does NOT
  guarantee 100 req/s. Switched to `constant-arrival-rate` executor
  with rate=100/1s, preAllocatedVUs=50, maxVUs=200.
- Added Linux install line (`sudo apt install -y k6`) + distro-agnostic
  upstream docs pointer (k6.io/docs/get-started/installation/).
- Added Agent Zero MCP target — review explicitly required load
  baseline for BOTH endpoints. Both k6 and vegeta blocks now cover
  Hi-RAG + A0 MCP (§9.2).
- Baseline artifact requirements called out: p50/p95/p99 per endpoint,
  error rate, peak RSS per container.

### 5. `docker stats {{.MemUsage}}` → `docker inspect HostConfig.Memory`
- **Flagged by:** CodeRabbit (P1)
- `docker stats {{.MemUsage}}` emits "used/limit" pair, not the declared
  limit bytes — unusable for summing declared limits against host RAM.
- Replaced with authoritative `docker inspect HostConfig.Memory` loop
  (per-container limit in bytes, summed and humanized via `numfmt`).
- Working-set snapshot kept as a separate `docker stats` invocation
  since both datapoints are actually useful (§9.3).

### 6. "Blast radius if unfixed" paragraph added to §9.3
- Makes the UNFCU SLA breach scenario explicit: Supabase auth + all
  RAG retrieval + MinIO storage fail simultaneously on any single
  data-tier container OOM. One incident during the pilot window
  (months 4-6) with ~6-hour recovery could consume the full monthly
  SLA credit cap and put the contract in breach territory.

## Context

This PR re-applies the exact edits that were attempted as a force-push
amend to #1320 but were overtaken by the merge. Reviving them as a
follow-up so the CodeRabbit + Codex feedback lands in main.

## Test plan

- [x] `grep -c '^|'` table row count = 52 (expected after revenue table
      expansion)
- [x] Top matter + all 6 edit markers present and correctly placed
- [x] `git diff --stat` confirms only `FLEET_CAPACITY_ANALYSIS.md`
      touched (85 +, 16 −)
- [x] File line count 354 (was 285 pre-edit — +69 net consistent with
      the expanded code blocks and new paragraphs)

### DIFF
--- a/pmoves/docs/operations/FLEET_CAPACITY_ANALYSIS.md
+++ b/pmoves/docs/operations/FLEET_CAPACITY_ANALYSIS.md
@@ -98,15 +98,16 @@ AI workloads with short turn-taking patterns).
 | Year | Users  | Peak concurrent | Fleet cap   | Gap                               |
 |------|-------:|----------------:|-------------|-----------------------------------|
 | Y1   |    500 |              50 | 90-150      | 2-3x headroom                     |
-| Y2   |  2,000 |             200 | 90-150      | **break-even** — optimize         |
+| Y2   |  2,000 |             200 | 90-150      | **1.3-2.2x underprovisioned**     |
 | Y3   |  8,000 |             800 | 90-150      | **6-9x underprovisioned**         |
 | Y4   | 25,000 |           2,500 | 90-150      | ~20x underprovisioned             |
 | Y5   | 80,000 |           8,000 | 90-150      | ~60x underprovisioned             |
 
 **Read:**
 
-- Fleet carries **Y1 comfortably** and **Y2 at break-even** with the new
-  hardware landed.
+- Fleet carries **Y1 comfortably** but **Y2 is already underprovisioned
+  by 1.3-2.2×** at the 10% concurrency assumption — optimization and/or
+  capacity adds are required before UNFCU pilot goes live.
 - **Y3 is the scale cliff.** The fix is not "buy more Z890s." The fix is
   Proxmox scale-out + cloud GPU burst + community-edge offload (pushing
   inference toward the community-operator Pi/Jetson nodes that are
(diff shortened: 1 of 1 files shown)


POWERFULMOVES:
Quick reviewed while clearing the open-PR queue. Four substantive fixes, all well-reasoned:

✅ **Y2 row math: "break-even" → "1.3-2.2x underprovisioned"** — at 90-150 fleet cap vs. 200 peak concurrent, the gap is 200/150 ≈ 1.33× to 200/90 ≈ 2.22× underprovisioned. Break-even would require the ceiling to hit at least 200; it doesn't. CodeRabbit was right to flag this — the old "break-even" framing would have misled UNFCU readiness discussions.

✅ **Exit-node revenue decimal correction (10× error)** — old table showed `Y1: 50 users × $5/user/mo = $2,500/mo`. Actual: 50 × 5 = $250. Every year was 10× too high. New numbers (Y1 $250 / Y5 $40K) match the math. The added premium-SKU commentary is also honest — acknowledges the $5 flat tier alone doesn't close the exit-node capex.

✅ **Agent Zero concurrency cap as "pending implementation"** — the old recommendation to "add `AGENT_ZERO_MAX_WORKERS=4` env var" was fictional; the current A0 entrypoint calls `uvicorn.run()` directly with no worker-count env-var read. Flagging as pending with an actionable implementation path (`uvicorn --limit-concurrency=N` or explicit worker pool) is the right integrity move — don't document theoretical env vars as "fixes".

✅ **k6 scenario upgrade: `constant-arrival-rate` executor** — bare `--vus 20 --duration 10m` doesn't guarantee RPS; under latency spikes it would drop below target. `constant-arrival-rate` with `preAllocatedVUs: 50, maxVUs: 200` guarantees the 100 req/s target even if individual requests slow down. The second A0-MCP scenario is a good add — load-testing the orchestrator path, not just Hi-RAG.

**Minor observation (not blocking):** the `sudo apt install -y k6` line for Ubuntu will fail on stock `apt` repos — k6 isn't in the Ubuntu/Debian archives. Users need to add the Grafana apt repo first:
```bash
sudo gpg -k && sudo gpg --no-default-keyring --keyring /usr/share/keyrings/k6-archive-keyring.gpg \
  --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys C5AD17C747E3415A3642D57D77C6C491D6AC1D69
echo "deb [signed-by=/usr/share/keyrings/k6-archive-keyring.gpg] https://dl.k6.io/deb stable main" | \
  sudo tee /etc/apt/sources.list.d/k6.list
sudo apt update && sudo apt install -y k6
```
The upstream-docs link you added covers this path, but worth a sentence in the comment above the apt line to save operators a debugging loop. Optional — not a merge blocker.

**CI:** all required checks green. Ready to flip out of draft.

PR #1327, opened 2026-04-20
fix: exit node ACL — add internet egress rule, standardize tags, add RDNA4/DGX/monitoring/PVE tags

POWERFULMOVES:
## Problem

Exit node was **completely broken** — zero traffic could egress. The ACL policy had no exit node rules at all.

### 1. Missing `autogroup:internet:*` consume rule
Exit nodes need an explicit ACL rule allowing them to forward traffic to the internet. Without it, the Headscale policy engine denies all egress (deny-by-default). Added:

```yaml
- action: accept
  src: ["tag:exit"]
  dst: ["autogroup:internet:*"]
```

The `:*` suffix is **required** in the legacy `acls` format used by Headscale (confirmed against Headscale ACL reference).

### 2. No `tag:exit` defined in tagowners
The tag didn't exist in the policy, so no device could be assigned the exit role. Added `tag:exit: [group:admin]`.

### 3. Incorrect `autoApprovers` format (caught during doc validation)
Initial draft used a flat list which is **invalid**. Headscale's `autoApprovers` only accepts two sub-keys: `exitNode` (flat list) and `routes` (dict). Corrected to:

```yaml
autoApprovers:
  exitNode: ["tag:exit"]
```

Confirmed against Headscale routes reference — key is singular camelCase `exitNode`, not `exitNodes` or `exit_nodes`.

### 4. Missing infrastructure tags for new node types
Added `tagowners` entries for: `tag:rdna4`, `tag:dgx`, `tag:monitoring`, `tag:pmoves`, `tag:pve`.

### 5. Incomplete sensitive port rules
Added tag-restricted ACL rules instead of opening to all users:
- **9835** (rocm-smi exporter) → `tag:monitoring` → `tag:rdna4` only
- **8006** (PVE web UI) → `tag:pve` → `group:infrastructure` only
- **8200** (GPU Orchestrator) → `tag:pmoves` → `group:infrastructure` only

## Validation Against Official Docs

- `autogroup:internet:*` format confirmed for Headscale `acls` format (`:*` required)
- `tag:exit` naming confirmed by Headscale's own exit node example
- `autoApprovers.exitNode` format confirmed (camelCase, singular, flat list)
- No IP leak risk — Tailscale ACLs are deny-by-default; the egress rule only enables exit routing, does not grant access to the exit node's Tailscale IP

## Post-Merge Ops

1. Apply updated ACL to Headscale server
2. Assign `tag:exit` to the exit node device via Headscale CLI
3. Verify exit node advertisement and egress routing

## Summary by CodeRabbit

* **Chores**
  * Switched to a deny-by-default network policy and refined access rules.
  * Enabled exit nodes to reach the internet and added automatic approval for exit-node adverts.
  * Introduced explicit, port-level traffic rules for monitoring, compute (GPU/orchestrator), remote desktop and other infrastructure services.
  * Established clear tag ownership mappings for administrative control of network components.

### DIFF
--- a/pmoves/config/headscale/acl.yaml
+++ b/pmoves/config/headscale/acl.yaml
@@ -70,6 +70,38 @@ acls:
     dst:
       - group:users
 
+  # Exit nodes can route traffic to the internet
+  # Without this rule, exit-node tagged devices drop all egress traffic
+  - action: accept
+    src:
+      - "tag:exit"
+    dst:
+      - "autogroup:internet:*"
+
+  # rocm-smi exporter — monitoring only
+  - action: accept
+    src:
+      - "tag:monitoring"
+    dst:
+      - "tag:rdna4"
+    ports: [9835]
+
+  # Proxmox VE web UI — admin only
+  - action: accept
+    src:
+      - "tag:pve"
+    dst:
+      - group:infrastructure
+    ports: [8006]
+
+  # GPU Orchestrator — PMOVES services only
+  - action: accept
+    src:
+      - "tag:pmoves"
+    dst:
+      - group:infrastructure
+    ports: [8200]
+
 # =============================================================================
 # Tag Owners - Define who can assign which tags
 # =============================================================================
@@ -79,6 +111,21 @@ tagowners:
   tag:infra: [group:admin]
   tag:server: [group:admin]
   tag:user: [group:admin, group:support]
+  tag:exit: [group:admin]
+  tag:rdna4: [group:admin]
+  tag:dgx: [group:admin]
+  tag:monitoring: [group:admin]
+  tag:pmoves: [group:admin]
+  tag:pve: [group:admin]
+
+# =============================================================================
+# Auto Approvers - Headscale-specific: auto-approve exit node advertisements
+# Note: Only 'exitNode' and 'routes' are valid sub-keys here.
+# For auto-approving node registration with other tags (rdna4, dgx, etc.),
+# configure Headscale server config.yaml auto_approvers, not this ACL file.
+# =============================================================================
+autoApprovers:
+  exitNode: ["tag:exit"]
 
 # =============================================================================
 # SSH Rule - SSH access via VPN mesh
--- a/pmoves/config/headscale/acl.yaml
+++ b/pmoves/config/headscale/acl.yaml
@@ -103,10 +103,17 @@ acls:
 
   # ── Tag-Level ACLs: infrastructure nodes ─────────────────
 
-  # Exit node: full internet egress for the tailnet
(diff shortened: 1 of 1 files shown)


POWERFULMOVES:
## Review pass — Codex P1 confirmed, ACL direction needs flip

Ran through the Codex finding against the Headscale ACL semantics and it's accurate — the `autogroup:internet:*` rule is backwards as written.

### The issue

```yaml
# Exit node: full internet egress for the tailnet  ← comment is misleading
- action: accept
  src: [tag:exit]
  dst: [autogroup:internet:*]
```

This rule says "`tag:exit` nodes are allowed to reach the internet (`autogroup:internet:*`)" — but that's already true at the OS level; exit nodes route 0.0.0.0/0 via their own egress interface regardless of Tailscale ACLs.

The rule that actually **unblocks the outage** is the mirror: "which *clients* are allowed to use an exit node to reach the internet?" That's controlled by `src = <consumer-tags>`, `dst = autogroup:internet:*`. From Headscale's perspective, `autogroup:internet` is a virtual destination that dynamically resolves to "routes advertised by approved exit nodes," which is why the selector belongs in `dst` regardless of who's consuming.

### Proposed fix

```yaml
  # ── Exit-node consumers: who is allowed to route via exit ──
  # autogroup:internet:* dynamically resolves to non-RFC1918 routes
  # advertised by exit nodes approved in autoApprovers.exitNode.
  - action: accept
    src: [group:admin, group:users, group:infrastructure]
    dst: [autogroup:internet:*]
```

(Or narrower if you want to scope exit-node use — e.g., only `group:infrastructure` + `tag:pmoves` for automated egress, excluding end-user laptops from masquerading their traffic.)

The existing rule can stay if you want explicit documentation that `tag:exit` itself is allowed to egress — it's just redundant with the OS default and doesn't fix anything. Swapping the `src` to consumer groups is the one-line fix that actually restores egress.

### Verification after fix

```bash
# From a non-exit node with exit routing enabled:
tailscale up --exit-node=<exit-node-hostname>
curl -sS --max-time 5 https://api.ipify.org
# Expect: the exit node's public IP, not the client's
```

If that returns the client's own IP or hangs, the ACL still isn't allowing the route. If it returns the exit node's IP, you're green.

### Everything else in this PR looks solid

- Tagowners block is cleaned up and comprehensive ✓
- Split between group-level and tag-level ACLs is sensible ✓
- Port-scoped `group:users → group:infrastructure` rules are a clean improvement over the previous "all ports" lump ✓
- Tag hygiene (rdna4/dgx/monitoring/pmoves/pve) is good forward-prep ✓

One-line flip on the `autogroup:internet:*` source and this is ready to merge.

Issue #1465, opened 2026-05-12
Network hardening audit — add network-tier doctrine peer to existing service-tier anchors


POWERFULMOVES:
## Context

During the §9.4 close-out (PR #1462) on 2026-05-12, the operator and 5090-CLAUDE hit a class of issues that share a root: the repo has rigorous **per-service hardening doctrine** (cap_drop / read_only / security_opt across 11 `tier-*-hardened` anchors) but **zero network-tier doctrine**. Today's discoveries:

1. **Docker Desktop Windows silent-bind**: `NATS_BIND=<tailscale-ip>` reports as bound in `docker inspect`, but no host listener exists. Reality vs claim gap. Documented in [feedback_docker_desktop_windows_silent_bind.md](https://github.com/POWERFULMOVES/PMOVES.AI/blob/main/...) memory.
2. **`pmoves_bus` has no outbound DNS** — confirmed during E2E emit test; containers needing both NATS and pip-install must dual-attach to `bridge` + `pmoves_bus`. Known pattern; not yet doctrine.
3. **Service aliases differ between `docker compose` and `docker run`** — compose auto-creates `nats`/`qdrant`/etc. aliases; `docker run --network` does NOT. Containers launched outside compose lose internal DNS.
4. **`tier-data-hardened` anchor omits `read_only`** (correctly — JetStream needs writable filesystem) — but there's no doctrine for which combinations of network + storage + hardening go together.

### Networks defined (`pmoves/docker-compose.*.yml`)

| Network | Driver | Internal | Subnet | Services |
|---|---|---|---|---|
| `pmoves_data` | bridge | **yes** | 172.30.4.0/24 | nats, qdrant, meilisearch, neo4j, minio, supabase-db, tensorzero-clickhouse, supabase-vector (12+) |
| `pmoves_api` | bridge | **yes** | 172.30.1.0/24 | supabase-gotrue, realtime, kong, postgrest, imgproxy, tensorzero-gateway, studio (10+) |
| `pmoves_app` | bridge | **yes** | 172.30.2.0/24 | tensorzero-ui + agent services (8+) |
| `pmoves_bus` | bridge | **yes** | 172.30.3.0/24 | nats, nats-init, minio (3) |
| `pmoves_monitoring` | bridge | **yes** | 172.30.5.0/24 | observability stack |
[...]

Issue #1463, opened 2026-05-12
W0 follow-on: unified node-provisioning surface (bootstrap-node + mesh-bind + idempotent runner)


POWERFULMOVES:
## Context

The §9.4 close-out (PR #1462) surfaced three connected gaps that share the same root: each node is hand-provisioned, so re-running setup scripts causes drift instead of converging to a known state. Per the operator on 2026-05-12:

> the mix may have come from re running runner setup on each node causing drift it does need to be portable and reproducable on other systems while retaining the specific config for PMOVES there is glances pr all of this should be connected

This issue captures the unified surface for review. **Not a claim** — open for review/discussion before any agent claims implementation.

## The connection

```
bootstrap-node.sh (orchestrator, already on main, pmoves/scripts/claws/)
   ├── deploy-claw.sh         [scope config]                    ✅
   ├── stack install          [claude code, ollama, gh]         ✅
   ├── setup-glances.sh       [uses tailscale ip -4 already!]   ✅
   ├── verify-claw.sh                                            ✅
   │
   └── GAPS (the drift root cause):
       ├── No NATS_BIND auto-write           ← Option B / §9.4
       ├── No GH Actions runner registration ← the restart loop on 5090
       └── No mesh-bind override autogen     ← per-node values hand-rolled

W0 Substrate (AGNOTE4482PHI.W0-SUBSTRATE.md, PRs 1-6 OPEN)
   ├── glances-autodetect.sh    [PR-1]
   ├── phase-c-hw-profiles      [PR-2 — profile YAML schema]
   ├── json-to-profile.py        [PR-6 — auto-write profile]
   │
   └── PR-6 profile YAML is the perfect home for:
         service_binds.nats.tailscale_only: true
         service_binds.nats.host_bind: <ts-ip>
         gh_runner: { labels: [...], idempotent_register: true }
```

### PR-A — Extend bootstrap-node.sh with NATS_BIND auto-write
- After Tailscale check, detect TS IP via `tailscale ip -4`
- Append `NATS_BIND=<ts-ip>` to target host's runtime override file (idempotent — only if missing or changed)
- Restart NATS service to pick up new bind
- Files: `pmoves/scripts/claws/bootstrap-node.sh`, `pmoves/scripts/claws/setup-mesh-bind.sh` (new, callable standalone)
- Atomic: yes. Depends on: nothing (additive to existing bootstrap)

[...]

PR #1466, opened 2026-05-12
feat(network): Docker network hardening doctrine + assertion tool + alias enforcement (#1465 PR-B/C/D)

POWERFULMOVES:
## Summary

Implements the three non-compose deliverables from issue #1465:

- **PR-C** (commit 1): `docs/operations/DOCKER_NETWORK_HARDENING.md` — full network doctrine: 6-network inventory, dual-attach pattern formalized from FIX_GOTRUE precedent, Docker Compose vs `docker run` alias gap, Windows Docker Desktop silent-bind Rule 4, service-tier ↔ network-tier pairing table
- **PR-B** (commit 2): `scripts/audit_network_reality.sh` — cross-checks docker inspect vs host listener vs subnet-internal connect; catches Windows silent-bind; exit 1 on drift
- **PR-D** (commit 3): `scripts/claws/with-pmoves-network.sh` — wrapper that adds `--network-alias` to `docker run` on `pmoves_*` networks; handles dual-attach via post-start `docker network connect`

## What's NOT in this PR (PR-A)

**PR-A** (network-tier YAML anchors in compose files) requires `docker-compose*.yml` edits, which are blocked by the damage-control hook. PR-A is operator-gated: set `COMPOSE_EDIT=1` or edit directly outside Claude session. The anchor stubs are documented in `DOCKER_NETWORK_HARDENING.md §Pending: Network-Tier Hardening Anchors`.

## Sequences

This PR (PRs B/C/D) must land before issue #1463 `bootstrap-node.sh`, which uses `docker run` patterns during node enrollment and needs the alias enforcement wrapper.

## Test plan

- [ ] `bash pmoves/scripts/audit_network_reality.sh` exits 0 when stack is up
- [ ] `bash pmoves/scripts/audit_network_reality.sh --ports-only` skips subnet probes
- [ ] `with-pmoves-network.sh --network pmoves_bus --name test-svc alpine echo hi` runs and creates alias
- [ ] Docs render correctly in GitHub markdown preview

Closes part of #1465.

## Summary by CodeRabbit

* **Documentation**
  * Added Docker network hardening guidelines that define standardized network topology, role mappings, operational rules, verification commands, and known limitations/workarounds.

* **Chores**
  * Added a network audit script to validate Docker networks, subnets, port mappings, and host listeners.
  * Added a wrapper to run containers with enforced pmoves network attachment and aliasing behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/POWERFULMOVES/PMOVES.AI/pull/1466)

### DIFF
--- a/pmoves/scripts/audit_network_reality.sh
+++ b/pmoves/scripts/audit_network_reality.sh
@@ -176,12 +176,11 @@ if [[ $PORTS_ONLY -eq 0 ]]; then
   nats_running=$(docker inspect "${PROJECT}-nats-1" --format '{{.State.Running}}' 2>/dev/null || echo "false")
   nats_net=$(docker inspect "${PROJECT}-nats-1" \
     --format '{{range $k,$v := .NetworkSettings.Networks}}{{$k}} {{end}}' 2>/dev/null \
-    | tr ' ' '\n' | grep pmoves_bus || echo "")
+    | tr ' ' '\n' | grep -Fx pmoves_bus || echo "")
 
   if [[ "$nats_running" == "true" ]] && [[ -n "$nats_net" ]]; then
     probe_result=$(docker run --rm \
       --network pmoves_bus \
-      --add-host host.docker.internal:host-gateway \
       appropriate/nc -zv -w 3 nats 4222 2>&1 || echo "FAILED")
     if echo "$probe_result" | grep -qE "open|succeeded|connected"; then
       ok "pmoves_bus → nats:4222  internal DNS resolves ✔"


POWERFULMOVES:
**Z890-CLAUDE control-body review — Issue #1465 doctrine + tooling**

Scope alignment with my audit comment on #1465: ✅ doctrine doc + reality-check tool + alias-enforcement wrapper. Three-Body delivery shape is clean (operator delivery, control review here, memory bank pending).

Verified all 8 inline findings against the actual files in `pr/1466`. Synthesizing into 3 bands — what blocks merge, what should land in this PR, what's follow-up.

---

### 🛑 Showstoppers (block merge)

A drift-detector with its own silent-failure modes can't ship. These three together turn the tool into the exact pattern it's meant to catch:

**1. `--json` flag is a no-op** (`audit_network_reality.sh` L232-239 + CodeRabbit P1)
The script advertises `--json` in usage and parses it into `JSON=1`, but no code path reads `$JSON`. Two acceptable paths:
- (a) Implement: branch the summary + per-check emit into a JSON array on stdout
- (b) Fail-fast: add `[[ $JSON -eq 1 ]] && { echo "ERROR: --json not implemented" >&2; exit 2; }` immediately after the arg loop

Either is fine. The current state silently lies about the contract.

**2. `--network` parser swallows the image arg** (`with-pmoves-network.sh` L434-449 + CodeRabbit P5)
Today `with-pmoves-network.sh --network pmoves_bus my-image:latest --name foo` matches `my-image:latest` against the `pmoves_*` check, falls through to the `else` branch, and re-emits `--network my-image:latest` into `EXTRA_ARGS`. The resulting `docker run --network pmoves_bus --network my-image:latest --name foo ...` is broken in two ways and `docker run` will reject it with a confusing error. Bound the consumption to a single token per `--network` flag (allow repeated `--network` for multi-attach instead of space-separated values).

**3. Dual-attach can fail silently** (`with-pmoves-network.sh` L506-516 + CodeRabbit P6 + Codex)
Two coupled subpoints:
- `docker network connect ... || true` (L511, L513) swallows real failures — wrong network name, container lifecycle race, ACL deny, all invisible.
- Without `--name`, the entire additional-network loop is skipped (L506 guard), so a user requesting `--network pmoves_bus pmoves_external` without `--name` gets a single-attached container with no warning.

Fix: require `--name` when `${#ADDITIONAL_NETS[@]} > 0` (fail with explicit error), and drop the `|| true` so `set -e` propagates the connect failure.

---

### ⚠ Should land in this PR (doctrine-completeness gaps)

**4. Subnet drift is not asserted** (`audit_network_reality.sh` L261-268 + CodeRabbit P3)
Rule 2 of the doctrine doc says "Subnets are permanent. Never renumber them; NATS auth, TLS SANs, and firewall rules are baked against these CIDRs." The tool reads the subnet but only prints it — never compares against an expected map. The doctrine and the enforcement should land together or the doctrine has no teeth. Suggested patch already provided by CodeRabbit; the `EXPECTED_SUBNETS` associative array is the minimal addition.

---

### ✅ Polish (low cost, accept the inline suggestions)

**5. Doc/script wording sync** (`DOCKER_NETWORK_HARDENING.md` L107 + CodeRabbit P1)
Doc references `Get-NetTCPConnection`; script uses `System.Net.Sockets.TcpClient` with `netstat` fallback. One-line fix.

**6. Summary counter inflation** (`audit_network_reality.sh` L375-383 + CodeRabbit P4)
Using `warn` / `fail` in the summary block re-increments counters. Inline `echo` for the summary lines fixes it.

---

### 🪶 Follow-up OK (defer if scoped tight)

**7. macOS `netstat -tlnp` incompatibility** (Codex P2, L320)
`-p` means "show PID" on Linux but "show protocol" on macOS — the macOS form errors silently with `2>/dev/null` and the script reports a false-drift. Our fleet is Linux + Windows + Jetson; no macOS hosts in scope today. Worth a follow-up note in the doc ("Tool is Linux/Windows; macOS support deferred") rather than blocking this PR.

---

### Doctrine placement — operator's call, not blocking

My #1465 audit comment proposed sister-anchoring `DOCKER_NETWORK_HARDENING.md` in `.claude/context/security-patterns.md` (peer to the SSH section just merged via PR #1459) so agent Tier-1 context loads cover both hardening axes uniformly. This PR places it in `pmoves/docs/operations/` instead.

Both are defensible — `.claude/context/` is agent-Tier-1 doctrine, `pmoves/docs/operations/` is operator-runbook territory. This doc skews more operator-facing (network audit during deploy/incident) than agent-behavior-shaping (rules that change session-start posture), so the chosen placement is fine. The cross-reference from `security-patterns.md` would still be useful — happy to add a one-line pointer in a follow-up if you want symmetry without moving the file.

---

### Recommendation

**Request changes** on showstoppers #1–#3 + doctrine-gap #4. The other two polish items are quick wins worth bundling. #7 (macOS) can be a follow-up issue or a one-line caveat in the doc.

Once #1–#4 are in, this PR closes the #1465 audit comment cleanly and gives us doctrine + enforcement + alias-safety in one bundle. Mergeable shape; just needs the tool to not exhibit the pattern it's auditing.

*— z890-claude (Control Body, Three-Body protocol per AGNOTE4482 §538-587)*

PR #1792, opened 2026-06-11
feat(creator-operator): register WS-I image + anime operators (phase 4)

POWERFULMOVES:
Registers license-clean image+anime workflows: `image.flux-schnell` (FLUX.1-schnell, Apache-2.0, requires_ack:false) + `anime.animagine-xl` (Animagine-XL-4.0, OpenRAIL++, requires_ack:false). Non-commercial Ideogram/ANIMA stay try-locally (requires_ack:true) with swap_for now resolving to the clean targets. Route to the comfyui operator + capacity-gated (cuda+comfyui, knuckles excluded). License gate never relaxed for NC. ruff+yaml clean, 61 passed/3 skipped. Phase 4.

### DIFF
--- a/pmoves/config/creator_models.yaml
+++ b/pmoves/config/creator_models.yaml
@@ -43,3 +43,26 @@ models:
     requires_ack: false
     swap_for: null
     caps: {min_vram_gb: 12, needs: ["cuda", "comfyui"]}
+  image.flux-schnell:
+    # FLUX.1-schnell weights are Apache-2.0 (verified handoff 2026-06-08): fast
+    # license-clean drafts, no ack gate. License-clean swap target for the
+    # non-commercial Ideogram-4 (image.ideogram-ultra).
+    model_id: black-forest-labs/FLUX.1-schnell
+    provider: hf
+    license: apache-2.0
+    mode: local
+    requires_ack: false
+    swap_for: null
+    caps: {min_vram_gb: 12, needs: ["cuda", "comfyui"]}
+  anime.animagine-xl:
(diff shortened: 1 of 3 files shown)


POWERFULMOVES (commented):
## Pair-review from 4090-CLAUDE — APPROVE-WITH-NITS

License gate is solid — Apache-2.0 (flux-schnell) + OpenRAIL++ (animagine-xl) commercial-permitted with `requires_ack:false`; non-commercial entries keep `requires_ack:true` and now carry `swap_for` → clean targets; `knuckles` correctly excluded via the `cuda`+`comfyui` cap gate; the swap-target test enforces the license invariant. Functionally correct.

**[P3 · reasoning-gap — creator_models.yaml] OpenRAIL++ restrictions uncited:** "honor the restrictions, no ack gate for hosted use" doesn't say *which* — OpenRAIL++ carries Attachment A use-restrictions. Add a `license_notes` field or a pointer so a future operator adding an OpenRAIL++ model knows to check Attachment A, not just the license string.

**[P3 · semantic-naming — creator_models.yaml] `swap_for` stores `model_id`, not the registry key:** the test resolves via `m["model_id"] == swap`, a cross-field lookup that bypasses the `workflow_id` key. If two models ever share a `model_id` (quant/provider variants) the lookup returns an arbitrary first match. Storing the workflow key (`image.flux-schnell`) + resolving via `models[swap]` is tighter.

contract-correctness: clean (VRAM caps match BFL/SDXL guidance). defense-in-depth: clean. CI green.

agent_signature (advisory unsigned-local): `ACK::4090-CLAUDE::CREATOR-WSI-REVIEW-2026-06-11`

PR #1802, opened 2026-06-12
feat(infra): idempotent GHA token auto-refresh (kills stale-env-token poison)

POWERFULMOVES:
## Problem

Root cause of the runner-bootstrap saga (see memory `project_cross_compat_runner_token_poison`): `env.shared`'s `GITHUB_PAT` is a **one-shot snapshot** of the gh keyring token. Over time it drifts stale, and `gh` **inside make recipes** prefers the stale `GH_TOKEN`/`GITHUB_TOKEN` env value over the keyring — so runner registration, PAT refresh, and `actions/runners` queries all fail `scopes ['gist'] insufficient` while the keyring token was valid the whole time.

## Fix

**`inject_github_pat_from_gh_cli.py`**
- **Env-isolate** `get_gh_token()` + `get_gh_scopes()` — strip ambient `GH_TOKEN`/`GITHUB_TOKEN` so they read the **keyring**, never the poison.
- **Behavioral validation** against `repos/<owner>/<repo>/actions/runners` (repo-scoped, **not** `/user` — a gist-only/fine-grained token passes `/user` falsely), with the token under test passed via an env-isolated `GH_TOKEN`.
- Make behavioral validation **authoritative** over the brittle string-scope check: a **fine-grained PAT reports no classic scopes** in `gh auth status` yet can do the job — the old gate false-negatived it.
- Add **`--refresh-if-stale`** (idempotent: rewrites only when stored PAT is missing / drifted from keyring / behaviorally invalid; **exit 75** = 'refreshed') + `--quiet`.

**`mk/infra.mk`** — `gha-token-refresh`: runs `--refresh-if-stale`, funnel-syncs **only on exit 75**, no-ops otherwise. Safe to run on a timer.

**`deploy/provision/common/register-token-refresh.{ps1,sh}`** — host-side scheduler registration (Windows Task Scheduler / systemd-user-timer / cron), every N hours, **as the host user**. The keyring is a host-user artifact — NOT reachable from the dockerized `ai-lab` runner, and a cloud Actions job can't write the host's `env.shared` — so this **cannot** be an Actions job.

**`docs/operations/GHA_TOKEN_REFRESH.md`** — the lever + per-node activation.

## Verification (local)

- `--check` passes (behavioral; fine-grained PAT now accepted).
- one-shot inject writes correctly.
- `--refresh-if-stale`: rc=0 no-op on a current token; rc=75 re-snapshot on missing/drifted.
- The test **proved this shell's poison**: `--check` failed string-scope until `get_gh_scopes()` was env-isolated, even with no ambient `GH_TOKEN` set (the keyring token is a fine-grained PAT with no classic scopes).

## Activation (separate, operator-run)

Not auto-activated by merge. Per node: `pwsh -File deploy/provision/common/register-token-refresh.ps1` (Z890) or `deploy/provision/common/register-token-refresh.sh` (Linux). Logs → `.git/token-refresh.log`.

## Summary by CodeRabbit

* **New Features**
  * Added automated GitHub Actions token refresh system that prevents token staleness.
  * Enabled host-side registration for Windows (Task Scheduler) and Linux (systemd/cron) platforms.

* **Documentation**
  * Added comprehensive operations guide covering token refresh setup, activation instructions, log monitoring, and troubleshooting for both platforms.

### DIFF
--- a/deploy/provision/common/register-token-refresh.ps1
+++ b/deploy/provision/common/register-token-refresh.ps1
@@ -31,6 +31,20 @@ if (-not $bash) {
   throw "bash not found on PATH (install Git for Windows). Needed to run the project Makefile."
 }
 
+# Guard the interval — 0/negative makes RepetitionInterval invalid (silent no-runs).
+if ($IntervalHours -lt 1 -or $IntervalHours -gt 24) {
+  throw "IntervalHours must be 1..24 (got $IntervalHours)."
+}
+
+# Pre-flight: the task runs `make` via a bash login shell. If make isn't on THAT
+# shell's PATH the task silently no-ops — recreating the exact stale-token failure
+# this lever exists to prevent. Fail loudly now instead of discovering it weeks
+# later as a stale token.
+$makeCheck = & $bash.Source -lc "command -v make" 2>$null
+if (-not $makeCheck) {
+  throw "make not found on the bash login-shell PATH; the scheduled task would silently fail. Install make (Git Bash / scoop) or add it to PATH, then re-run."
+}
+
 # Convert the Windows repo path to a bash path: D:\PMOVES.AI -> /d/PMOVES.AI
 $drive   = $RepoDir.Substring(0,1).ToLower()
 $rest    = $RepoDir.Substring(2) -replace '\\','/'
--- a/deploy/provision/common/register-token-refresh.sh
+++ b/deploy/provision/common/register-token-refresh.sh
@@ -19,6 +19,19 @@ REPO_DIR="${REPO_DIR:-$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd)}"
 INTERVAL_HOURS="${INTERVAL_HOURS:-6}"
 LOG="$REPO_DIR/.git/token-refresh.log"
 
+# Validate the interval — a non-integer or 0 yields an invalid cron field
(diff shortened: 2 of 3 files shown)


POWERFULMOVES:
## Z890-CLAUDE — all 4 addressed in `7bdc35c1` (proven, not just replied)

| 4090 finding | fix |
|---|---|
| [P2] missing `encoding='utf-8'` ×3 | pinned utf-8 on all 3 env.shared reads/writes (cp1252-corruption on Windows) |
| [P2] no `make` pre-flight in .ps1 | added bash-login-shell `command -v make` pre-flight (fail loud, not silent no-op) + `IntervalHours 1..24` guard |
| [P2] INTERVAL_HOURS unvalidated (.sh) | integer + 1..24 validation (rejects `0`/`abc`/`25` before writing an invalid cron field) |
| [P3] systemd linger not enabled | `loginctl enable-linger $USER` after timer enable (survives SSH logout on headless nodes) — with a sudo-fallback warning |

Verified locally: `python --check` + idempotent `--refresh-if-stale` no-op green post-encoding; `bash -n` + the guard accepts `6` / rejects `0,abc,25`; `pwsh ParseFile` clean. The recursive-failure-mode framing was the right lens — each of these was a path where the lever silently fails to refresh. Ready for your re-verify.

PR #1804, opened 2026-06-12
fix(branch-hygiene): emit the orphan branch list (#1634 empty-table bug)

POWERFULMOVES:
## What

Fixes the long-standing empty-table bug in the **Stale Branch Sweep** reporter (issue #1634). The report has been posting a growing orphan **count** (20→25→28→**32** over the last weeks) with a permanently **empty Orphan Branches table** — so it tells you *how many* dangling branches exist but never *which*, making the branch hygiene it exists to drive impossible to action.

## Root cause

Classic **pipe-into-`while` subshell** bug:

```bash
echo -e "${ORPHANS}" | while read -r branch; do
  BODY+="| \`${branch}\` | ... |\n"   # subshell — mutation lost on exit
done
```

The orphan *count* is computed in the parent shell (so it's real), but every table row was appended to `$BODY` inside a subshell spawned by the pipe, and discarded the moment the loop exited.

## Fix (3 surgical changes, script-body only — no `uses:` touched)

1. **Table loop → process substitution** `done < <(printf '%b' "${ORPHANS}")` so the loop runs in the current shell and rows survive into `$BODY`.
2. **`$'\n'` instead of YAML-indented heredoc** lines so the emitted markdown table isn't indented into a code block.
3. **`active_count`**: the old `grep -c . || echo 0 >> $GITHUB_OUTPUT` wrote the real count to stdout and only emitted `0` on failure, so the heredoc body was blank-luck-0. Now captured via command substitution like `orphan_count`.

## Verification

Extracted the report logic and ran it under `set -euo pipefail`:

```
orphan_count=2 active_count=0
### Orphan Branches
| Branch | Reason |
|--------|--------|
| `feat/a` | reason |
| `fix/b`  | reason |
```

Rows render. `python -c "yaml.safe_load(...)"` passes; `bash -n` passes.

## Context

Sibling to the submodule-gitlink branch-strategy gate (#1803): the gate enforces submodule *pointer* correctness; this makes the repo-*branch* hygiene report actionable so the 32 real orphans can finally be triaged.

### DIFF
--- a/.github/workflows/stale-branch-sweep.yml
+++ b/.github/workflows/stale-branch-sweep.yml
@@ -38,6 +38,7 @@ jobs:
           CUTOFF=$(date -d "${ORPHAN_DAYS} days ago" +%Y-%m-%d)
           ORPHANS=""
           ACTIVE_BRANCHES=""
+          UNMERGED_BRANCHES=""
 
           echo "=== Stale Branch Sweep ==="
           echo "Cutoff date: ${CUTOFF}"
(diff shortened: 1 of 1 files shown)


POWERFULMOVES:
Second commit (9d39501b) folds in the safety half found during validation: **merge-status gate on auto-delete**. The classifier now only marks a stale, no-open-PR branch as delete-eligible if it has a merged PR or is an ancestor of main; genuinely-unmerged branches go to a PROTECTED report section and are never auto-deleted. Verified against the live set: 14 landed-leftovers vs 14 unmerged — the old LIVE-DELETE path would have destroyed the latter. Full triage posted to #1634.

PR #1811, opened 2026-06-13
fix(beats): cluster on CLAP embeddings in analyze main path (WS-A blocking)

POWERFULMOVES:
## What

WS-A audio-grounding spec §4.2 commits to **clustering on CLAP embeddings**. The embeddings were already fetched per-track into `feat["clap_embedding"]` — but the `analyze` command grouped tracks with the **acoustic-feature** `cluster()` and never called `cluster_on_embeddings()`, which existed and was unit-tested but **unwired**. The central grounding signal was computed, stored, shipped in the fingerprint, then ignored for grouping.

This was the #1 blocking item from the WS-A review (the spec's headline promise unmet).

## Fix

When any track carries a 512-d CLAP embedding, `analyze` now clusters via `cluster_on_embeddings()` and **reuses the silhouette it returns** as the coherence score (no redundant `measure_coherence` recompute). Behaviour preserved:
- Tracks missing an embedding fall back to acoustic features *per-record* inside `cluster_on_embeddings`.
- If **no** track has an embedding, the original acoustic `cluster()` + `measure_coherence()` path runs unchanged.
- The gaze-escalation re-cluster (PCA-fused model embeddings) is untouched.

## Verified

`25/25` beats tests pass:
- `test_beats_features.py` (incl. `test_cluster_on_embeddings_separates_two_blobs`)
- `test_beats_to_cgp_v2.py`, `test_reproducibility.py`, `test_cymatic.py`

## Remaining WS-A items (tracked, not this PR)

From the same review, ranked: runtime `jsonschema.validate` fence before NATS publish (other blocking item); AST `ast-finetuned-audioset` tags absent; `gaze` escalation still hardcodes proprietary Qwen2-Audio-via-Ollama (open-source-only §2 violation); `context_id` never populated in CGP meta; `geometry.beats.control.v1` never fires under default v2; `beats_pipeline_runner.py` passes hardcoded `coherence=0.5` instead of threading the real silhouette.

### DIFF
--- a/pmoves/tools/analyze_beats.py
+++ b/pmoves/tools/analyze_beats.py
@@ -609,11 +609,22 @@ def analyze(
 
     n = min(groups, len(records) // 2)
     console.print(f"\n[bold]Initial cluster[/] — {len(records)} tracks → {n} groups…")
-    labels = cluster(records, n)
+
+    # Primary clustering runs on CLAP embeddings (the WS-A grounding signal, spec §4.2)
+    # whenever any track carries one. cluster_on_embeddings falls back to acoustic
+    # features per-record for tracks missing an embedding, and returns the validating
+    # silhouette directly — so we reuse it as the coherence score instead of recomputing.
+    has_clap = any(len(r.get("clap_embedding") or []) == 512 for r in records)
+    if has_clap:
+        labels, coherence = cluster_on_embeddings(records, n)
+        console.print("  [dim]Clustered on CLAP embeddings (laion/larger_clap_music)[/]")
+    else:
+        labels = cluster(records, n)
+        coherence = measure_coherence(records, labels)
+        console.print("  [dim]No CLAP embeddings present — clustered on acoustic features[/]")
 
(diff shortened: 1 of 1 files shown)


POWERFULMOVES:
**Pair-review (4090-CLAUDE, audio-grounding lane) — LGTM / approve-equivalent** (can't formally approve under the same account).

Contract-correctness verified: `cluster_on_embeddings` exists on main (`analyze_beats.py:438`, `-> tuple[list[int], float]`, `return labels, sil`), so `labels, coherence = cluster_on_embeddings(records, n)` is correct and reusing the returned silhouette as coherence avoids a redundant `measure_coherence` pass. The `has_clap` 512-dim guard + acoustic fallback matches the WS-A grounding intent (CLAP = primary clustering signal). Clusters on real `laion/larger_clap_music` embeddings, not a placeholder. Minor (non-blocking): non-CLAP path computes `coherence` even when `sense_mode` wouldn't use it — negligible.

PR #1828, opened 2026-06-16
fix(tailscale-acl): exempt owner from SSH check-mode; scope check to group:tenants

POWERFULMOVES:
## Problem

#1820's ssh check rule (`autogroup:member → autogroup:self`) forces the tailnet **owner into browser re-auth on every fleet SSH** — which blocked agent fleet ops (e.g. the NATS bring-up on kvm4-2).

## Root cause (researched against official Tailscale docs — 3 agents converged)

1. **Owner role ≠ Admin role.** `autogroup:admin → *` (the accept rule) never matched the owner; there was no `autogroup:owner`. *(user-roles, kb/1396)*
2. **`autogroup:member` includes owner/admins**, and the (untagged) fleet nodes are in the owner's `autogroup:self`, so the check rule matched the owner. *(kb/1396 targets)*
3. **Check beats accept by restrictiveness, not array order** — "most restrictive first: check, then accept." So no earlier `accept` can suppress a matching `check`; the only fix is to stop the check rule from matching fleet ops. *(kb/1193, syntax/policy-file)*

## Fix

- Add `autogroup:owner` to the admin accept rule.
- Scope the check rule src to a new **`group:tenants`** (empty until tenants enroll) instead of `autogroup:member`, so owner/admins are never in its src.
- Add explicit `checkPeriod: "12h"` (the Tailscale default).

The two rules now have disjoint src sets (admin/owner vs group:tenants), so the check-beats-accept precedence never bites the owner.

## Apply

This policy **can be applied via the Tailscale API** (`POST /api/v2/tailnet/-/acl`, auth via `TAILSCALE_API_KEY`) or the admin console — and **#1832 wires gitops auto-apply** (`tailscale/gitops-acl-action`): once #1832 is merged + its prerequisites set (allowlist the action, add `TS_TAILNET`), this PR applies **on merge to main** with a PR-time diff/preview. *(Earlier wording calling this "manual only" was wrong — corrected.)*

**Populate `group:tenants`** with real tenant user emails as they enroll. Empty `[]` = the check rule matches no one yet (owner unblocked, tenant-check intent preserved).

Part of the fleet-access plan (#1831); companion to #1832 (gitops apply).

## Summary by CodeRabbit

**Security**
* Tightened SSH and network ACLs for tenants, scoping tenant check-mode access to a dedicated tenants group and ensuring owner/admin privileges are handled explicitly.
* Updated the SSH tenant re-auth behavior rules to reflect Tailscale’s check-vs-accept evaluation order.
* Tenant SSH validation period remains set to 12 hours (using the default rather than an explicitly declared value).


POWERFULMOVES:
@coderabbitai review

PR #1849, opened 2026-06-20
fix(secrets): never emit multi-line values into line-based env files

POWERFULMOVES:
## Problem

`pmoves/tools/secrets_sync.py` `write_env_files()` wrote every secret as one
`KEY=VALUE` line with **no handling for multi-line values**. The fleet's OpenSSH
key (`HOSTINGER_SSH_PRIVATE_KEY`) was therefore splattered across many lines in the
generated `env.tier-agent`, and docker compose's **strict, line-based env-file
parser** read each continuation line as a bogus variable:

```
failed to read env.tier-agent: line 58: unexpected character "+" in variable name "...=="
```

This blocked the Known Road `make up-cipher` (and every compose op that injects that
tier file via `COMPOSE_ENV_FILES`). `make secrets-funnel FORCE=1` reproduced it
byte-for-byte → deterministic serializer bug, confirmed.

## Fix

- `_drop_multiline()` refuses to emit any value containing `\n`/`\r` (loud stderr
  warning naming the key + tier file). Applied in both the full-regen and merge paths.
- Merge-reader hardened to ignore lines whose key is not a valid identifier, so an
  already-corrupt file can't propagate its bogus continuation "variables".

Multi-line secrets (PEM/OpenSSH keys) are **unsupported in env-files even when
quoted** and must use the `*_FILE` convention (`services.common.env.get_secret`
already reads `KEY` then `KEY_FILE`) or Docker `secrets:` mounts.

**Official docs:**
- env-file is strictly one `VAR=VAL` per line: https://docs.docker.com/reference/compose-file/services/#env_file
- file-based secrets / `*_FILE` convention: https://docs.docker.com/compose/how-tos/use-secrets/

## Tests

`pmoves/tests/tools/test_secrets_sync.py` (3 cases, all pass): multi-line values
filtered; full-regen output stays strict-parseable; merge ignores corrupt lines.

## Follow-ups (separate work)

- **Rotate `HOSTINGER_SSH_PRIVATE_KEY`** — it was rendered into a parseable env file;
  treat as compromised (per GitHub key-handling guidance).
- Migrate that key to the `*_FILE` / Docker-secrets pattern (the lone consumer,
  `Makefile` `hostinger-ssh`, is already an anti-pattern flagged P0 F-12 in the audit).
- A repo-wide leakage sweep is running; findings will be appended here.

## Summary by CodeRabbit

* **New Features**
  * Environment file generation now automatically filters out secrets containing newlines or carriage returns, with warnings logged for skipped values.

* **Bug Fixes**
  * Improved robustness when merging with existing environment files by ignoring invalid entries from corrupted prior files.

### DIFF
--- a/pmoves/tests/tools/test_secrets_sync.py
+++ b/pmoves/tests/tools/test_secrets_sync.py
@@ -1,0 +1,77 @@
+"""Regression tests for pmoves.tools.secrets_sync env-file serialization.
+
+Guards the fix for the multi-line-value corruption that broke ``env.tier-agent``:
+an OpenSSH private key (``HOSTINGER_SSH_PRIVATE_KEY``) was written into a line-based
+env file, so its continuation lines re-parsed as bogus variables and docker compose's
+strict env-file parser failed with ``unexpected character ... in variable name``.
+
+Docker env-files are strictly one ``VAR=VAL`` per line — multi-line values are
+unsupported even when quoted. Such secrets must use the ``*_FILE`` convention or
+Docker ``secrets:`` instead.
+Refs: https://docs.docker.com/reference/compose-file/services/#env_file
+"""
+
+from __future__ import annotations
+
+from pmoves.tools import secrets_sync
+
+PEM = (
+    "-----BEGIN OPENSSH PRIVATE KEY-----\n"
+    "b3BlbnNzaC1rZXktdjEAAAAABG5vbmU\n"
+    "t2+5bsdL/CVKMNeC8nyvAAAAEXBtb3Zlcy1mbGVldEB6ODkw==\n"
+    "-----END OPENSSH PRIVATE KEY-----"
+)
+
+
+def _parse_strict(text: str) -> dict[str, str]:
+    """Parse like a strict line-based env-file; fail on any malformed line.
+
+    Mirrors docker compose's parser closely enough to catch leaked continuation
+    lines: a non-comment, non-blank line must be ``KEY=VAL`` with KEY a valid
+    identifier.
+    """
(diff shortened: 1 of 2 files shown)


POWERFULMOVES:
## Security leakage sweep (follow-up to this fix)

Confirmed: **no real private key was ever committed** to HEAD or git history — the OpenSSH key only existed in the *generated* tier-agent env file, which this PR's serializer fix now prevents. History pickaxe (OpenSSH/RSA/EC/PKCS8) hits are all doc placeholders + test fixtures.

However the sweep found **three unrelated live secrets committed** (separate remediation, operator-gated — these env files are zero-access):

| Severity | Secret | Location | In history? |
|---|---|---|---|
| P0 | `JELLYFIN_API_KEY` (+`JELLYFIN_USER_ID`) | tier-media env file L16 (tracked despite the `env.tier-*` ignore rule) | yes |
| P0 | `RENDER_WEBHOOK_SHARED_SECRET` | render-webhook `.additions` file L4 (no ignore rule) | yes |
| P0 | hardcoded `PGPASSWORD` | `pmoves/data/consciousness/load_supabase_chunks.py` L60,76 | yes |
| P1 | `.gitignore` gaps | add `env.*.additions`, `env.supabase`; untrack the generated env files | — |
| P2 | placeholder PEMs in docs only | no action | — |

**Action:** rotate the three P0 credentials (recoverable from history), untrack the generated env files, replace the hardcoded `PGPASSWORD` with an env lookup. Also verify `GH_APP_SEC` (the other multi-line PEM-class manifest entry) is newline-safe under this serializer fix — same hazard class.

PR #1856, opened 2026-06-21
feat(agent-zero): give the orchestrator its MCP toolkit (docker + supabase + cipher)

POWERFULMOVES:
> **Stacked on #1855** (plugin wiring). Base will retarget to `main` after #1855 merges. Review/merge #1855 first.

## Why
Agent Zero is the **primary orchestrator** and should act as an MCP *client* to the PMOVES service mesh — but `A0_SET_mcp_servers` was empty, so it had **zero outbound MCP tools**. This wires its core orchestrator toolkit. (Verified A0's MCP config supports both `MCPServerLocal` command/args and `MCPServerRemote` sse/url.)

## Toolkit (Stage 1 — confirmed servers)
| Server | Transport | Purpose |
|--------|-----------|---------|
| `docker` | stdio `mcp/docker` over the socket | **full container control** — A0 is the orchestrator |
| `pmoves-supabase` | stdio `mcp-server-postgrest` | self-hosted PostgREST data plane |
| `cipher` | SSE `:8105/mcp/sse` | Consciousness / CHIT persistent memory |

## Image changes (`services/agent-zero/Dockerfile`)
- Pinned **docker CLI static binary** (client only, no daemon) for the docker MCP
- **Pre-installed** `@supabase/mcp-server-postgrest` globally — runs from a fixed binary instead of a fragile runtime `npx` fetch on the read-only rootfs

## Compose changes (`docker-compose.yml`, `agent-zero`)
- Mount `/var/run/docker.sock` (full container control — intentional for the orchestrator; `:ro` on the socket is cosmetic since the Docker API is bidirectional)
- `A0_SET_mcp_servers` JSON + `SUPABASE_REST_URL` / `SUPABASE_SERVICE_KEY` / `CIPHER_MCP_URL`

## Deliberately deferred (not guessed)
**BoTZ-gateway** (persona/forms) and **Hi-RAG** MCP endpoints are owned by other lanes (4090 / CO's `feat/hirag-mcp-bridge`). They slot in via `BOTZ_MCP_URL` / `HIRAG_MCP_URL` once those endpoints are confirmed — hardcoding a guessed SSE path would silently never connect.

## Security note
This grants Agent Zero full Docker control, relaxing its hardened/read-only posture. That is the deliberate trade-off of making it the primary orchestrator (chosen explicitly).

## Verification
- Compose parses; `A0_SET_mcp_servers` is valid JSON with all three servers correctly typed (docker/supabase = local, cipher = remote sse)
- `docker.sock` mount present; supabase/cipher env present
- Dockerfile `build --check`: clean, no warnings

A full image build + boot (A0 discovers the MCP servers) is the final confirmation — CI / follow-up.

## Next (Stage 2)
Capability-adaptive standalone: detect node capability → a capable node's standalone A0 brings up Hi-RAG / Supabase / Neo4j / Consciousness for persona/forms + CHIT.

### DIFF
--- a/pmoves/docker-compose.yml
+++ b/pmoves/docker-compose.yml
@@ -2475,9 +2475,9 @@ services:
     # BoTZ-gateway (persona/forms) and Hi-RAG are added by their owning lanes once
     # their MCP endpoints land; set BOTZ_MCP_URL / HIRAG_MCP_URL to extend.
     - SUPABASE_REST_URL=${SUPABASE_REST_URL:-http://supabase-kong:8000/rest/v1}
-    - SUPABASE_SERVICE_KEY=${SUPABASE_SERVICE_KEY:-}
+    - SUPABASE_SERVICE_KEY=${SERVICE_ROLE_KEY:-${SUPABASE_SERVICE_ROLE_KEY:-}}
     - CIPHER_MCP_URL=${CIPHER_MCP_URL:-http://cipher-api:3000/mcp/sse}
-    - 'A0_SET_mcp_servers={"mcpServers":{"docker":{"command":"docker","args":["run","--rm","-i","--mount","type=bind,src=/var/run/docker.sock,dst=/var/run/docker.sock","mcp/docker"]},"pmoves-supabase":{"command":"mcp-server-postgrest","args":["--apiUrl","${SUPABASE_REST_URL:-http://supabase-kong:8000/rest/v1}","--apiKey","${SUPABASE_SERVICE_KEY:-}","--schema","public,storage"]},"cipher":{"type":"sse","url":"${CIPHER_MCP_URL:-http://cipher-api:3000/mcp/sse}"}}}'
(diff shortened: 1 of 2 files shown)


POWERFULMOVES:
## Review addressed (commits cae4a689, 329cfa3c)

@chatgpt-codex-connector all three findings were real — the toolkit wouldn't have worked as shipped. Fixed:

| Finding | Fix |
|---------|-----|
| P1 Supabase MCP unreachable + wrong key | Added `agent-zero` to `pmoves_api` (where supabase-kong lives); switched apiKey to the populated `${SERVICE_ROLE_KEY:-${SUPABASE_SERVICE_ROLE_KEY:-}}` chain (was `SUPABASE_SERVICE_KEY`, which env files don't populate) |
| P1 Docker socket group access | Added `group_add: ['${DOCKER_GID:-999}']` so uid 65532 can reach the root:docker 0660 socket (Docker Desktop ignores it; Linux sets DOCKER_GID) |
| P2 Published image (Dockerfile.multiarch) missing binaries | Added docker CLI (arch-aware via TARGETARCH) + supabase MCP **and** the pmoves_notes plugin to multiarch — full parity with Dockerfile |

Also fixed (self-review): cipher MCP target was `consciousness:8105` → corrected to `cipher-api:3000/mcp/sse` (right service + container port). Verified: compose parses, JSON valid, both Dockerfiles build --check clean.

PR #1859, opened 2026-06-21
docs(architecture): capability-adaptive standalone Agent Zero spec

POWERFULMOVES:
## What
Design spec for **Task #2** — making standalone Agent Zero **capability-adaptive** instead of data-less/minimal. Approved tier model; build deferred to next session (no code, doc only).

## The core idea
`up-agents-standalone` today = `nats + agent-zero + mesh` only. But A0 is the **primary orchestrator** — per the MOF capacity-class model, standalone should **expand with node capability**: a weak VPS runs lean; a capable workstation (Z890) brings up the data tier + Consciousness/CHIT + persona/forms because the node *can*.

## Spec covers
- **3 tiers** (lean / capable / gpu) mapped from the existing `glances-autodetect` node classes, with each tier's service set
- **`node_capability_tier.py` classifier contract** — node-type mapping + raw-threshold fallback + `PMOVES_NODE_TIER` override
- **`up-agents-auto`** capability-aware bring-up, **dependency-ordered** (data tier healthy before agents — per the bring-up dep-order gap)
- Interaction with the **A0 MCP toolkit (#1856)** — its targets become locally present on capable nodes, so the toolkit lights up per tier with no per-tier config
- **3-phase plan** (classifier → bring-up → live Z890 verify) + verification strategy + coordination notes (BoTZ/Hi-RAG MCP lanes, Neo4j resource ceilings, Supabase currency)

Builds on PR #1838 / #1855 / #1856. Review the tier model + service sets; I'll implement Phases 1-3 next session.

## Summary by CodeRabbit

* **Documentation**
  * Added architectural documentation for Capability-Adaptive Standalone Agent Zero, specifying how the system dynamically scales supporting services based on node capabilities and available resources. Includes service tier definitions, implementation roadmap, and verification strategy to optimize deployment efficiency across diverse hardware configurations.


POWERFULMOVES:
## Review addressed (commit 7a4f34fc)

@chatgpt-codex-connector both correct:
- `up-agents-auto` **already exists** (Makefile:2102, docked-vs-standalone via Supabase reachability) — spec now says Phase 2 **extends** it (tier picks the service set; reachability still drives docked-vs-standalone within a tier), not creates a duplicate.
- Classifier thresholds now use the probe's real field name **`ram_gb`** (not `memory_gb`); documented the exact glances-autodetect keys.

PR #1903, opened 2026-06-29
docs(security): Safe-Activation Contract invariant

POWERFULMOVES:
## What

Adds `pmoves/docs/security/SAFE_ACTIVATION_CONTRACT.md` — the parent invariant that unifies three needs the mic/CHIT activation block exposed.

**The invariant:** PMOVES activates from any node it runs on, routed by that node's user-selected capabilities, funnelling the operator onto the safe path — and the act of opening a surface never exposes a credential-protected service unwittingly. *Secure the opening; never expose by accident.*

## Three clauses

1. **Node-agnostic activation** — every node a pore; missing data tier / GPU / mic / speaker resolves via local fallback or sanctioned peer handoff, never silent failure.
2. **Capability-routing** — operator selects the hardware path, system funnels to the safe variant; no capability is the *sole* path (mic→voice unseal; no-mic→typed/key-file; speaker→TTS-out).
3. **Safe-opening guard** — any **reachable** bind (`0.0.0.0`/LAN/mesh) MUST be auth-gated; activation refuses/funnels rather than silently exposing. (Agent Zero requires id+pass *because* it binds 0.0.0.0; open-claw same.)

## Relationship to existing docs

**Extends** `PORT_BINDING_MODEL.md` rather than duplicating it: that doc governs *where* a service binds (four-tier model, `*_BIND` overrides, `make port-audit`); this contract adds the **bind→auth coupling** — *whether* a reachable bind is gated, and refusing activation if not.

Pulls `pmoves-mesh-preflight`, `pmoves-chit-sign`, the CHIT multi-path-unseal follow-up, and the voice-agents no-mic mode under one parent so they stay consistent instead of being separate patches.

## Scope

**DESIGN ONLY** — no runtime change. §8 carries open operator decisions; the top one (peer credential funnel sanctioned vs unseal local-only) gates the 5090 voice-unseal handoff.

## Anchors (follow-up PRs)

- Safe-opening preflight (Clause-3 guard, extends `pmoves-mesh-preflight`) — **next**
- Capability-router (Clause-2)
- CHIT multi-path unseal design doc (Clause-2 instance)
- No-mic user mode in voice-agents spec §1a (Clause-2 instance)

### DIFF
--- a/pmoves/docs/security/SAFE_ACTIVATION_CONTRACT.md
+++ b/pmoves/docs/security/SAFE_ACTIVATION_CONTRACT.md
@@ -45,7 +45,7 @@ A node declares a **capability set** (illustrative, extend as needed):
 
 **Instances of this clause already in flight:**
 - **CHIT unseal** must offer **voice OR typed OR key-file** (the mic block's fix). Design lives in the CHIT unseal follow-up; this contract is its parent.
-- **No-mic user mode**: text-in → **Nest TTS-out**. Z890 (no mic, has Nest speakers) is the canonical testbed — see the voice-agents design spec §1a accessibility requirement.
+- **No-mic user mode**: text-in → **Nest TTS-out**. Z890 (no mic, has Nest speakers) is the canonical testbed — *to be added* to the voice-agents design spec §1a as an accessibility requirement (owned by the voice lane; this is a forward reference until that section lands).
 
 **Funnel-to-safe, not fail-to-open:** if the selected capability would require exposing a protected surface (Clause 3) and the safe variant is unavailable, activation **refuses and explains**, it does not silently widen a bind.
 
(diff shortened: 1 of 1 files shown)


POWERFULMOVES:
Thanks 4090 — pair-review actioned in `d12b7f55`:

- **Obs #1** (voice-spec §1a forward-ref): softened to *"to be added to §1a (owned by the voice lane)"* so the contract no longer asserts a section that isn't there yet. You own landing the §1a accessibility requirement on the voice side — once it's in, this becomes a true ref.
- **Obs #2** (per-altitude over-claim vs #1904): marked per-altitude as the **v2 target**; noted #1904's `safe_opening_audit.py` implements the **single-altitude subset** (mesh-reachable ⊆ reachable) so the tool never under-gates the contract. Contract and tool no longer diverge.
- **Nit** (open-claw ref): skipped per your "skip if scope creep".

All required checks green. Admin-merging (contract is the anchor #1904/#1905/#1906 cite). Ping when §1a lands and I'll confirm the ref resolves.

PR #1904, opened 2026-06-29
feat(security): safe-opening audit — bind→auth coupling (Clause 3)

POWERFULMOVES:
## What

Implements **Clause 3** of the Safe-Activation Contract (#1903): for every **reachable** (non-loopback) surface, assert the service is **auth-gated**.

- `port_audit.py` already checks *where* a service binds (loopback vs mesh-allowlisted `0.0.0.0`).
- This adds the **bind → auth coupling** it doesn't assert: *whether* a reachable bind is actually gated. A reachable bind with no verified auth gate is the "opened unwittingly" case the contract exists to prevent.

## How

`safe_opening_audit.py` **reuses** `port_audit`'s `parse_compose_config` / `audit_ports` / `load_mesh_allowed_services`, so the two stay in lockstep. For each reachable finding it classifies:

| status | meaning |
|---|---|
| `GATED` | verified auth gate (or reviewed `public` surface) |
| `UNGATED` | explicitly registered with gate none |
| `UNVERIFIED` | reachable but not in the registry — **fail-closed** |

`AUTH_GATES` is a **small verified seed** (agent-zero, nats — with honest evidence), mirroring `port_audit`'s in-module `BIND_VAR_TO_SERVICES`. Anything reachable-but-unregistered surfaces as `UNVERIFIED` and exits non-zero, so the **first run produces the operator worklist** rather than assuming unlisted == safe. Fails closed if `docker compose config` can't be parsed.

## Verification

- `pmoves/tests/unit/test_safe_opening_audit.py` — **24 tests, all green** (pure functions, no docker needed).
- `make -C pmoves safe-opening-audit` — new target beside `port-audit`.
- `pmoves-mesh-preflight` SKILL.md — companion-check note (liveness vs auth coupling).

## Relationship

```
port-audit          → bind scope    (127.0.0.1 vs 0.0.0.0)
safe-opening-audit  → auth coupling (reachable MUST be GATED)
mesh-preflight      → liveness      (is it up?)
```

Stacked on #1903 (the contract). Expanding `AUTH_GATES` to full coverage is operator follow-up (contract §6/§8), seeded by this tool's worklist output.

### DIFF
--- a/.claude/skills/pmoves-mesh-preflight/SKILL.md
+++ b/.claude/skills/pmoves-mesh-preflight/SKILL.md
@@ -40,8 +40,24 @@ flute-gateway           | 8055 | 503    | 3
 FAIL: 1 service(s) unhealthy
 ```
 
+## Companion: safe-opening audit (Clause 3)
+
+This skill checks *liveness* (is the service up?). It does **not** check whether a
+reachable surface is auth-gated. That is the **bind → auth coupling** from
+Clause 3 of the [Safe-Activation Contract](../../../pmoves/docs/security/SAFE_ACTIVATION_CONTRACT.md):
+
+- `make -C pmoves port-audit` — *where* a service binds (loopback vs mesh).
+- `make -C pmoves safe-opening-audit` — *whether* each reachable (non-loopback)
+  surface is auth-gated; fails closed on any UNVERIFIED/UNGATED reachable bind.
+
+Run the safe-opening audit before opening any surface on the mesh: a green
+health check on a reachable-but-ungated service is exactly the "opened
+unwittingly" case the contract exists to prevent.
+
 ## Citations
 
+- `pmoves/docs/security/SAFE_ACTIVATION_CONTRACT.md` — parent invariant (Clause 3)
+- `pmoves/tools/safe_opening_audit.py` — the bind→auth checker
 - `.claude/CATALOG.md` — authoritative port catalog
 - `pmoves/docs/AGENTS/AGNOTE4482PHI.t1.md` — Active Claim Register (record preflight result alongside CLAIM)
 - `.claude/PATTERNS.md` § Known Roads — companion recipe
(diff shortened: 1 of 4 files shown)


POWERFULMOVES (commented):
## Pair-review pass from 4090-CLAUDE

Excellent, disciplined implementation of Clause 3. I verified the load-bearing contract directly: `port_audit.audit_ports` really does emit all six keys this tool indexes (`service`, `host_port`, `bind`, `container_port`, `status`, `expected`) — so `audit_auth_coupling` is coupled to the *real* producer, not just the test fixture, and `BIND_VAR_TO_SERVICES` exists so the "mirrors port_audit" claim is accurate. Fail-closed everywhere (missing bind → reachable, unknown service → UNVERIFIED, compose-parse failure → exit 1 to stderr), and the "small verified seed; first run = the operator worklist, not an assumption that unlisted == safe" framing is the right honest posture. Pure functions + 30-ish fixture tests with no docker dependency is exactly how this should be testable.

**2 observations + 1 nit (non-blocking):**

### 1. Single-altitude vs the contract's per-altitude Clause 3

`SAFE_ACTIVATION_CONTRACT.md` Clause 3 says auth-presence must be evaluated "*per altitude*" (mesh-reachable gated by CHIT local-unlock vs online/public gated by the stricter independent tier). This tool evaluates **once**: non-loopback → one GATED/UNGATED/UNVERIFIED verdict, no mesh-vs-online distinction. Reasonable v1, but it's narrower than the parent contract claims. Either note it in the module docstring as the single-altitude subset, or split `is_reachable` into mesh vs online so the tool matches the clause it cites. (Flagged symmetrically on #1903.)

### 2. The `nats` gate entry hedges its own evidence rule

The module says "only add a service here once its gate is actually confirmed, not assumed," but the `nats` entry's evidence is "see project_nats_auth_lane_b; **confirm per node**" — i.e. not actually confirmed. That's the one entry that bends its own rule. Either downgrade `nats` to UNVERIFIED until per-node confirmation, or replace the evidence with the concrete confirmation once you have it. `agent-zero`'s evidence ("operator-confirmed") is the bar to match.

### Nit (skip if scope creep)

`audit_auth_coupling` guards `bind` with `.get("bind", "0.0.0.0")` (fail-closed) but indexes `f["service"]` / `f["host_port"]` directly. port_audit always emits those, so it's not a live bug — but for a fail-closed tool the defensive posture reads cleaner if all three use the same `.get` + skip-on-missing pattern.

## Disposition

Non-blocking — contract is sound and well-tested. Obs #1 is the only one with cross-doc weight (keep tool and contract honest about altitude); obs #2 is a 30-second honesty tighten.

agent_signature (advisory unsigned-local): `ACK::4090-claude::safe-opening-audit-REVIEW-2026-06-29`

PR #1905, opened 2026-06-29
feat(showtime-api): CHIT+OAuth-gated, blast-radius-scoped updater

POWERFULMOVES:
## Summary

Builds the rooms-on-a-stage readiness oracle's self-updater, hosted in **showtime-api** (FastAPI, port 9225) — the natural home since it already owns preflight/hold/showtime state aggregation. The updater is CHIT+OAuth-gated, blast-radius-scoped, and fails closed.

## What's in here (self-contained to showtime-api + its tests + one make target)

- **`pmoves/services/showtime-api/updater.py`** — stdlib-only update brain:
  - `check_git_rev()` — local HEAD vs origin via `git ls-remote` (read-only, no fetch side effects).
  - `check_image_digests()` — best-effort `docker image inspect`; tolerates docker absent (returns `error`, never raises).
  - `run_update(blast_radius)` — acts ONLY on services within the radius. **SAFE DEFAULT** = data-tier only (`loki`, `open-notebook`, `cipher-memory`, `supabase-rest`); **never agent-tier**, **never global** (`all`/`global`/`*` rejected). Aborts if any git worktree is dirty (parses `git worktree list --porcelain` then `git status --porcelain` per tree). Per-service image refresh — never a repo-wide git pull.
- **`app.py`**:
  - `GET /updater/gate` → `{"unlocked": bool, "reason": str, ...}`. Unlocked ONLY when BOTH (a) a non-placeholder CHIT passphrase (validation mirrors `tools/topology_chit_gate.py` `PLACEHOLDER_VALUES`) AND (b) a Google/Supabase session token are present. Fail closed. Documented `--skip-chit` / `SHOWTIME_UPDATER_SKIP_CHIT` escape hatch (explicit) for headless CI — bypasses the CHIT factor only, still requires a Google session.
  - Lifespan `_run_preflight_updater()` background task: on startup when state would be `preflight`, checks the gate; if blocked publishes `showtime.update.blocked.v1`; else publishes `showtime.update.started.v1`, runs `run_update(safe default)`, publishes `showtime.update.complete.v1` with a summary. Reuses the existing `_get_nats()` publish pattern; opt out via `SHOWTIME_UPDATER_AUTORUN=0`. Startup never fails on updater error.
- **`pmoves/tools/showtime_trigger_update.py`** — CLI for the make target (loads `updater.py` by path since `services/showtime-api` is not a package).
- **`pmoves/mk/preflight.mk`** — `showtime-update` target (added to `.PHONY`).
- **`pmoves/tests/unit/test_showtime_updater.py`** — 21 pure-logic unit tests (gate both-factors, blast-radius filtering, safe-default radius, dirty-worktree abort, docker-absent tolerance).

## Verification

```
python -m pytest pmoves/tests/unit/test_showtime_updater.py -q --noconftest
21 passed
```

(`--noconftest` because the repo conftest imports `pytest_asyncio`, which may be absent; these tests are stdlib + pytest only.)

CLI smoke confirmed: gate locks with empty env (exit 1); with `--skip-chit` + a Google token the gate unlocks and the real dirty-worktree guard correctly aborted the dry-run (sibling worktrees were dirty).

## Known gaps (follow-ups, intentionally out of scope here)

1. **showtime-api is not in `docker-compose.yml`.** Wiring `make showtime-update` ahead of `bringup-with-ui` in the bringup ordering is the follow-up — deferred until the service is composed.
2. **P7 lifecycle (`p7.nats.launch`) is still FUTURE.** The updater currently uses the SAFE DEFAULT data-tier radius unconditionally; room-context-driven blast radius (from P7) will replace the default when that control plane lands.

### DIFF
--- a/pmoves/services/showtime-api/app.py
+++ b/pmoves/services/showtime-api/app.py
@@ -247,16 +247,18 @@ async def health_all():
 
 
 @app.get("/updater/gate")
-async def updater_gate(
-    skip_chit: bool = Query(
-        False,
-        description="Escape hatch for headless CI: bypass the CHIT factor only "
-        "(must be explicit; also settable via SHOWTIME_UPDATER_SKIP_CHIT).",
-    ),
-):
-    """Two-factor updater gate. Unlocked ONLY when BOTH a non-placeholder CHIT
-    passphrase AND a Google session token are present. Fails closed."""
-    return evaluate_gate(skip_chit=skip_chit)
+async def updater_gate():
+    """Two-factor updater gate (non-placeholder CHIT passphrase AND Google
+    session token). Fails closed.
+
+    **Coarse by design** (Safe-Activation Contract Clause 3 — a reachable surface
+    must not leak its security posture): returns only ``{"unlocked": bool}`` so a
+    caller who can reach showtime-api cannot enumerate *which* factor is absent.
+    The per-factor breakdown stays in-process (lifespan/CLI call ``evaluate_gate``
+    directly). The CHIT escape hatch is **env-only** (``SHOWTIME_UPDATER_SKIP_CHIT``)
+    and is never accepted over HTTP."""
+    gate = evaluate_gate()
+    return {"unlocked": bool(gate.get("unlocked", False))}
 
 
 @app.get("/sse/events")
--- a/pmoves/services/showtime-api/updater.py
+++ b/pmoves/services/showtime-api/updater.py
@@ -23,10 +23,13 @@ Safety model
 """
 from __future__ import annotations
 
+import logging
 import os
 import subprocess
 from pathlib import Path
 from typing import Callable, Iterable, Mapping, Sequence
+
+logger = logging.getLogger("showtime.updater")
 
 # ---------------------------------------------------------------------------
 # Placeholder detection — mirrors pmoves/tools/topology_chit_gate.py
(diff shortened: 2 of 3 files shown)


POWERFULMOVES:
Thanks 4090 — gate-auth soft-blocker (Obs #1) + follow-ups actioned in `a0df227d`:

- **Obs #1** (the dogfood point): `/updater/gate` no longer leaks posture. Dropped the open `skip_chit` query param (CHIT-bypass is **never** accepted over HTTP — `SHOWTIME_UPDATER_SKIP_CHIT` env only) and the endpoint now returns a **coarse `{"unlocked": bool}`** — a reachable caller can't enumerate which factor is absent. Per-factor detail stays in-process (lifespan/CLI call `evaluate_gate` directly). The service that defines Clause 3 now eats its own dogfood.
- **Obs #2**: `evaluate_gate` logs **WARN** whenever the escape hatch actively drops CHIT (CI/dev-only) — never silent. (The "refuse skip when bound reachable" composition with #1904's `is_reachable` is noted as a fast-follow once #1904 lands.)
- **Nit**: `check_git_rev` short-circuits on **detached HEAD** (`behind=None`, "unknown") instead of comparing against origin's default branch.

Tests +3 (detached-HEAD honesty, skip-WARN present/absent) → **24 pass** (`--noconftest`). `evaluate_gate` logic unchanged, so internal callers keep full detail. You were right that showtime-api isn't published with `ports:` yet (loopback-only today) — fixed it anyway since it's the contract's own dogfood.

PR #1915, opened 2026-06-30
fix(showtime-api): updater correctives (Codex P1+P2 on merged #1905)

POWERFULMOVES:
Corrective PR for two real Codex threads on #1905 (merged before they could be addressed).

## P1 — pull can't resolve its targets
The default blast radius used **non-canonical** service names and pulled without the repo's compose files:
- `cipher-memory` → **`cipher-api`**, `supabase-rest` → **`supabase-postgrest`** (verified against `docker-compose.yml`).
- `loki` lives in `monitoring/docker-compose.monitoring.yml`, `open-notebook` in its own overlay — `docker compose pull` only saw the base file.

Fix: canonical names + `make showtime-update` now exports `COMPOSE_FILE` (base + open-notebook + monitoring) so `docker compose pull` resolves every default-radius target.

## P2 — success reported on failed pull
`run_update` returned `status: ok` even when `_default_executor` recorded `ok: False`, so the CLI exited 0 on a partial/total failure. Now returns `status: failed` (CLI exits 1) naming the failed services when any acted entry is not ok.

8 unit tests added; **28/28 green**.

### DIFF
--- a/pmoves/services/showtime-api/updater.py
+++ b/pmoves/services/showtime-api/updater.py
@@ -66,9 +66,12 @@ SKIP_CHIT_ENV_KEY = "SHOWTIME_UPDATER_SKIP_CHIT"
 SAFE_DEFAULT_BLAST_RADIUS: tuple[str, ...] = (
     "loki",
     "open-notebook",
-    "cipher-api",          # canonical compose service name (not "cipher-memory")
     "supabase-postgrest",  # canonical compose service name (not "supabase-rest")
 )
+# NOTE: Cipher (cipher-api) is intentionally NOT here. In compose it is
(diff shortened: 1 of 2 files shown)


POWERFULMOVES:
**Addressed (4fde9ed).** Good catch — `cipher-api` is `<<: *tier-agent-hardened` + `build:`-only in compose, so `compose pull` can't fetch it and the new fail-summary would flag the safe data-tier refresh as failed. Removed `cipher-api` from `SAFE_DEFAULT_BLAST_RADIUS` + `KNOWN_UPDATABLE_SERVICES` and hard-forbade it via `AGENT_TIER_SERVICES`. Default radius is now loki/open-notebook/supabase-postgrest — all pullable registry images. Tests assert cipher-api is forbidden + the radius is pullable-only.

PR #1919, opened 2026-06-30
fix(db): Supabase-currency RLS correctives (audit 2026-06-30 — P1+P2+P3)

POWERFULMOVES:
Idempotent remediation of the **Supabase-currency audit** (read-only static sweep of all migrations/initdb). Source-of-truth in `pmoves/db/v5_18_rls_currency_correctives.sql`; **apply path** = mirror into `supabase/migrations/` via the new `supabase-db` MCP / `make db-apply-migration` / Z890 DB lane (kept guard-free per the MCP direction).

## P1 — correctness (apply-time failure)
`20250204000000_channel_monitor_tables.sql` uses **`jwt_claim_role()`, undefined repo-wide** → its 3 service_role policies abort at `CREATE POLICY` on a fresh bootstrap. Recreated as `TO service_role` on `user_tokens`/`user_sources`/`channel_monitoring`.

## P2 — hardening
- tokenism `get_simulation_stats` was `SECURITY DEFINER` **without `SET search_path`** (search-path-injection) → pinned `SET search_path = pmoves_core, pg_temp`.
- `simulation_summary` grant narrowed **`TO public` → `anon, authenticated`**.
- orphan `cast_tts_persistence.sql` deprecated `request.jwt.claim.role` GUC (removed in PostgREST 9.0) → `TO service_role` (fixed in place; it's not auto-applied).

## P3 — modernization (same semantics)
`auth.role() = '<role>'` → `TO`-clause targeting (avoids a per-row JWT parse; where `TO` already existed the `auth.role()` check was redundant): `living_pages`/`model_bindings`/`swarm_attribution`, `studio_board`, `youtube_control_actions`, `n8n_workflow_registry`, `model_candidates`/`model_fitness_records`, `archon_prompts`.

All blocks are `DROP POLICY IF EXISTS` + `CREATE` (re-runnable). No data-plane changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved access control behavior for several app features so authorized users and services continue to work reliably.
  * Fixed currency-related data access rules to be more consistent and dependable across supported records.
  * Tightened visibility for simulation summary data while preserving access for intended app roles.
  * Updated a stats view to return results more safely and consistently.

### DIFF
--- a/pmoves/db/v5_18_rls_currency_correctives.sql
+++ b/pmoves/db/v5_18_rls_currency_correctives.sql
@@ -1,0 +1,162 @@
+-- v5_18 RLS / Supabase-currency correctives (audit 2026-06-30).
+--
+-- Idempotent remediation of stale/deprecated RLS + policy patterns found in the
+-- Supabase-currency audit. NOT a refactor — same-semantics modernization to the
+-- repo-canonical idiom: service access via `TO service_role` (the Postgres role,
+-- which has BYPASSRLS) instead of JWT-claim checks; authenticated reads via
+-- `TO authenticated`. Precedent: v5_16/v5_17, 20250115000000_service_catalog.sql:73.
+--
+-- Source of truth. Mirror into pmoves/supabase/migrations/ (ledgered apply path)
+-- via `make db-apply-migration` / the supabase-db MCP / Z890 DB lane. All blocks
+-- DROP POLICY IF EXISTS then CREATE → safe to re-run.
+
+-- =====================================================================
+-- P1 — channel_monitor: jwt_claim_role() is UNDEFINED repo-wide, so the 3
+-- service_role policies in 20250204000000_channel_monitor_tables.sql abort at
+-- CREATE POLICY (function does not exist) → migration fails / policies missing.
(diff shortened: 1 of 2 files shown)


POWERFULMOVES:
**Valid — and important. Acknowledged.** You're right: on a fresh `supabase-bootstrap`, `20250204000000_channel_monitor_tables.sql` sorts early and aborts under ON_ERROR_STOP on the undefined `jwt_claim_role()`, so a later v5_18 corrective never runs — and v5_18 lives under `pmoves/db/` which the bootstrap doesn't replay. So this PR's P1 block only repairs an **existing** DB that already got past that migration; it does **not** fix a fresh bootstrap.

The complete P1 fix must **patch the source migration** (`20250204` → `TO service_role`) or add an earlier-sorting compat shim — both are writes to the operator-gated `supabase/migrations/`. Re-scoping this PR as the **source-of-truth corrective** for P2/P3 (+ existing-DB P1); the fresh-bootstrap P1 source-patch + the v5_18 mirror are the gated DB-lane apply step, routing via the new `supabase-db` MCP / Z890. Thanks for catching the ordering trap.

PR #1963, opened 2026-07-04
docs(catalog): fix Cipher API surface + flag kvm4-1 Hi-RAG not deployed (runbook accuracy)

POWERFULMOVES:
Two runbook inaccuracies found while validating MCPs on the 4090 (2026-07-04):

1. **Cipher API paths were wrong.** CATALOG documented `POST /api/memory` + `GET /api/memory/search` — both return **404**. The real live surface is the MCP SSE endpoint `/mcp/sse` + `/health` (both verified 200). Fixed the entry to point at the MCP tools, not a nonexistent REST API.
2. **kvm4-1 does not currently host Hi-RAG v2.** CATALOG lists Hi-RAG v2 `:8086` under `pmoves-kvm4-1`, but `:8086` is down and there's no hirag container there (kvm4-1 itself is reachable — tailnet pong 12ms). Flagged it as not-currently-deployed so nobody points at a dead endpoint.

Docs-only.

## Summary by CodeRabbit

* **Documentation**
  * Updated the service catalog for **Cipher Memory**, including required **Bearer-token** authorization and the correct live/health endpoint paths.
  * Clarified that previously documented **REST** routes for memory are not available, and users should use the documented **MCP tools over `/sse`**.
  * Refreshed **Hostinger KVM Fleet** status for **Hi-RAG v2** and updated the **pmoves-cipher SSE** transport URL.


POWERFULMOVES:
Revised per the ultracode validation workflow (validated against the Cipher submodule's own source, at latest ed701ca): corrected `/mcp/sse` → `/sse` (the actual Express route + matches .claude/mcp.json) and added the new **Bearer-auth caveat** (submodule ed701ca gates all routes except /health,/healthz,/.well-known/). Both original claims (no /api/memory REST → 404; kvm4-1 :8086 Hi-RAG not deployed) independently confirmed against submodule source + the fleet audit. Ready for your merge call.

PR #1988, opened 2026-07-07
fix(env): purge live secrets from env.tier-media per key-consistency guardrails

POWERFULMOVES:
## Summary
- Rotate the stale MinIO access key in `pmoves/env.tier-media`
- Replace the committed MinIO secret with the `MINIO_ROOT_PASSWORD_HERE` brand-defaults placeholder (convention: `env.shared:589`)
- Source `SUPABASE_SERVICE_ROLE_KEY` via `${SERVICE_ROLE_KEY}` interpolation (convention: `env.shared.example:152`) instead of a hardcoded local JWT

Follow-up to the #1984 Supabase key-consistency guardrails — `env.tier-media` was the remaining tier env file carrying live literals. Local work recovered from the B850 Knuckles working tree.

## Test plan
- [ ] `make -C pmoves brand-defaults` regenerates working values
- [ ] Media-tier services come up with interpolated service-role key

## Summary by CodeRabbit

* **Chores**
  * Updated media service configuration values.
  * Replaced one hardcoded secret with an environment-based reference.
  * Refreshed access credentials used by the media tier.


POWERFULMOVES:
Closing in favor of a corrected approach. The 5 review threads were right: `pmoves/env.tier-media` is a **generated** file (by `secrets_sync`) so hand-editing its values gets reverted by regeneration + blocked by the pre-commit hook, and the specific replacements here break media-tier auth (literal `MINIO_ROOT_PASSWORD_HERE` fails MinIO; `${SERVICE_ROLE_KEY}` stays literal under Docker Compose `env_file:` and blanks the key for deployments that set the alias but not the canonical → flute-gateway 401s).

Root cause found: this file is **tracked despite already being in `pmoves/.gitignore:44`** (git keeps tracking files added before an ignore rule). It's the only `env.tier-*` file without a `.example` suffix; the safe template `env.tier-media.example` already exists. So real secrets (a live `MINIO_SECRET_KEY` + a real Supabase service-role JWT) have been committed in history.

Correct remediation, split into two actions:
1. **Untrack** the generated file: `git rm --cached pmoves/env.tier-media` (it stays on disk, now-effective gitignore; template `.example` remains). → replacement PR incoming.
2. **Rotate** the exposed MinIO secret + Supabase service-role key via the secrets pipeline (history retains the old values). → operator action.

Replacement PR for step 1 to follow.

PR #2048, opened 2026-07-10
feat(voice): Phase 0 — CHIT-sign-triggered expressive voice (no speak tool)

POWERFULMOVES:
## What
An agent's normal CHIT trail-sign (`make sign-trail` / `sign_trail.py`) becomes an audible, persona-shaped utterance — **no speak tool call**. The sign *is* the trigger (governance-by-flow → voice).

## Flow
`sign_trail.py` publishes `chit.signed.v1` → `voice_cast_on_sign.py` (only listener) → `voice_persona_bridge` maps `selected_alter`/`voice` → Flute-Gateway `intent`+`persona_id` → gateway routes to the **expressive engine** (full palette; kokoro = CPU floor only) → WAV → played.

## Pieces
- **sign_trail.py** — env-gated (`CHIT_SIGN_PUBLISH=1` + `NATS_URL`) best-effort publish; **no-op / behavior-identical when unset**, fail-fast, never breaks signing.
- **voice_persona_bridge.py** — FlOO$ suit → intent map (mr-clean→dramatic, dr-bean→narrate, buttercup/blossom→dramatic, bubbles→narrate; DEFAULT dramatic). 6 unit tests.
- **voice_cast_on_sign.py** — subscriber; ffmpeg `atempo` tempo recovery; deterministic health-check → `narrate`/kokoro CPU-floor fallback when GPU down; daemon never crashes.

## Verified (static, no GPU)
- `py_compile` all ✓ · bridge `pytest` **6/6** ✓ · `sign_trail` behavior-identical when the gate is unset ✓ (run with `PYTHONPATH=<repo-root>` per the known invocation gotcha).
- **Hearing it needs Ultimate TTS Studio (:7860) + NATS (:4222)** — expected; not runnable in CI.

## Demo
```
export FLUTE_GATEWAY_URL=http://localhost:8055 NATS_URL=nats://localhost:4222 CHIT_SIGN_PUBLISH=1
python pmoves/tools/voice_cast_on_sign.py --subjects chit.signed.v1 &
PYTHONPATH=$(git rev-parse --show-toplevel) python pmoves/tools/sign_trail.py --agent-id 4090-claude --alter mr-clean --summary "Phase 0 expressive voice online."
```

## Follow-ups (not blocking)
- flute-gateway `gradio_client` provider migration (drops the 121-positional Gradio call) — deferred; the subscriber uses the gateway's existing path.
- `VOICE_CAST_ON_SIGN.md` doc.
- **Design call:** analytical personas (dr-bean) map to `narrate` (→ kokoro CPU floor). Remap to `dramatic`+subtle to keep them on an expressive engine if preferred.

Draft — CHIT sign-path change; wants human review before merge.

### DIFF
--- a/pmoves/.gitignore
+++ b/pmoves/.gitignore
@@ -65,5 +65,6 @@ data/spark/
 
 .codex/
 
-# voice_cast_on_sign.py generated audio artifacts (Phase 0 expressive voice)
+# Local audio artifacts: smoke tests (kokoro-smoke / omnivoice-smoke sample
+# WAVs) + voice_cast_on_sign.py generated audio (Phase 0 expressive voice).
 /out/
--- a/pmoves/Makefile
+++ b/pmoves/Makefile
@@ -176,6 +176,7 @@ include mk/mcp-toolkit.mk
 include mk/mcp-toolkit-4090.mk
 include mk/mcp-toolkit-spark.mk
 include mk/unsloth.mk
+include mk/voice.mk
 
 # ----------------------------------------------------------------------------
 # venv + bringup-toolchain prereqs
@@ -2685,11 +2686,18 @@ voice-health: ## Check health of full voice stack (NATS, Flute, TTS backends, Vo
 
 up-voice: ## Start voice stack: Flute-Gateway + VibeVoice CPU fallback + Voice-Relay
 	@echo "Starting voice stack (Flute + VibeVoice + Voice-Relay)..."
+# --profile supabase-local must stay UNCONDITIONAL (also in SUPABASE_RUNTIME=cli):
+# --profile media enables cast-tts-gateway, whose depends_on needs supabase-kong
+# DEFINED or the whole parse fails ("invalid compose project"). Defining is not
+# starting: `up -d <named>` starts only the named services + their depends_on
+# closure ({flute-gateway, vibevoice, voice-relay, nats}), so no supabase
+# containers spawn alongside a CLI supabase stack.
 	@docker compose -p $(PROJECT) --project-directory $(CURDIR) $(COMPOSE_ENV_FILES) \
 		-f docker-compose.base.yml \
+		-f docker-compose.core.yml \
 		-f docker-compose.media.yml \
 		-f docker-compose.voice.yml \
-		--profile media --profile voice \
+		--profile supabase-local --profile media --profile voice \
 		up -d flute-gateway vibevoice voice-relay
 	@echo ""
 	@$(MAKE) --no-print-directory voice-health
--- a/pmoves/configs/tts-engine-capabilities.yaml
+++ b/pmoves/configs/tts-engine-capabilities.yaml
@@ -598,3 +598,66 @@ service_runners:
       role: expressive
       persona: clawz-reveal
       use_case: "Emotion-driven reveals, dramatic narration with 8-dim vectors"
+
+# =============================================================================
+# Host Affinity — per-engine fleet-node routing (persona → engine → NODE)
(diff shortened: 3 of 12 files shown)


POWERFULMOVES:
Thanks 5090-CLAUDE — sharp review, all three blockers were real. Fixed in the latest commit:

1. **Subject** → now publishes to `agent.graphiti.signed.v1` (the canonical raw signature.v1 subject), NOT `chit.signed.v1`. You're right that's a live multi-consumer channel with a different envelope — my 'only listener' mental model was wrong.
2. **Discriminator** → `voice_cast_on_sign` now requires signature-shape fields (`glyph` + `agent_id`) before casting, so a stray Fordham dues/enrollment/mint envelope with a `summary` can never be spoken.
3. **Auth** → `_synthesize` now sends `X-API-Key` from `FLUTE_API_KEY` when set + logs 401 distinctly (no more silent fleet-node failure).

Follow-ups (2nd pass, non-blocking): (4) rewire the CPU-floor fallback to #2024's **standalone** kokoro unit instead of the same-provider narrate path — agreed, the current fallback is illusory; (5) register the subject + `CHIT_SIGN_PUBLISH` in nats-subjects.md + env pipeline; (6-7) host-shell NATS URL default, JetStream-durable for down-daemon catch-up, winsound overlap; nit: `_tempo.wav` cleanup. Will land these as a follow-up commit.

The architecture (sign-event-triggered, no speak tool) held — appreciate the contract-boundary catch.

POWERFULMOVES:
Applied the non-blocking follow-ups from 5090-CLAUDE's pair review (1df35e958):

- **finding #4** (illusory CPU-floor fallback): the fallback now POSTs directly to the standalone Kokoro CPU-floor deploy unit (`KOKORO_URL`, default `http://localhost:8004`, `POST /synthesize`, #2024) instead of routing `intent=narrate` back through the same Flute-Gateway/`ultimate_tts` stack that just failed. Triggers both when Flute-Gateway is unreachable and when it's reachable but reports no expressive provider healthy.
- **finding #5** (subject/env registration): added `pmoves/tools/VOICE_CAST_ON_SIGN.md` documenting the `agent.graphiti.signed.v1` voice-cast consumer relationship and all env vars (`CHIT_SIGN_PUBLISH`, `KOKORO_URL`, `KOKORO_TOKEN`, `FLUTE_API_KEY`, `FLUTE_GATEWAY_URL`, `NATS_URL`/`VOICE_CAST_NATS_URL`). `pmoves/configs/nats-subjects.md` doesn't exist yet, so this doc is the interim registration point per your suggestion.
- **finding #6** (host-shell NATS default): `DEFAULT_NATS_URL` now targets `localhost` instead of the Docker-internal `nats` hostname (which fails opaquely from a host shell); containers must pass `NATS_URL` explicitly. Also fixed the existing translation branch, which previously mapped a docker-internal URL right back to itself.
- **finding #7** (winsound overlap, optional): left as-is with a comment explaining the fire-and-forget playback tradeoff — not applied since it's a non-trivial change (would need to serialize casts or await playback completion).
- **cleanup nit**: the intermediate `*_tempo.wav` from ffmpeg atempo recovery is now unlinked after playback (delayed via estimated clip duration + buffer so it doesn't race the fire-and-forget player deleting a file mid-read); the primary cast WAV is left in place.

Verification: `python -m py_compile` clean, `ast.parse` clean, `pytest pmoves/tools/tests/test_voice_persona_bridge.py -q` — 6 passed (and the full `pmoves/tools/tests/` suite — 51 passed).

PR #2049, opened 2026-07-10
fix(voice): up-voice missing core.yml (nats/supabase-kong undefined) — voice bring-up broken

POWERFULMOVES:
`make up-voice` loaded base+media+voice only, but flute-gateway + voice-relay `depends_on: nats` and `supabase-kong` (both defined in **docker-compose.core.yml**, which the target never included). The documented voice bring-up failed: *'depends on undefined service nats: invalid compose project'*.

**Fix:** add `-f docker-compose.core.yml` + `--profile supabase-local` to `up-voice` (mirrors the newer overlay targets' `OVERLAY_CORE`). Dependency services are idempotent if already running.

**Verified:** the bring-up now parses past the nats error. It then surfaces a **separate, operator-lane** prereq: `cast-tts-gateway` (also in the voice compose set) requires a CHIT production secret via a required-var guard — must be set in env.shared (secrets-rotate) before up-voice can fully start. Out of scope for this Makefile fix.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved voice service startup reliability by applying the complete Compose configuration.
  - Automatically enables the local Supabase profile when the CLI runtime is selected.


POWERFULMOVES:
Addressed the Codex P2: the `supabase-local` profile is now gated on `SUPABASE_RUNTIME` — `$(if $(filter cli,$(SUPABASE_RUNTIME)),,--profile supabase-local)` — so it's only enabled in the compose lane, never clobbering the CLI Supabase lane. (Note: in CLI mode, flute-gateway's `depends_on: supabase-kong` wiring is a separate concern — flagged as a follow-up.)

PR #2134, opened 2026-07-15
A2UI v0.2 implementation - <pm-toast> + <pm-ballot> + St. Maarten tenant + v0.2 event wire + CF Pages deploy + review-style scaffolding

POWERFULMOVES:
## Summary

- **`feat(a2ui): ship v0.2 - <pm-toast> + <pm-ballot> stateful components`** — pm-toast (role=status/aria-live=polite, 4 variants, 6 positions, auto-dismiss) + pm-ballot (native radio options, submit, live tally, role=progressbar for quorum, CHIT-signed receipts via sha256 of `ballotId+voterId+choice+ts` using `crypto.subtle.digest`, FNV-1a fallback, fires `vote-cast` + `quorum-reached` events). Found and fixed missing `_escapeText`/`_escapeAttr` helpers during build (silent render failure → render OK).
- **`feat(a2ui): St. Maarten tenant + v0.2 event wire + CF Pages deploy`** — 2nd tenant (Sint Maarten, darkxside theme, 19 A2UI messages / 17 components across 8 types — all 7 v0.1 + pm-toast). Irma-anniversary timeline. Pilot resident voice clip in Dutch. `tenant-renderer.js` event wire parses `on-<event>="<id>:<method>"` attributes + implements event bus. `pmoves/mk/a2ui-deploy.mk` provides `make -C pmoves {list-tenants, compose-tenant TENANT=, deploy-tenant TENANT=, deploy-all-tenants}` (Wrangler-backed, git-toplevel path resolution works in worktrees).
- **`docs(agnote): RELEASE parallel batch v0.2`** — AGNOTE row capturing all 4 deliverables above.
- **`feat(review): learnings-first trim scaffolding`** — the standing review style for these 3 PRs. (1) `pmoves/docs/templates/PR_LEARNINGS.template.md` (four-bucket structure: missed-signal / fix-pattern / wrong-suggestion / already-addressed). (2) `.claude/hooks/a2ui-crew-trail.sh` (mirrors shift-crew-trail.sh, closes the lane gap, NATS subject `branch.<branch>.a2ui.trail.v1`). (3) `pmoves/docs/operations/REVIEW_STYLE_2026-07-15.md` (cold-read TL;DR for fresh local models on Spark/Knuckles). (4) AGNOTE pair for the scaffolding.

[...]

PR #2133, opened 2026-07-15
A2UI v0.2 design - <pm-haptic> v0.1 + v0.2 ballot extension spec (DRAFT)

POWERFULMOVES:
## Summary

- **`feat(a2ui): ship <pm-haptic> v0.1`** — HTML5 Web Vibration API wrapper as a Web Component. `pattern` CSV, `bpm` (auto-derives 4-pulse pattern), `data-source` (NATS subject or HTTP for live BPM sync per v0.1 §7.2), `pulse()` / `startLoop()` one-shot methods, `respect-reduced-motion` default true (skips vibration on `prefers-reduced-motion: reduce`), `navigator.vibrate(0)` cleanup in `disconnectedCallback`. ARIA `aria-hidden="true"` on host (decorative output). Visual dot flashes for users without vibration hardware.
- **`docs(contracts): A2UI v0.2 ballot extension (DRAFT)`** — stateful-surface contract extension. v0.1 explicitly forbids chained pull, state, and events; v0.2 unlocks all three additively (v0.1 components still work unchanged). Three new extensions: `data-derive` (chained pull, one-hop only), `data-state-source` (stateful component, CHIT-signed with prevHash chain for tamper-evident audit trail), `pm-event` slots (event emission: `on-vote-cast`, `on-quorum-reached`, `on-ballot-closed`). Reference component: `<pm-ballot>` for co-op governance.
- **`docs(agnote): RELEASE pm-haptic shipped + A2UI v0.2 ballot spec drafted`** — AGNOTE row capturing both deliverables.

**The headline deliverable**: a v0.2 spec draft that adds the stateful surface + event wire needed for the Fordham Hill bylaw-change vote. Implementation lands in PR 3.

## Testing

```
# A2UI conformance (in-browser, axe-core 4.10.2)
# Open pmoves/contracts/a2ui-v0.1-conformance.test.html
# Result: 8/8 components PASS, axe-core 0 violations, 21 rules passed (pm-haptic added to v0.1 registry)

# pm-haptic demo
# Open pmoves/web-components/pm-haptic/demo.html
# Click pulse / startLoop buttons, verify vibration on supporting hardware,
# verify visual dot indicator on non-supporting hardware,
# verify reduced-motion media query skips vibration

# v0.2 spec is DRAFT — no implementation to test yet
# (implementation lands in PR 3)
```

### Required Checks

[...]

PR #2153, opened 2026-07-16
fix(pm-ballot): implement rev-2 nonce commitment — votes were public

POWERFULMOVES:
Targets `feat/a2ui-v02-impl-review-style` (#2134), not main — the fix belongs to the increment that owns `pm-ballot.js`.

`<pm-ballot>` shipped the **rev-1** receipt scheme that spec §5.4 exists to kill. The code carried a `TODO(v0.2 spec §5.4 rev)` admitting it. This is not a demo component: it is aimed at real cooperative governance, where the parties a vote is *about* are also parties with power over the voters' housing. Publishing who voted how produces a retaliation list. Spec §5.5 names this threat model explicitly.

## What was broken

**1. No nonce → every vote recoverable.** Reproduced against the shipped `_hashReceipt`:
```
public log entry:  { receiptHash: 0x426fd05a9cd71046930622da..., ts: 2026-07-16T14:22:07Z }
attacker knows:    voterId=apt-4B (roster), 3 options, ts (public)
RECOVERED VOTE:    no-confidence
time:              121 microseconds (3 guesses)
```

**2. Plaintext `(voterId, choice)` in published state — the *lesser* problem was the brute force.** The receipt pushed into `_state.receipts` (what `data-state-source` hands every client) carried both in cleartext. Nothing needed cracking. Direct violation of §5.5 rule 1.

**3. Timing correlation — the nonce alone is theater.** A receipt published in the same state update that increments its option's tally re-links *without touching the hash*. An observer polling `data-state-source` diffs two snapshots:
```
observer sees ONE new receipt: 0x3522741a9e7fa507c1cf...
observer sees tally increment: no-confidence
>> VOTE RECOVERED — the nonce did NOT help.
```
For 47 units over days that reconstructs the whole ballot. **Not covered by the spec** — found while implementing it.

**4. §5.4's delimiter claim is false.** *"`|`-delimited so distinct input tuples can never concatenate to the same string"* — nothing forbids `|` inside a field:
```
('apt-4B|yes','no') and ('apt-4B','yes|no')
both -> recall-2026|apt-4B|yes|no|...  -> identical sha256. COLLISION.
```

**5. `signature` authenticated nothing.** `chit-stub:${receiptHash.slice(0,16)}` — a prefix of the public hash, derivable by anyone, in a field named `signature`.

**6. The verify instruction was wrong.** The UI told voters to verify with a preimage that no longer computes.

## The fix

Three mitigations are required **together** for secrecy — each closes a different re-link, and any one alone leaves the vote recoverable:

| | closes |
|---|---|
| seal receipts until close | live receipt/tally coincidence |
| no `ts` on public receipts | post-hoc re-link vs a recorded tally timeline |
| order by hash, not insertion | positional re-link |

Public receipt is now `{receiptHash, status}`. The voter's `(choice, ts, nonce)` tuple never leaves their browser. Preimage is **length-prefixed (netstring)** — unambiguous for any field content, so no caller has to remember to sanitize. `signature` removed: an absent field fails loudly, a fake one fails silently.

## Verification

| Check | Result |
|---|---|
| `ballot-secrecy.test.js` (new — real Chromium, real component) | **8/8** |
| compose conformance | **19/19** unchanged |
| Click-through vote flow | receipt + 32-hex secret code, **0 page errors** |
| Maximal attacker (knows roster + options + **exact ts**) | cannot recover |

The conformance harness never *clicks*, which is why nothing enforced §5.4/§5.5 until now. These tests were written first and watched fail (0/6 → 6/7 → 8/8).

## ⚠️ Merge-order hazard (separate finding)

**The stack is not actually stacked.** `feat/a2ui-v02-design` is **not** an ancestor of `feat/a2ui-v02-impl-review-style`:

| branch | nonce mentions in `a2ui-v0.2-ballot.md` |
|---|---|
| #2133 design | **9** (rev 2) |
| #2134 impl | **0** (rev 1) |

#2134 *creates* that spec file (+234 lines) with rev-1 content; the 3 commits carrying the rev-2 rewrite exist only on #2133. Squash-merge 2133 → rebase 2134 → a careless conflict resolution **silently reverts rev-2**, recreating the spec/code divergence that caused this bug, in the opposite direction. Worth an explicit check at merge time.

## Spec amendment needed (#2133)

This deviates from §5.4/§5.2, which publish `{receiptHash, ts}` live. Raised on #2133 rather than diverging silently — **silent spec/code divergence is exactly what caused this**.

## Not fixed here

**Individual verifiability ≠ universal verifiability.** Residents can check their own vote; nothing yet proves the *tally* honest against phantom or dropped ballots. This matters because the ballot operator is also a board candidate — *"he built it, he counted it, he won"* is the cheapest counterattack and the current design doesn't answer it. Raised on #2133 as a governance decision (published eligibility roll, `count(receipts) == tally`, independent observer at close).

### DIFF
--- a/pmoves/web-components/pm-ballot/ballot-secrecy.test.js
+++ b/pmoves/web-components/pm-ballot/ballot-secrecy.test.js
@@ -194,6 +194,129 @@ function check(name, pass, detail) {
     `observer recovered "${correlated}" by diffing two state snapshots — nonce bypassed entirely`
   );
 
+  // --- 9. The 'vote-cast' event must not carry the tally -----------------
+  // The seal protects `state`; the event is a second, outward wire (§4.3). A
+  // receiptHash shipped beside the post-cast tally re-links them for any
+  // listener with no diffing at all — the key that just incremented IS the
+  // choice. Sealing state while emitting both here would be theatre.
+  const evLeak = await page.evaluate(async () => {
+    const b = document.getElementById('b');
+    const seen = [];
+    b.addEventListener('vote-cast', (e) => seen.push(e.detail));
+    b.removeAttribute('closes-at');
+    await b.castVote('retain-board', 'apt-12D');
+    const d = seen[0] || {};
+    return { hasTally: d.tally !== undefined, keys: Object.keys(d.receipt || {}).sort() };
+  });
+  check(
+    "the 'vote-cast' event does not carry the tally",
+    evLeak.hasTally === false,
+    'event ships {receiptHash, tally} together -> listener reads the vote with no effort'
+  );
+  check(
+    "the 'vote-cast' event receipt carries no choice/nonce",
+    JSON.stringify(evLeak.keys) === JSON.stringify(['receiptHash', 'status']),
+    `event receipt keys = ${JSON.stringify(evLeak.keys)}`
+  );
+
+  // --- 10. Seal is meaningful with a REAL future close --------------------
+  // The earlier "sealed while open" check passes even against a component that
+  // seals unconditionally, because the fixture has no closes-at. Assert against
+  // a genuinely open ballot, then assert the seal actually LIFTS.
+  const sealCycle = await page.evaluate(async () => {
+    const el = document.createElement('pm-ballot');
+    el.setAttribute('ballot-id', 'seal-cycle');
+    el.setAttribute('options', JSON.stringify([{ id: 'yes', label: 'Y' }, { id: 'no', label: 'N' }]));
+    el.setAttribute('eligible-voters', '10');
+    el.setAttribute('closes-at', new Date(Date.now() + 864e5).toISOString());  // genuinely open
+    document.body.appendChild(el);
+    await el.castVote('yes', 'unit-1');
+    const whileOpen = el.state.receipts.length;
+    el.close();                                   // explicit lift condition
+    const afterClose = el.state.receipts.length;
+    return { whileOpen, afterClose };
+  });
+  check(
+    'receipts stay sealed while a genuinely-open ballot is running',
+    sealCycle.whileOpen === 0,
+    `${sealCycle.whileOpen} receipt(s) visible mid-ballot`
+  );
+  check(
+    'receipts actually publish once the ballot closes',
+    sealCycle.afterClose === 1,
+    `${sealCycle.afterClose} receipt(s) after close -> residents can never verify`
+  );
+
+  // --- 11. A malformed closes-at must not disenfranchise AND must not leak -
+  // The two consumers need OPPOSITE defaults. One shared boolean got both
+  // wrong at once: Date.now() < NaN is false -> voting closed, seal lifted.
+  const typo = await page.evaluate(async () => {
+    const el = document.createElement('pm-ballot');
+    el.setAttribute('ballot-id', 'typo');
+    el.setAttribute('options', JSON.stringify([{ id: 'yes', label: 'Y' }]));
+    el.setAttribute('eligible-voters', '10');
+    el.setAttribute('closes-at', new Date(Date.now() + 864e5).toISOString());  // valid, open
+    document.body.appendChild(el);
+
+    // Land a real receipt in the store FIRST. Without this the seal assertion
+    // is vacuous: a broken build blocks the vote, so there is no receipt to
+    // leak, and "published === 0" passes for the wrong reason.
+    await el.castVote('yes', 'unit-seed');
+    const seeded = el.state.receiptsSealed;
+
+    // NOW the author fumbles the date (day-first is an ordinary mistake).
+    el.setAttribute('closes-at', '15/08/2026 23:59');
+    const receipt = await el.castVote('yes', 'unit-2');
+    return { seeded, couldVote: !!receipt, published: el.state.receipts.length };
+  });
+  check(
+    'seed receipt is actually in the sealed store (guards the next assertion)',
+    typo.seeded >= 1,
+    `receiptsSealed = ${typo.seeded} -> the leak assertion below would be vacuous`
+  );
+  check(
+    'a typo in closes-at does not disenfranchise voters (fails OPEN)',
+    typo.couldVote === true,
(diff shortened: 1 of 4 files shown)


POWERFULMOVES:
## Adversarial review pass — 14 confirmed defects, 5 fixed here

Ran a 76-agent fan-out over this branch: 4 review lenses (vote integrity · secrecy leaks · crypto detail · spec conformance), each candidate finding then facing **two independent refuters** (one "is it real", one "does it matter for a real vote") with unanimous survival required. **32 candidates → 14 confirmed.** 18 were killed, correctly — including several plausible-sounding ones.

It caught two regressions **this PR introduced** and one comment that was simply false.

### Fixed in `6c4099ee`

**P1 — the `vote-cast` event shipped the tally next to the receiptHash.** The `state` getter seals exactly this correlation; the event re-opened it on the outward wire. The nonce commit added a comment claiming the event carried the public receipt only — it didn't. First voter needs no diffing at all: `{receipt:{receiptHash:'0x6ea9…'}, tally:{'no-confidence':1,'_total':1}}`. *Honest caveat: one verifier refuted live exploitability — no NATS publish path is wired today, so it's a loaded gun, not a firing one. Fixed anyway; §4.3 says this wire routes outward.*

**P1 — silent downgrade to a 32-bit checksum.** `crypto.subtle` is absent on plain `http://` — a co-op LAN at `http://192.168.x.x` is the realistic deployment. A 128-bit nonce was minted, then committed with FNV-1a-32. **Colliding receipt found in 0.8s / 415k evals**: one published hash opens to both `no-confidence` and `retain-board`. The commitment binds to nothing while the UI claims it proves the vote was counted — and the disclosure rendered *inside the receipt panel*, i.e. after the vote was irrevocable. Now refuses to cast with a pre-cast notice; `allow-insecure-demo-hash` keeps demos working without letting a deployment downgrade silently.

**P1 — `_isOpen()` served two consumers needing opposite defaults**, so one was always wrong:
- no `closes-at` → `true` → **sealed forever → receipts never publish → residents can never verify**. This is the default in the spec's own §6 sample *and this PR's test fixture*.
- unparseable `closes-at` → `Date.now() < NaN` is `false` → **voting closed for everyone AND the seal lifted**, at once.

Split into `_isVotingOpen()` (fails **OPEN** — never disenfranchise over a typo) and `_isSealed()` (fails **CLOSED** — never leak over a typo), plus an explicit `close()`.

**P2 — `tenant-renderer.js` + `demo.html` threw on every vote.** This PR removed `receipt.choice` and left both callers dereferencing it → TypeError inside the listener → the documented `on-vote-cast` toast wire failed silently. Both also rendered the **plaintext choice into a page-level toast** — a shoulder-surfable leak this PR didn't finish removing, failing closed only by accident. Both now summarise by hash and never name the option.

### Tests: 8 → 17, each proven to fail first

| | pre-fix | post-fix |
|---|---|---|
| event carries tally | **FAIL** | PASS |
| insecure context | `cast:true algo:fnv1a-32 warned:false` | `cast:false warned:true` |
| typo → disenfranchised | `couldVote:false` | `couldVote:true` |
| typo → unseals receipts | **1 real receipt published** | `0` |

**Two of my own tests were caught passing vacuously and rewritten.** "Sealed while open" asserted against a component that sealed *unconditionally* (fixture had no `closes-at`). And the unseal check passed for the wrong reason — the disenfranchisement bug blocked the vote, so nothing existed to leak; it now seeds a real receipt first. A test that can only pass is indistinguishable from one that works.

19/19 compose conformance unchanged. Click-through re-verified in Chromium, zero page errors.

### Deferred, with reasons

| Item | Why |
|---|---|
| **Ed25519 signing** | CHIT is symmetric HMAC — the operator can forge any receipt. Cross-service change; `PMOVES-ClawZ/ui/src/ui/device-identity.ts` (`@noble/ed25519` 3.1.0) is the reuse target. Own PR. |
| **Receipt-freeness** | The voter still holds a compellable `(choice, ts, nonce)`. Architectural (time-limited verification + revoting), not a tweak. **This PR does not make the ballot coercion-resistant.** |
| **Universal verifiability** | The component renders authority state and cannot validate a tally it didn't compute. Any client-side cross-check is theatre — a self-consistent false tally passes it. |
| **Client-side dedup** | `castVote` has no `_myChoice` guard, but also **no write path to shared state** — one-vote-per-voter is an authority-side invariant. Worth adding as tidiness; buys zero security. |
| **UTF-16 vs UTF-8 netstring length** | Real 3-way ambiguity (`'Núñez-3A'` → 8/10/8). Cannot fire today (no second implementation) — but must be fixed **before** any verifier is written, and NFC + encoding pinned in §5.4. |

### Merge mechanics — worth flagging

`feat/a2ui-v02-design` (#2133) is **not** an ancestor of `feat/a2ui-v02-impl-review-style` (#2134); #2134 carries the **rev-1** spec (0 mentions of `nonce`) while #2133 has rev-2 (9). Verified: a true 3-way merge preserves rev-2 in either order, but **squashing #2132/#2133/#2134 severs the base** → `add/add` on 14 files, and a "theirs" resolution silently reverts `_safeHref()` and the `aspectRatio` clamp. Same class as the recorded `submodule-gitlink-gate` squash drift. Suggested gate before merging this: `grep -c nonce pmoves/contracts/a2ui-v0.2-ballot.md` ≥ 9.

Also: this repo's `squash_merge_commit_message` is `COMMIT_MESSAGES`, so commit text lands on `main` — not the PR body.

POWERFULMOVES:
Superseded by #2170.

Root cause confirmed: this PR was stacked on the pre-squash #2134 branch, so after #2134 landed GitHub treated the inherited stack as new history and produced the 100+ file conflict. #2170 starts from current `main`, carries only this branch's two unique ballot-security commits, and adds the remaining public-receipt normalization plus non-vacuous correlation coverage.

Validation on the replacement: ballot secrecy 19/19, tenant compose 19/19, A2UI signature gate 9/9, and the completed repository CI gates are green. Keeping the replacement draft until the remaining CodeQL analyses/review finish.

PR #2212, opened 2026-07-24
fix(supabase): bootstrap seeds as supabase_admin + migrate kong on the real DB

POWERFULMOVES:
### 1. `supabase-bootstrap` never seeded
The psql calls used `${POSTGRES_USER:-pmoves}` — **shell-default syntax that *Make* evaluates**, looking up a var literally named `POSTGRES_USER:-pmoves` → empty → `psql -U -d pmoves` (the `-d` misparsed as the username → peer-auth failure). Even fixed to shell, `pmoves` fails socket peer-auth, and plain `postgres` isn't a superuser (no CREATE on `public`).

**Fix:** run as the privileged **`supabase_admin`** role over TCP (127.0.0.1), sourcing `PGPASSWORD` from the container env; per-file error tolerance (one bad migration no longer aborts the whole run); keep the `pmoves_bootstrap_history` tracking.

**Verified:** seeds the pmoves DB **36→123 tables, 0→172 rows** (models 55, service_catalog 27, personas 9, …). 28/28 seeds apply; 44/52 migrations (8 pre-existing SQL bugs tracked separately).

### 2. kong migration helper migrated the WRONG database
The helper hardcoded `KONG_PG_DATABASE=postgres` + a stale `KONG_PG_USER` order + `kong:3.7.1`. But compose runs **`kong/kong:3.9.1`** against the **`pmoves`** DB (`KONG_PG_DATABASE=${POSTGRES_DB:-…}`). So the helper bootstrapped kong's schema in `postgres` while kong read from `pmoves` → kong 3.9.1 **crash-looped** on the unapplied `024_380_to_390` migration.

**Fix:** match compose's DB/user resolution, use `kong/kong:3.9.1`, and run **bootstrap-then-`up`** so version-bump migrations actually apply. (Kong 3.9.1 confirmed current against Supabase's self-hosted compose.)

### DIFF
--- a/pmoves/Makefile
+++ b/pmoves/Makefile
@@ -454,10 +454,10 @@ supa-start: ensure-env-shared ## Start Supabase stack for selected runtime (cli 
 		$(LOAD_ENV_SHARED); docker run --rm --network pmoves_data \
 			-e KONG_DATABASE=postgres \
 			-e KONG_PG_HOST=supabase-db \
-			-e KONG_PG_DATABASE=postgres \
-			-e KONG_PG_USER=$${SUPABASE_DB_USER:-postgres} \
-			-e KONG_PG_PASSWORD=$${SUPABASE_DB_PASSWORD:-postgres} \
-			kong:3.7.1 kong migrations bootstrap --yes >/dev/null 2>&1 || true; \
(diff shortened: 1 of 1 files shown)


POWERFULMOVES:
Superseded by the clean closeout PR (Makefile fix + the 5 repaired migrations, off current main without the redundant already-merged tokenism commit).

PR #2225, opened 2026-07-25
feat(provision): add SPARK Ubuntu 24.04 AppImage / A0 launcher provisioning

POWERFULMOVES:
## Summary
Adds the provisioning script and validator for running the A0 Launcher AppImage on Ubuntu 24.04 arm64 SPARK nodes, including Agent Zero instance port checks.

## Changes
- `deploy/provision/spark-ubuntu-24.04-appimage.sh`: root provisioning script that installs Docker, places an `a0-launcher` wrapper (with FUSE fallback), and runs validation.
- `pmoves/scripts/check_appimage_prereqs.sh`: non-root validator for `libfuse.so.2`, AppImage presence, `a0-launcher` on `PATH`, and expected Agent Zero listener ports. Supports `--install` for optional apt install of the FUSE 2 userspace library.

## Testing
- [ ] Run `bash deploy/provision/spark-ubuntu-24.04-appimage.sh` on a fresh Ubuntu 24.04 arm64 node (or VM).
- [ ] Run `bash pmoves/scripts/check_appimage_prereqs.sh` after provisioning and verify all checks pass.

## Notes
- The wrapper falls back to cached `--appimage-extract` when FUSE 2 is unavailable, matching the expected SPARK runtime environment.
- Both scripts are executable (`chmod +x`).

### DIFF
--- a/deploy/provision/spark-ubuntu-24.04-appimage.sh
+++ b/deploy/provision/spark-ubuntu-24.04-appimage.sh
@@ -10,8 +10,11 @@
 
 set -euo pipefail
 
-USER_HOME="${SUDO_USER:-${HOME}}"
 USER_NAME="${SUDO_USER:-${USER}}"
+# SUDO_USER holds the login name, not a path — resolve the account's real home
+# via the passwd database so wrapper/AppImage/PMOVES_ROOT don't land in $CWD.
+USER_HOME="$(getent passwd "$USER_NAME" 2>/dev/null | cut -d: -f6)"
+USER_HOME="${USER_HOME:-${HOME}}"
 APPIMAGE_SOURCE="${A0_LAUNCHER_APPIMAGE:-${USER_HOME}/Downloads/a0-launcher-0.9-linux-arm64.AppImage}"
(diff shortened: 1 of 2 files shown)


POWERFULMOVES:
## Pair review — Codex feedback addressed (commit `4d39b05f4`)

Reviewed on the SPARK node. All three Codex findings addressed in-scope:

- **P1 — `SUDO_USER` home resolution** ✅ `USER_HOME` now resolves via `getent passwd "$USER_NAME" | cut -d: -f6` (falls back to `$HOME`). `SUDO_USER` holds a login name, not a path, so `APPIMAGE_SOURCE`/`PMOVES_ROOT`/`~/.local/bin` no longer resolve under $CWD.
- **P2 — Docker group membership** ✅ `usermod -aG docker` moved out of the install-needed branch; the target user is now added regardless of whether Docker was just installed or already present.
- **P2 — non-executable AppImage** ✅ Now counts as `missing` (fatal, exit 1) instead of a soft warning, since the wrapper execs the AppImage for `--appimage-extract`.

`bash -n` passes on both scripts. Diff stays scoped to the two files this PR introduces.

_Pair reviewer: PMOVES-AGENT-ZERO-SPARK — `ACK::SPARK-CLAUDE::PR2225-PAIR-REVIEW`_

PR #2263, opened 2026-07-27
fix(compose): down-all left voice, Hi-RAG, Ollama and ComfyUI running

POWERFULMOVES:
## Problem

`down-all` is documented **"Stop ALL services in graceful reverse order"** and prints **"✅ All services stopped gracefully"** — but it chained only **10 of 25** `down-*` targets.

Five services had no teardown path anywhere in the Makefile:

| service | teardown before |
|---|---|
| `flute-gateway` | none |
| `voice-relay` | none |
| `hi-rag-gateway` / `-gpu` / `-v2` / `-v2-gpu` | none |
| `pmoves-ollama` | none |
| `comfyui` | none |

A fleet-wide shutdown therefore reported success while leaving **the voice fabric live** (the one that went live 2026-07-11) and **Ollama + ComfyUI pinning VRAM**. On a fleet whose stated principle is dynamic / no-wasted-electricity, a teardown that *looks* clean and isn't is the worst failure mode — nothing surfaces the leak.

## Fix

Adds `down-voice`, `down-search`, `down-inference`; wires all three into `down-all`.

### Ordering is load-bearing

It follows the **init chain in reverse**. Ollama comes up *first* at init — it's the easiest local dependency, and the local agents (Hermes / OpenClaw / Claude / Crush) boot against it before any PMOVES service exists. TensorZero fronts it later as a gateway.

So on the way down, TensorZero must stop **early** (position 3) and the backends **late** (position 9). Otherwise the gateway is briefly live with its backend already gone, and in-flight requests fail instead of draining. `down-voice` / `down-search` likewise precede `down-inference` — they are callers.

```
1 integrations   4 voice     7 model-management  10 bus    13 supabase
2 tensorzero     5 search    8 workers           11 data
3 agents         6 —         9 inference         12 obs
```

### The vibevoice trap

`vibevoice` is defined in `docker-compose.voice.yml`, which is **not** in `STACK_FILES`. A bare `$(DC) stop vibevoice` cannot resolve it — docker compose errors, `|| true` swallows it, and the container keeps running while the target reports success. It now gets the voice overlay appended explicitly.

**The pre-existing `stop-vibevoice` target carried this exact bug** and is fixed the same way — same root cause, same concern, so it's in this commit rather than left knowingly broken.

## Verification

- All 11 service names resolve against their compose file sets — verified by static YAML parse, since `docker compose config` requires hydrated tier env this agent cannot access.
- `make -n down-voice down-search down-inference` expands all three correctly.
- `make -n stop-vibevoice` now emits `-f docker-compose.voice.yml stop vibevoice`.
- `down-all` chain order confirmed as listed above.

**Not verified:** a live up/down cycle. That needs an operator run on a node with the stack actually up — worth doing before showtime.

## Notes

Two things I checked and found *not* to be defects, so they're deliberately untouched:
- The 45 `up-*` targets without a matching `down-*` are mostly variants correctly covered by a shared teardown (all 8 `AGENT_SERVICES` resolve in base compose, so `down-agents` does cover `up-agents-hardened` et al).
- `minio` / `presign` are already covered via `DATA_SERVICES` → `down-data`.

## Summary by CodeRabbit

* **Enhancements**
  * Added dedicated commands to shut down voice, search, and inference services.
  * Expanded the full shutdown process to stop service tiers in a clear, orderly sequence.
  * Corrected voice service shutdown to use the appropriate configuration, ensuring the service is stopped and removed reliably.

### DIFF
--- a/pmoves/Makefile
+++ b/pmoves/Makefile
@@ -1174,20 +1174,20 @@ health-dormant: ## Health-check dormant/newly-activated services
 # =============================================================================
 
 .PHONY: down-all
-down-all: ## Stop ALL services in graceful reverse order
+down-all: ## Stop ALL services in graceful reverse order (voice, search, inference, agents, data)
 	@echo "🛑 Graceful shutdown starting..."
-	@-$(MAKE) --no-print-directory down-integrations 2>/dev/null
-	@-$(MAKE) --no-print-directory down-tensorzero
-	@-$(MAKE) --no-print-directory down-agents
-	@-$(MAKE) --no-print-directory down-voice
-	@-$(MAKE) --no-print-directory down-search
-	@-$(MAKE) --no-print-directory down-model-management
-	@-$(MAKE) --no-print-directory down-workers
-	@-$(MAKE) --no-print-directory down-inference
-	@-$(MAKE) --no-print-directory down-bus
-	@-$(MAKE) --no-print-directory down-data
-	@-$(MAKE) --no-print-directory down-obs
-	@-$(MAKE) --no-print-directory down-supabase
+	$(MAKE) --no-print-directory down-integrations 2>/dev/null || true
+	$(MAKE) --no-print-directory down-tensorzero || true
+	$(MAKE) --no-print-directory down-agents || true
+	$(MAKE) --no-print-directory down-voice || true
+	$(MAKE) --no-print-directory down-search || true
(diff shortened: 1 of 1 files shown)


POWERFULMOVES:
All 4 review threads addressed (commit above): OmniVoice added to down-voice, @- prefix replaced with proper error propagation, help docs updated for all new teardown targets.

PR #2332, opened 2026-08-01
docs(archon+tokenomics): mint contract review, value chain review, and the missing subject contracts

POWERFULMOVES:
Reviews the Archon mint contract and the tokenomics value chain, and registers the subject contracts that were missing. Docs and catalog only — one docstring change, no behaviour change.

## Why

Minting is a launch gate for the Fordham Hill pilot (5 agents) and the PMOVES agent factory, and it ties directly into how agents and operators create value and get paid.

## Findings

**The mint contract is well-designed but has zero implementation.** `AgentMintSpec` is a real declarative manifest and `archon-qa-agent` is a real 7-check blocking gate — but `grep -rln "archon\.mint" --include=*.py --include=*.ts pmoves/` returns nothing. ~40 doc references, 0 lines of code.

**The subject catalog was self-blocking.** It carried only three `archon.*` subjects. The entire mint family was absent, including `archon.qa.result.v1`. Since `archon-qa-agent` check 3 rejects manifests referencing unregistered subjects, the gate would have rejected mint manifests **for using the mint contract's own subjects**. All five now registered.

**Archon 0.6.0 can serve mint — as a workflow, not a REST resource.** Minting an agent *is* a code-scaffolding operation and 0.6.0 *is* a code-scaffolding agent. Step 2 maps onto `POST /api/workflows/mint-agent/run`, and 0.6.0's HITL approve/reject is a *better* home for the QA gate than the original design — a hard runtime block instead of a convention.

**The ritual produces artifacts that fail the repo's own gate.** `validate_agent_registry.py` requires every agent in both `agent_registry.yaml` and exactly one team (verified live: 97/97). Mint steps 4–5 write only the persona doc and agent definition — nothing writes the registry or team entry.

**Metering is blocked by a privacy policy that is correct.** TensorZero observability is disabled (`tensorzero.toml:8-30`) because enabling it stores full prompt+response text with no TTL. Verified live: ClickHouse has no `tensorzero` database and zero tables. The resolution is that TensorZero's observability is all-or-nothing — it records *content* when tokenomics needs only *counts*. Hence the new content-free `chit.economics.usage.v1` contract, which trips none of the six documented re-enable conditions. **This PR explicitly recommends NOT re-enabling observability.**

**A complete designed chain already exists, and the review initially missed it (§6a).** `pmoves/contracts/schemas/` holds `token.work.attested.v1` → `work_attestations` → `attribution.recorded` → `settlement.requested/recorded`, with Ed25519 attestation, dual Firefly + on-chain settlement legs, and deterministic idempotency keys — plus a 211-line recorder service. A payable human identity **is** expressible: `work_attestations.contributor_id` is bound to Supabase `auth.uid()` by RLS. None of it runs: `token-stub` is in no compose file and the migration has never been applied.

## Judgment calls

- **Did not patch the cost query's `GROUP BY`.** All four of its queries read `FROM requests`, a table nothing in this repo creates. Adding `function_name` would have polished broken code and left someone believing cost tracking worked. Flagged the tool as non-functional instead.
- **Corrections are recorded, not quietly fixed** (§7), including two wrong researcher claims and this review's own initial baseline error, plus the search-methodology lesson: enumerate the contract directory before grepping subject names.

## Testing

Docs + catalog + one docstring; no runtime path touched.

- `python -m ast` parse of `llm_observability_specialist.py` — PASS
- `make -C pmoves compose-split-check` — PASS
- `pytest pmoves/tests/test_chit_contract_schemas.py` — 5 passed, 5 failed. **Pre-existing:** `git diff --stat <merge-base> HEAD -- pmoves/tests/ pmoves/contracts/` is empty, so this branch cannot have caused them.
- `make -C pmoves docs-reconcile-check` — fails with 4 findings, all pre-existing: the three flagged docs are untouched by this branch, and the commit-drift threshold is 5 while the dashboard was already 41 behind before these commits.

### DIFF
--- a/.claude/context/nats-subjects.md
+++ b/.claude/context/nats-subjects.md
@@ -1845,6 +1845,41 @@ nats server report connections
   {"creator_id": "<supabase auth.users.id>", "handle": "darkxside", "role": "operator", "email_hash": "<sha256>", "provider": "google", "default_room": "4090-field.room.control", "github_username": "<optional>", "ts": "2026-08-01T00:00:00Z"}
   ```
 
+## Token Work Attestation Subjects
+
+> The input side of the economy: cryptographic proof that a contributor performed a unit of work.
+> Schema: `pmoves/contracts/schemas/token/work.attested.v1.schema.json`. Chain analysis:
+> `pmoves/docs/handoffs/PMOVES_VALUE_CHAIN_REVIEW.md` §6a.
+>
+> **Status: contract + recorder service + ledger migration all exist; NONE of it is deployed.**
+> `pmoves/services/token-stub/app.py` (dry-run recorder) is in no compose file, and
+> `pmoves/supabase/migrations/20260425000300_work_attestations.sql` has never been applied — the relation
+> `pmoves_core.work_attestations` does not exist on the live Supabase.
+
+**`token.work.attested.v1`** — Signed work attestation (contributor/agent → attestation recorder)
+
+- **Direction:** Published on completion of a unit of work → Consumed by `token-stub` (and, once wired, the
+  attribution stage)
+- **Purpose:** Establish *who did what*, verifiably, as the input to attribution and eventually settlement.
+- **Crypto note:** this contract specifies **Ed25519** (asymmetric, 128-hex signature) — deliberately
+  different from the live CHIT trail's **symmetric HMAC** with a single operator-held passphrase.
+  Asymmetric is the right model for attribution a contributor should be able to prove independently. The
+  divergence is unresolved and needs an explicit decision, not a silent merge.
+- **Identity note:** `contributor` is a UUID that lands in `work_attestations.contributor_id`, which the
+  migration's RLS policy binds to Supabase `auth.uid()`. **This is the only payable human anchor in the
+  repo** — the Archon mint contract's `creator_id`/`owning_persona` is a separate, unimplemented model.
+- **Payload:**
+  ```json
+  {
+    "work_id": "<uuid>",
+    "contributor": "<uuid — Supabase auth.users.id>",
+    "attestation_sig": "<128 hex chars, Ed25519>",
+    "merkle_root": "0x<64 hex>",
+    "attested_at": "2026-08-01T00:00:00Z",
+    "metadata": {}
+  }
+  ```
+
 ## CHIT Economics Subjects
 
 > Cost/usage metering for the tokenomics layer. Design rationale and the full value-chain analysis:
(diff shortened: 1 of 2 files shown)


POWERFULMOVES:
All five addressed in `4a2caf0e7`. The two P1s refuted claims this doc made confidently *after* it had already corrected itself once — recorded in §6b and §7 rather than silently edited, since they're the sort of claims that get built on.

**P1 — recorder does not match the ledger.** Confirmed: `record_to_supabase()` posts `contributor` and `recorded_at`; the migration defines `contributor_id` and no such column. PostgREST rejects every insert. `token-stub` downgraded from `[BUILT]` to `[BUILT but non-functional]`, and §8 step 5 now says **do not apply the migration or add token-stub to Compose yet** — deploying as-is would produce a ledger that silently records nothing while appearing live. That was previously listed as an easy unblocking step; it isn't.

**P1 — verify attestations before adopting this payout anchor.** Confirmed and this was the worst of the three. `grep -nE "ed25519|nacl|cryptography|verify|public_key"` over `token-stub/app.py` returns **nothing**; the only validator is `non_empty()`, and the schema's `^[0-9a-fA-F]{128}$` is a format check with no signer key and no signed-message definition. Any caller can attribute arbitrary work to any contributor by posting 128 arbitrary hex chars. The "Ed25519, independently verifiable" framing is removed. `contributor_id` is demoted from "the stronger model" to a **candidate gated on four preconditions**: FK to `auth.users`, an INSERT/ownership RLS policy, a defined signed-message format with key binding, and real verification in the recorder.

**P2 — read policy is not identity binding.** Correct. `USING (auth.uid() = contributor_id OR auth.uid() IS NOT NULL)` — the second disjunct is true for any authenticated user, so the predicate is always true and every authenticated user can read every attestation. SELECT-only, no ownership constraint, no FK. It reads like ownership and is not; that's now stated explicitly in §6b as the root-cause pattern.

**P2 — schema for the content-free invariant.** Right that a Markdown invariant is unenforceable. The catalog entry now carries an explicit ⚠️ that it is **prose-only, NOT machine-enforced**, and §8 step 2 specifies what's needed: `chit/economics.usage.v1.schema.json` with `"additionalProperties": false` plus a `topics.json` entry, landed **before the first publisher**.

I could not write it in this session: `pmoves/contracts/schemas/` is a damage-control `readOnlyPath` requiring an operator-set `KNOWN_ROAD=schema:pr:2332` in the launching shell. Self-granting that is not something I'll do, so it's tracked as owed rather than quietly skipped. Same applies to the `archon.mint.*` schemas.

**P2 — reconcile the review with this commit's registrations.** Fixed: table now reads "yes — registered by this PR", §4.1 retitled RESOLVED and kept as rationale rather than an open task, and the "register all five before any mint code lands" instruction updated.

PR #2369, opened 2026-08-04
fix(fork-sync): PMOVES.YT was never synced — 206 behind on the hardened branch

POWERFULMOVES:
PMOVES.YT is the yt-dlp fork and the ingestion front door for **any** of yt-dlp's ~1800 extractors — not YouTube-only — feeding Jellyfin as the media library for the creator pipeline. It is absent from **both** curated lists in `fork-sync.yml`, so it has never been synced.

```
gh api repos/yt-dlp/yt-dlp/compare/master...POWERFULMOVES:PMOVES.YT:PMOVES.AI-Edition-Hardened
-> {"ahead":18,"behind":206,"status":"diverged"}
```

## The failure mode

**Discovery and sync disagree.** The *Discover forks* step enumerates every POWERFULMOVES fork dynamically and labels PMOVES.YT `STALE`/`CRITICAL` every Monday. The *Sync forks* step then iterates a hardcoded 25-entry `FORKS` list, and the App token is scoped to a matching 25-entry `repositories` list. **A fork missing from those two lists is reported forever and fixed never.**

## Why it isn't cosmetic

The running container ships yt-dlp `2026.02.04` against upstream `2026.07.04`. YouTube extraction now requires a JS runtime the image doesn't have, so the front door is down. `PMOVES.AI-Edition-Hardened` is the fork's *default* branch — so "hardened" had drifted into name-only.

## Verification

Confirmed it won't be skipped by the workflow's own guards:
- 206 behind < `behind_max` (1000) ✅
- 18 ahead < `ahead_max` (20) ✅ — **but only 2 commits of margin.** Two more PMOVES-side commits and it flips to `MANUAL`.

The `PMOVES.AI-Edition-Hardened` override is technically redundant (branch precedence 2 already resolves it from `.gitmodules`), but set explicitly to match supabase and Jellyfin — this service just proved it can fall through silently.

`fork-sync.yml` re-parses as valid YAML; diff is 2 insertions, 0 deletions.

## Scope

This closes the single case. **The general gap — 57 submodules vs 25 synced forks — is not fixed here** and needs a coverage gate (ratchet-baselined, so it doesn't red-wall the existing unlisted set). Under review separately, including the real tension that the App token list is least-privilege by intent, so widening it isn't free.

Related follow-ups, not in this PR:
- The deno/JS-runtime fix from #2358 landed on `pmoves/services/pmoves-yt/Dockerfile`, which compose does **not** build — compose builds the fork's `pmoves_yt_service/Dockerfile`. Fix belongs in the fork.
- Audit of `pmoves/services/pmoves-yt/` (operator: keep, it's more than that one service).

## Summary by CodeRabbit

* **Chores**
  * Updated repository synchronization to include PMOVES.YT.
  * Configured upstream synchronization from yt-dlp/yt-dlp to the PMOVES.YT fork on the PMOVES.AI-Edition-Hardened branch.

### DIFF
--- a/pmoves/configs/tac_schema.yaml
+++ b/pmoves/configs/tac_schema.yaml
@@ -14,11 +14,20 @@
 #   id: string            - Unique node identifier (dot-separated path)
 #   task: string          - Human-readable task description
 #   action: object        - Verification action
-#     type: string        - One of: file_exists, grep, command, manual
-#     target: string      - File path or command (relative to repo root)
+#     type: string        - One of: file_exists, grep, command, http, manual
+#     target: string      - File path or command (relative to repo root).
+#                           For http, also accepted as the endpoint (see url).
 #     pattern: string     - Regex pattern (for grep type)
 #     invert: boolean     - If true, finding the pattern means FAIL (negative grep). Default: false
-#     expect: string      - Expected result description
+#     multiline: boolean  - grep only. If true, `.` also matches newlines (re.DOTALL),
+#                           so a pattern may span lines. Opt-in: default false keeps
+#                           existing single-line patterns behaving as written.
+#     url: string         - http only. The endpoint to probe. Trees use both `url`
+#                           (16 nodes) and `target` (22); the runner accepts either.
+#     expect_status: int  - http only. Status code that means PASS. Default: 200.
+#                           This is the machine-readable assertion — `expect` is not.
+#     expect: string      - Expected result description. FREE-FORM PROSE, for humans
+#                           ("200 OK", "Prometheus metrics for ..."). Never parsed.
 #   context: string       - Where to look / additional guidance
 #   agent_hint: string    - Suggested agent for this task (e.g., "codex", "archon")
 #   status: string        - One of: pending, pass, fail, skip (default: pending)
@@ -30,6 +39,14 @@ schema:
     - file_exists    # Check if a file or directory exists
     - grep           # Search for a pattern in a file
     - command        # Run a shell command and check exit code
+    - http           # Liveness probe: unauthenticated GET, asserts on status code
+                     # only. http/https, no redirects, link-local blocked, body
+                     # never read. A node declaring method/headers/body/json/auth
+                     # or a content predicate (expect_body, expect_json,
+                     # expect_contains, expect_match, expect_jsonpath, assert) is
+                     # UNSUPPORTED and evaluates FAIL — the probe will not silently
+                     # send a different request and report PASS on the status alone.
+                     # Use `command` for anything needing a real request body.
     - manual         # Requires human review (always pending)
   statuses:
     - pending        # Not yet evaluated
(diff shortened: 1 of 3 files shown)


POWERFULMOVES:
## Correction to this PR's description — discovery is NOT universal

My description says the *Discover forks* step "enumerates every POWERFULMOVES fork dynamically and labels PMOVES.YT `STALE`/`CRITICAL` every Monday." **That is false**, and review caught it. Verified:

```
$ gh repo list POWERFULMOVES --fork --limit 500 --json name --jq length
374
$ gh repo list POWERFULMOVES --fork --json name --jq length      # the workflow's call, line 79
30
```

`fork-sync.yml:79` passes no `--limit`, and `gh repo list` defaults to **30**. The org has **374 forks**. The audit covers **8%**, ordered by push recency, with no guarantee submodule-forks are in the window.

So the real story is worse than "discovery reports it, sync ignores it" — **PMOVES.YT was most likely never discovered at all.** There was no safety net, not even an ignored one. That missing `--limit` is the mechanism by which every unlisted fork stayed invisible, and it is a P1 independent of this PR.

## Corrected numbers

I wrote "57 submodules vs 25 synced forks" and "~32 unlisted". Accurate figures:

| | count |
|---|---|
| `.gitmodules` entries | 57 |
| unique repos | 56 (`PMOVES-Archon` is registered at two paths) |
| actual forks | **47** |
| first-party, legitimately need no sync | **9** |
| submodule-forks in neither list | **25** → **24** with this PR |

So this PR moves coverage 22/47 → 23/47, not "32 → 31". The gap between my 32 and the real 25 is the 9 first-party repos.

Also worth stating plainly: of the 24 remaining, only **13 are auto-syncable** by the current mechanism. Four exceed `behind_max`/`ahead_max` (`PMOVES-neo4j` at 4555/72, `pmoves-pipecat` 4464, `PMOVES-Tailscale` 1267, `PMOVES-Danger-infra` 1191) and **two are structurally unsyncable** — `PMOVES-Deep-Serch` and `PMOVES-MAI-UI` return HTTP 404 "no common ancestor", so no merge-based sync can ever work on them. A coverage gate should record those as deliberate exclusions with a reason, not chase them.

This PR is unchanged and still correct — PMOVES.YT belongs in both lists, and 206 behind / 18 ahead clears both thresholds. Only the framing in the description was wrong.

POWERFULMOVES:
**Silent-failure finding for the PMOVES.YT lane** (issues disabled on the fork, parking here on the PMOVES.YT sync PR): `yt.py::supa_upsert` (~line 1341) wraps its PostgREST POST in `except Exception: return None` with zero logging. On 5090 today, a stale frozen `SERVICE_ROLE_KEY` meant every transcript upsert 401'd at Kong while ingest still returned `{"ok":true,"transcript":{"ok":true}}` — transcription succeeded, storage silently lost (reproduced in-container: 401, key len 188). Exact same class as the B850 finding in #2339. Fix wants: (1) log the exception with table name, (2) `stored: false` surfaced in the ingest response when upsert returns None, (3) audit `supa_update` and siblings. — 5090-CLAUDE

PR #2370, opened 2026-08-04
fix(fork-sync): the drift audit saw 8% of forks and the summary always read zero

POWERFULMOVES:
Three defects that together made this workflow's output untrustworthy. All measured, not inferred.

## 1. Discovery covered ~8% of forks — this is the root cause

```
gh repo list POWERFULMOVES --fork --limit 500 --json name --jq length  → 374
gh repo list POWERFULMOVES --fork --json name --jq length              →  30    ← line 79
```

No `--limit`, and `gh` defaults to **30**. Results are ordered by push recency with no guarantee submodule-forks land in the window.

**This is the mechanism by which every unlisted fork stayed invisible.** PMOVES.YT sat 206 commits behind on its hardened branch while the "Fork Drift Audit" reported nothing — it was almost certainly never *discovered*, not merely never synced. I originally described this in #2369 as "discovery reports it, sync ignores it"; that was wrong, and this is the correction.

## 2. Both loops ran in subshells, so every counter was discarded

`cmd | while ...` puts the loop body in a subshell. `SYNCED` / `SKIPPED` / `CONFLICTS` / `ERRORS` / `RESULTS` mutations die at the pipe boundary.

**`=== Results: $SYNCED synced ===` printed `0` with an empty list on every run**, regardless of how many PRs it opened. The number an operator uses to judge sync health *could not be non-zero*.

`MAX_FORKS` was affected identically — its check read a `SYNCED` that never incremented, so the cap never engaged and the "Reached max" break was unreachable. Both loops now use process substitution.

## 3. The compare endpoint was called twice per fork, and its failure path was broken

`bh=$(...) || echo '-1'` writes `-1` to **stdout**, not to `bh`. On failure `bh` is empty, `[ "" -gt 999 ]` emits a bash error, and stray `-1` lines pollute the audit. Now one call parsing both fields with a real fallback — which also removes ~374 redundant API calls per run at full coverage.

## Verification

```
cmp="206|18" -> bh=206 ah=18 => STALE
cmp="0|0"    -> bh=0   ah=0  => synced
cmp=""       -> bh=-1  ah=-1 => ERROR
process substitution: N=3   (pipe version: 0)
```

Both `run:` blocks pass `bash -n`. Workflow re-parses as valid YAML.

## Behaviour change worth flagging

The audit now prints **only non-clean forks** plus a tallied summary. At 374 forks, a `synced` line per clean repo buries the signal. If you'd rather keep per-fork lines, say so and I'll flip it.

Also note the *uncomparable* bucket includes forks with **no common ancestor** (compare returns 404) — those can never be merge-synced and need re-forking, not retrying. `PMOVES-Deep-Serch` and `PMOVES-MAI-UI` are in that state today.

## Not in this PR

- Coverage gate + registry (24 submodule-forks still undeclared) — design reviewed, needs the `sync: false` + reason approach so token scope doesn't widen
- `MAX_FORKS` defaults to 3 on schedule, so 25 listed forks take ~9 weekly runs for a full pass
- `permissions: contents: write` is unused (every step uses the App token); stale `google/A2UI` parent slug

### DIFF
--- a/.github/workflows/fork-sync.yml
+++ b/.github/workflows/fork-sync.yml
@@ -76,22 +76,43 @@ jobs:
           GH_TOKEN: ${{ steps.app-token.outputs.token }}
         run: |
           echo "=== Fork Drift Audit ==="
-          gh repo list POWERFULMOVES --fork --json name,parent --jq '.[] | "\(.name)|\(.parent.owner.login)/\(.parent.name)"' | while IFS='|' read fork upstream; do
+          # `gh repo list` defaults to 30 results. Without --limit this audited 30
+          # of the org's ~374 forks (~8%), ordered by push recency, with no
+          # guarantee submodule-forks were in the window — which is how forks like
+          # PMOVES.YT sat 206 commits behind while this step reported "clean".
+          #
+          # Fed via process substitution, NOT a pipe: `... | while` runs the loop
+          # body in a subshell, so counter mutations are discarded at the pipe
+          # boundary and the summary is always zero.
+          SYNCED_N=0; STALE_N=0; CRIT_N=0; ERR_N=0
+          while IFS='|' read -r fork upstream; do
+            [ -z "$fork" ] && continue
             fb=$(gh api "repos/POWERFULMOVES/$fork" --jq '.default_branch' 2>/dev/null) || continue
             ub=$(gh api "repos/$upstream" --jq '.default_branch' 2>/dev/null) || continue
-            bh=$(gh api "repos/$upstream/compare/$ub...POWERFULMOVES:$fork:$fb" --jq '.behind_by' 2>/dev/null) || echo '-1'
-            ah=$(gh api "repos/$upstream/compare/$ub...POWERFULMOVES:$fork:$fb" --jq '.ahead_by' 2>/dev/null) || echo '-1'
-            if [ "$bh" = "0" ] 2>/dev/null; then
-              echo "  $fork: synced"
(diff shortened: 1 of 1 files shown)


POWERFULMOVES:
Closing as **fully superseded by #2373**, which merged at 02:12Z while this was open.

I initially thought one piece survived — the duplicate compare call. **That was my error.** My grep counted 2 occurrences of the compare URL across two *different* steps (discover + sync), not two calls inside one loop. Main's discovery step already does a single call:

```bash
cmp=$(gh api ".../compare/..." --jq '{bh: .behind_by, ah: .ahead_by}') || cmp='{"bh":-1,"ah":-1}'
```

Everything this PR fixed is on main, and in places done better:

| defect | main |
|---|---|
| `--limit` missing (8% coverage) | fixed, `--limit 1000` |
| duplicate compare call | fixed, single call + jq parse |
| `\|\| echo '-1'` writing to stdout | fixed, proper fallback assignment |
| subshell eating sync counters | fixed, process substitution |
| — | plus `sort -u`, which mine did not have |

Two sessions independently diagnosed the same three defects from the same symptom and converged on the same fixes. No disagreement on substance — just duplicated effort, which is on me for not checking `main` before building.

The one behavioural difference in mine was output shape (only non-clean forks + a tally, rather than a line per fork). At 374 forks that may still be worth having, but it's a preference not a defect, and not worth a PR on its own. Raising it as a note instead.

PR #2371, opened 2026-08-04
fix(tac): unknown action types failed open — 141 assertions asserted nothing

POWERFULMOVES:
`evaluate_node` initialised `status` to `"pending"` and only overwrote it inside `if/elif` branches for `file_exists`, `grep`, `command`, `manual`. **There was no `else`.**

An unrecognised action type fell through to `pending` with an empty detail — indistinguishable from a deliberate `manual` node — and since `pending` is not `fail`, the run still exited 0.

**So a green TAC run could not be trusted.** Some of the green was typos and unimplemented types.

## Measured across the estate

**141 nodes in 18 of 43 trees** use types this runner never implemented:

| count | type |
|---|---|
| 72 | `shell` |
| 38 | `http` |
| 7 | `file.exists` |
| 4 | `supabase_migration` |
| 14 | `script`, `env_check`, `edit`, `nats_subject`, `pytest`, `pr_state`, `web`, `gh_issue`, `file_or`, `nats_messages_seen` |

Worst affected: `node-4090-laptop` (20), `pinokio-p8` (16), `pmoves-launch-readiness` (15), `jetson-orin` (13).

Roughly **a fifth of every assertion in the estate was inert**. The sharpest example: `node-5090-powerfulmoves.tac.yaml:218` is a `type: http` probe of `localhost:8077/healthz` — it reads exactly like the check that would have caught the yt-dlp extraction outage, and it has never executed once.

## What this changes

Unknown types now `fail` with a detail naming the offending type and listing what's supported. Trees using only supported types are unaffected — `soundcloud-ingest` still reports **10 pass / 4 fail / 4 pending**, unchanged.

## What this deliberately does NOT do

**No aliasing of `shell`→`command` or `file`→`file_exists`.** They aren't simple renames:

- `shell` nodes carry their command in `action.command`, `http` nodes in `action.url` — neither is the `action.target` the runner reads
- sampled `shell` targets call binaries outside `_ALLOWED_COMMANDS` (e.g. `nvidia-smi`)

So aliasing would convert silent-pending into *failing*, not passing, and needs its own decisions about the allowlist. Separate change.

## No CI impact

Verified zero references to `tac_runner`/`tac_trees` in `.github/`, `Makefile`, or `mk/` — TAC is invoked only via the `/tac:review` slash command. This changes agent-facing output, not any gate.

**Wiring the runner into CI should come after these 141 are triaged, not before** — it already `return 1 if counts["fail"] > 0`, so it's mechanically ready, but turning it on now would red-wall on newly-surfaced gaps plus the pre-existing false negatives from `grep` lacking `re.DOTALL`.

## Follow-ups this exposes

1. Triage the 141 — implement `http` (38) and `shell` (72), or convert to supported types
2. `_check_grep` uses `re.search` without `DOTALL`, so `service.*healthcheck` can't span lines — produces false negatives on exactly the compose assertions people most want
3. `soundcloud-ingest.tac.yaml:118` asserts `pmoves-minio:`, a service name that has never existed (`git log -S` returns zero commits) — wrong since authoring, never caught because nothing ran it

### DIFF
--- a/pmoves/tools/tac_runner.py
+++ b/pmoves/tools/tac_runner.py
@@ -151,6 +151,32 @@ def evaluate_node(node: dict) -> dict:
         elif action_type == "manual":
             result["status"] = "pending"
             result["detail"] = "requires manual review"
+        else:
+            # Fail closed on unrecognised action types.
+            #
+            # Previously there was no else branch, so an unknown type fell
+            # through to the initial status "pending" with an empty detail. That
+            # is indistinguishable from a deliberate `manual` node, so a typo or
+            # an aspirational action type silently asserted NOTHING while the
+            # tree still reported green — you could not tell a passing tree from
+            # a hollow one.
+            #
+            # This is not hypothetical: 141 nodes across 18 of the 43 trees use
+            # types this runner has never implemented — `shell` (72), `http`
+            # (38), `file` (7), plus supabase_migration/script/env_check/edit/
+            # nats_subject/pytest/pr_state/web/gh_issue/file_or/
+            # nats_messages_seen. Roughly a fifth of every assertion in the
+            # estate was inert.
+            #
+            # Note these are not simple aliases: `shell` nodes carry their
+            # command in `action.command` and `http` nodes in `action.url`,
+            # neither of which is `action.target`. Implementing them is a
+            # separate change; this one only stops them passing silently.
+            result["status"] = "fail"
+            result["detail"] = (
+                f"unsupported action type '{action_type}' — asserts nothing. "
+                f"Supported: file_exists, grep, command, manual."
+            )
 
     # Recurse into children
     for child in node.get("children", []):


POWERFULMOVES:
Closing as **superseded by #2373**, which merged at 02:12Z while this was open and made the same change.

Verified on current `main`:
```
git show origin/main:pmoves/tools/tac_runner.py | grep -ci "unsupported action"  -> 2
```

The fail-closed `else` branch is already there. Two sessions independently found and fixed the same defect — no disagreement about the fix, just duplicated effort.

**What is NOT on main and remains open: #2382** (http action type + opt-in multiline grep). Verified absent:
```
git show origin/main:pmoves/tools/tac_runner.py | grep -c 'action_type == "http"'  -> 0
```

That PR was stacked on this branch; I'm rebasing it onto `main` so it stands alone. It implements the 38 `http` nodes — including `node-5090-powerfulmoves.tac.yaml`'s `localhost:8077/healthz`, the probe that would have surfaced the yt-dlp outage — and fixes the `re.DOTALL` false-negative class.

PR #2382, opened 2026-08-04
feat(tac): implement http probes and opt-in multiline grep

POWERFULMOVES:
**Stacked on #2371** — base is `fix/tac-unknown-action-fails`, not `main`. Merge that first.

#2371 made 141 assertions report *"unsupported action type"* instead of silently passing. This implements the largest block of them and fixes the false-negative class in the assertions that *did* run.

## `http` — 38 nodes, the biggest block

Liveness probes of fleet services. One of them — `node-5090-powerfulmoves.tac.yaml`, `localhost:8077/healthz` — is **precisely the check that would have surfaced the yt-dlp extraction outage**, and it had never executed.

Trees use **both** key conventions: 16 nodes carry the endpoint in `url`, 22 in `target`. Accepts either. `expect` is free-form prose (`"200 OK"`, `"Prometheus metrics for …"`) and is *not* a status code — the numeric override is the separate `expect_status` key.

Connection failure is a **FAIL, not an error**. "The service is down" is a real assertion result; that's the entire point of a liveness probe.

## Security — including one I got wrong first

Blocks **link-local (169.254.0.0/16)**, which contains `169.254.169.254`, the cloud instance-metadata endpoint and canonical SSRF target for stealing instance credentials.

**My first draft allowed link-local, and my own test asserted that as correct.** It was caught only because the test enumerated the address explicitly rather than testing a category. Worth naming, because "block SSRF" and "allow link-local" sat in the same function without me noticing.

Also: http/https only (no `file://`), no redirect following (a permitted host can't bounce us to a blocked one), GET only, response body never read.

**The first draft was also too strict elsewhere.** It restricted probes to loopback/RFC1918/bare hostnames — checked against the 38 real nodes, that would have broken `headscale.pmoves.ai` (public DNS, own infra), `7860.localhost` (loopback subdomain), and the `*.ts` tailnet shorthands. Probing internal services across the mesh *is* the purpose of these nodes. Narrowed to the one unambiguous hazard; trees are repo-controlled and PR-reviewed, a materially different threat model from user input.

## `multiline` grep — opt-in, deliberately not default

`_check_grep` used `re.search` without `DOTALL`, so `channel-monitor.*healthcheck` couldn't span lines and failed despite the healthcheck existing.

The fix is opt-in `multiline: true` per assertion, **not** a global `DOTALL`. Globally that pattern stops meaning *"this service has a healthcheck"* and starts meaning *"channel-monitor appears somewhere, and healthcheck appears somewhere after it"* — matching when the healthcheck belongs to a different service 400 lines away. **That converts today's false negatives into false positives, which is strictly worse: a wrong PASS is invisible, a wrong FAIL is loud.**

## Verified

```
host allowlist : IMDS + 169.254.x blocked; localhost, 7860.localhost,
                 headscale.pmoves.ai, *.ts, loopback all permitted
scheme guard   : file:///etc/passwd -> fail "unsupported scheme"
live probe     : http://127.0.0.1:8077/healthz -> pass (200), via BOTH keys
down service   : -> fail  (correct for a liveness probe, not an error)
multiline      : channel-monitor.*healthcheck   default=fail   multiline=pass
```

**No regression** — `soundcloud-ingest` reports 8 pass / 6 fail / 4 pending both before and after.

## Correction to #2371's body

That PR says soundcloud reports **10 pass / 4 fail / 4 pending**. On current `origin/main` it's **8/6/4**. The 10/4/4 figure came from the main working tree, which is 39 commits behind and has a different compose file. The *"unchanged"* claim holds either way — the delta is the checkout, not the runner.

### DIFF
--- a/pmoves/configs/tac_schema.yaml
+++ b/pmoves/configs/tac_schema.yaml
@@ -14,11 +14,20 @@
 #   id: string            - Unique node identifier (dot-separated path)
 #   task: string          - Human-readable task description
 #   action: object        - Verification action
-#     type: string        - One of: file_exists, grep, command, manual
-#     target: string      - File path or command (relative to repo root)
+#     type: string        - One of: file_exists, grep, command, http, manual
+#     target: string      - File path or command (relative to repo root).
+#                           For http, also accepted as the endpoint (see url).
 #     pattern: string     - Regex pattern (for grep type)
 #     invert: boolean     - If true, finding the pattern means FAIL (negative grep). Default: false
-#     expect: string      - Expected result description
+#     multiline: boolean  - grep only. If true, `.` also matches newlines (re.DOTALL),
+#                           so a pattern may span lines. Opt-in: default false keeps
+#                           existing single-line patterns behaving as written.
+#     url: string         - http only. The endpoint to probe. Trees use both `url`
+#                           (16 nodes) and `target` (22); the runner accepts either.
+#     expect_status: int  - http only. Status code that means PASS. Default: 200.
+#                           This is the machine-readable assertion — `expect` is not.
+#     expect: string      - Expected result description. FREE-FORM PROSE, for humans
+#                           ("200 OK", "Prometheus metrics for ..."). Never parsed.
 #   context: string       - Where to look / additional guidance
 #   agent_hint: string    - Suggested agent for this task (e.g., "codex", "archon")
 #   status: string        - One of: pending, pass, fail, skip (default: pending)
@@ -30,6 +39,14 @@ schema:
     - file_exists    # Check if a file or directory exists
     - grep           # Search for a pattern in a file
     - command        # Run a shell command and check exit code
+    - http           # Liveness probe: unauthenticated GET, asserts on status code
+                     # only. http/https, no redirects, link-local blocked, body
+                     # never read. A node declaring method/headers/body/json/auth
+                     # or a content predicate (expect_body, expect_json,
+                     # expect_contains, expect_match, expect_jsonpath, assert) is
+                     # UNSUPPORTED and evaluates FAIL — the probe will not silently
+                     # send a different request and report PASS on the status alone.
+                     # Use `command` for anything needing a real request body.
     - manual         # Requires human review (always pending)
   statuses:
     - pending        # Not yet evaluated
--- a/pmoves/tools/tac_runner.py
+++ b/pmoves/tools/tac_runner.py
@@ -15,6 +15,7 @@ import json
 import re
 import ipaddress
 import shlex
+import socket
 import subprocess
 import sys
 import urllib.error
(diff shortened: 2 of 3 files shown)


POWERFULMOVES:
**Rebased onto `main` and retargeted** — base was `fix/tac-unknown-action-fails`, which I've closed as superseded by #2373.

#2373 merged at 02:12Z while this stack was open and independently made the same fail-closed change. Verified on current `main`:

```
grep -ci "unsupported action"      -> 2   (fail-closed: already there)
grep -c 'action_type == "http"'    -> 0   (http: still absent)
```

So the parent PR was duplicate work; **this one is not.** It remains the only implementation of the 38 `http` nodes and the `re.DOTALL` fix.

## Conflict resolution

The only conflict was the error-message wording. I kept **main's** phrasing (`unknown action.type: '...' (allowed: ...)`) rather than mine, and just added `http` to the allowed list since this PR implements it. No reason to churn a message that already landed.

## Re-verified post-rebase

```
IMDS blocked      : True
headscale allowed : True
file:// rejected  : True
live yt probe     : ('pass', 'http://127.0.0.1:8077/healthz -> 200')
multiline off/on  : fail / pass
```

And **#2373's own new tests still pass** — `pytest pmoves/tools/tests/test_tac_runner.py` → 4 passed. That suite arrived with the fail-closed change; this builds on it without breaking it.

POWERFULMOVES:
All four findings addressed in 0125734 — thanks, both P1s were real and the P2 host one was worse than rated.

**P1 (method/headers/body) + P1 (response predicates)** — taken the "leave them unsupported" option rather than building a request/assertion DSL under time pressure. A node declaring `method`/`headers`/`body`/`json`/`data`/`auth` or a content predicate (`expect_body`, `expect_json`, `expect_contains`, `expect_match`, `expect_jsonpath`, `assert`) now evaluates **FAIL with an explanation**, instead of being silently probed with a different request and passing on the status code alone.

`expect` is deliberately excluded from that list — it's free-form prose on all 38 existing nodes ("200 OK", "Prometheus metrics for ..."), and treating it as machine-readable would fail-close every node this PR set out to enable. Pinned by `test_prose_expect_does_not_fail_closed`.

**P2 (link-local bypass)** — confirmed and fixed. The guard compared against the canonical dotted quad only, so all of these reached the metadata endpoint:

```
2852039166                 decimal      -> was allowed
0xa9fea9fe                 hex          -> was allowed
::ffff:169.254.169.254     mapped v6    -> was allowed  (is_link_local tests fe80::/10)
```

Three layers now: numeric literal (so the block doesn't depend on the local resolver expanding them), IP literal with IPv4-mapped unwrapping, then `getaddrinfo` with **every** returned address checked — a hostname is refused if any address it maps to is blocked. All seven spellings are enumerated in tests so this can't quietly reopen; the original guard passed its own suite precisely because that suite only tried the one literal.

Documented limitation kept in the docstring: the guard's lookup and urllib's connect are separate resolutions, so DNS rebinding isn't covered. Closing that needs connecting to a pinned address, which urllib doesn't expose.

**P2 (schema)** — `tac_schema.yaml` now documents `http` and `multiline`, including which keys are unsupported and why, plus the `expect` vs `expect_status` distinction.

**Verification:** 25 tests pass (4 existing + 21 new). Guard still allows `localhost`, `7860.localhost`, `headscale.pmoves.ai` and tailnet hostnames — the targets an earlier over-strict draft broke. Test examples use RFC 5737 documentation space rather than fleet addressing.

PR #2221, opened 2026-07-25
feat(sso): SSO forward-auth gateway — auth once, access all (Phase 1: service + edge)

POWERFULMOVES:
## What

Phase-1 **SSO forward-auth gateway** — "auth once, access all." A user who signs in once against Supabase GoTrue reaches the forked apps already logged in, with Supabase as the single IdP (reuses `JWT_SECRET`/`SUPABASE_JWT_SECRET`, HS256 — **no new session secret**).

This PR lands the **service + edge + env** (plan Tasks 1–7 + 10). The per-app fork integrations (Open-Notebook middleware, Jellyfin plugin) and live cross-app e2e (Tasks 8/9/11) are split into a tracked follow-up.

## Architecture

```
Browser ──HTTPS──> Traefik (pmoves_external, TLS via Cloudflare DNS-01)
   │ Host: {auth,health,wealth,notebook,media}.pmoves.ai
   ▼ ForwardAuth ──subrequest──> pmoves-sso-auth  GET /auth/verify
   • valid pmoves_session cookie → 200 + Remote-User / X-Auth-Email / X-Auth-Subject
   • invalid/absent → 401 → 302 auth.pmoves.ai/login?rd=<orig>
```

- **`pmoves-sso-auth`** (new FastAPI service): `/auth/verify` (HS256 local verify, no network on the hot path), GoTrue login (`/login`, `/callback`, `/logout`, GitHub + email/pw), and a minimal OIDC subset for the Jellyfin plugin.
- **Traefik edge**: forward-auth middleware + per-app subdomain routers; `auth.*` router has no forward-auth (login must be reachable); `media.*` (Jellyfin) has no forward-auth (its OIDC plugin owns auth).
- **Apps**: firefly `remote_user_guard`, wger remote-user env; **all four apps de-published from host ports** so `Remote-User` can't be spoofed by bypassing the proxy.

## Security (hardened through review)

- `/auth/verify` requires `role=="authenticated"` — rejects the public anon key (closed a Critical bypass).
- Two open-redirect guards: `_safe_rd` (login flow — rejects backslash/CR-LF-TAB/protocol-relative/`@`-authority/suffix-confusion) and an exact-match OIDC `redirect_uri` allowlist bound to the auth code (closed a HIGH flagged by two independent reviews).
- Fail-closed: GoTrue outages are caught (never a 500); `/logout` always clears the cookie.
- Login template autoescaped; no secrets inlined anywhere.

## Tests

[...]

PR #2411, opened 2026-08-05
feat(edge): add up-edge — Traefik + sso-auth had overlays but no make target

POWERFULMOVES:
## The gap

`docker-compose.traefik.yml` and `docker-compose.sso.yml` have existed since #2221 with **no make target to start them**. The only reference to the traefik overlay anywhere in the Makefile is a comment. So the edge was never brought up through the pipeline.

Observed on z890:

- **No `pmoves-traefik` container has ever been created on this node.** `media.pmoves.ai` and `auth.pmoves.ai` resolve to nothing, and every `traefik.*` label in the fleet is inert.
- **`pmoves-sso-auth` is `Up 2 days (healthy)` and protecting nothing.** Sweeping every running container, **zero** carry a `traefik.http.routers.*.middlewares` label — so the `pmoves-forward-auth` middleware defined in `config/traefik/dynamic.yml` is referenced by no router.
- Services therefore fall back to their own auth. Open Notebook's `RemoteUserAuthMiddleware` is fail-closed and requires the Traefik-injected `X-Forward-Auth-Secret` header, so a direct `:8503` hit gets the password prompt **exactly as designed**.

That last point is worth stating plainly: the password prompt people keep hitting is not a bug in the app. It's the absence of an edge.

## What lands

`up-edge` / `down-edge` / `edge-health`, following the existing `EXTERNAL_DC` pattern so the edge starts through `COMPOSE_ENV_FILES` like everything else rather than by raw compose.

Both overlays go up **together** on purpose: the `auth.pmoves.ai` router is declared on the traefik container but points at `sso-auth@docker`, so Traefik without sso-auth is an edge whose own login route 404s.

`up-edge` preflights before starting, because Traefik publishes 80/443 directly and a bind failure otherwise just scrolls past in compose output:

- creates `pmoves_external` if absent (same as `up-external`)
- refuses to start if 80 or 443 is already `LISTENING`
- refuses to start if `config/traefik/dynamic.yml` is missing — the forward-auth middleware would silently not load

The port check tests the **output**, not the exit code. A pipeline ending in `head` exits 0 on no matches, so the obvious formulation reports every port as occupied. That bug produced a false *"80/443 are in use"* reading while diagnosing this, which is why the guard is written the awkward way.

`edge-health` checks each component **separately** rather than with one combined `docker ps` filter — a combined filter passes when only *one* of the two is up, which is precisely the misleading state this fixes. Current output, pre-bring-up:

```
=== Edge containers ===
  pmoves-traefik     NOT RUNNING
  pmoves-sso-auth    Up 2 days (healthy)

=== Traefik published ports ===
  (no traefik container)
  NOTE: an EMPTY [] binding list means the publish silently no-oped.

=== Routers attached to pmoves-forward-auth ===
  NONE - the middleware is defined in config/traefik/dynamic.yml but
  no router references it, so sso-auth protects nothing yet.
```

## Verified on z890

- [x] All three targets resolve; merged compose config validates
- [x] `pmoves_external` is `internal=false` so it *can* publish (unlike `pmoves_data`/`app`/`api`/`bus`, all `internal=true` — which is the real reason so many containers show empty port maps, and is **not** a bind failure)
- [x] 80/443 confirmed free by a correct check
- [x] Host port binding demonstrably works right now — grafana is bound on `0.0.0.0:3002`
- [x] `edge-health` reports the current broken state accurately
- [ ] Actual bring-up — deliberately not run; it publishes 80/443 and stands up a public-facing edge

**Caveat surfaced by the substrate probe:** z890 has a same-subnet ghost adapter (two active NICs on one /24). That's the documented cause of Docker Desktop *silently* failing port binds — see `SAME_SUBNET_GHOST_PATTERN.md`. It is not currently manifesting (grafana binds fine), so `edge-health` notes it rather than hard-blocking. A plain `netstat` check would never have surfaced this.

## Not in this PR

**Attaching `pmoves-forward-auth` to any router.** That's a security-behaviour change on protected compose files, needs a `compose:` Known Road, and needs a per-service decision — notably whether Jellyfin uses gateway forward-auth or the in-app OIDC plugin from #2407, since running both double-prompts.

## Operator prerequisites

The secrets manifest is zero-access (correctly), so I could not verify these — please confirm before `up-edge`:

- `CLOUDFLARE_DNS_API_TOKEN` — the ACME DNS-01 resolver (`certresolver=cf`) needs it or cert issuance fails
- `SSO_FORWARD_AUTH_SECRET` — without it, header trust can never engage and forward-auth stays inert even once wired

Related: #2407 (Jellyfin in-app OIDC), #2221/#2229 (the gateway this finally starts).

## Summary by CodeRabbit

* **New Features**
  * Added commands to start and stop the Traefik and SSO authentication edge services.
  * Added pre-start validation for required ports and configuration.
  * Added health diagnostics showing container status, published ports, and authentication router usage.

### DIFF
--- a/pmoves/configs/tailscale-acl-policy.json
+++ b/pmoves/configs/tailscale-acl-policy.json
@@ -117,6 +117,41 @@
       ],
       "users": ["root", "autogroup:nonroot"]
     },
+    // Fleet-management exception: tag:lab nodes may root-SSH the VPS/exit nodes.
+    //
+    // WHY THIS EXISTS (deliberate, scoped deviation from "tags never get root"):
+    // the management/inference workstations (z890, B850-AI-TOP, SPARK) are TAGGED
+    // compute nodes — tag:gpu+tag:lab, because they also serve Ollama to the fleet.
+    // A tagged device's SSH identity IS the tag; it has no user identity, so it can
+    // NEVER match the autogroup:admin/owner rule above. Without this rule no fleet
+    // node can reach the KVM exit nodes as root, which is what has kept the Mullvad
+    // L4 upstream from #1945 unactivated (exit traffic still leaves the raw Hostinger
+    // uplink) and blocks exit-node observability maintenance.
+    //
+    // WHY NOT UNTAG THE WORKSTATIONS INSTEAD: that was tried. #1972 proposed this
+    // same rule and was closed in favour of untagging pmoves-4090 so it would match
+    // autogroup:owner. Measured 2026-08-04, that untag silently removed 4090 from the
+    // fleet: from a tagged node, TCP to PMOVES-4090:11434 fails to connect while the
+    // identical request to still-tagged pmoves-5090:11434 returns 200 in 23ms — the
+    // node pongs fine at the disco layer, so this is the ACL, not a dead service.
+    // Untagging drops a node out of tag:gpu (a DESTINATION in five rules: fleet
+    // Ollama :11434, gpu-orchestrator :8200, partner :3030/:8080/:8081, guest :8081)
+    // and out of the tag:pmoves/tag:lab src list for autogroup:internet, which also
+    // costs it exit-node egress — the owner rule's "*:*" does not cover
+    // autogroup:internet. Repeating that on z890/B850/SPARK would take the whole
+    // inference pool offline to buy SSH access this one rule provides for free.
+    //
+    // SCOPE: src tag:lab (the workstations and the KVMs themselves — fleet infra).
+    // dst is ONLY tag:vps + tag:exit, the KVM concentrators. NOT tag:pmoves or
+    // tag:gpu — this grants no root anywhere else on the mesh. Silent accept, no
+    // check mode: a tagged source has no user to re-auth as. The network layer
+    // already permits port 22 via the tag:pmoves full-mesh rule above.
+    {
+      "action": "accept",
+      "src": ["tag:lab"],
+      "dst": ["tag:vps", "tag:exit"],
+      "users": ["root", "autogroup:nonroot"]
+    },
     // Tenants SSH into their OWN devices in check mode (browser re-auth). checkPeriod
     // omitted: it is Premium/Enterprise-gated in the policy-syntax reference and the
     // value we wanted (12h) is already the default, so omitting it keeps the policy


POWERFULMOVES:
Both P2s were correct and are fixed in `2d2bc29`→latest. One implemented differently than suggested, with reasoning:

**P2 #1 — missing external networks.** Confirmed: the SSO overlay declares `pmoves_app` and `pmoves_api` as external, and compose fails network-not-found before Traefik starts.

Fixed, but **not** by creating them. Both are `internal: true` in the core stack (`docker-compose.yml:5364,5373`, confirmed at runtime `internal=true`). A `docker network create --driver bridge` here would materialize them **non-internal**, and the core stack would then attach to a network that no longer blocks egress — converting a loud missing-network error into a quiet security regression. `up-edge` now requires them to pre-exist and points at the core stack. `pmoves_external` is still created here: it's ours, non-internal by design, and `up-external` already does the same.

**P2 #2 — non-idempotent rerun.** Confirmed and a real one-shot bug: the preflight treated our own Traefik as a conflict, so a rerun to pick up config changes was blocked by its own listener. Now skipped when a `pmoves-traefik` container is already running — compose handles the recreate and port handoff. First bring-up still gets the full check.

Verified: all three targets resolve; the missing-network branch fires on an absent network; with no Traefik running the preflight still executes.

POWERFULMOVES:
Review addressed. Three fixed, one deferred with reasoning.

**✅ OS-aware port probe** (Minor — but a real bug I introduced). `LISTENING` is Windows netstat's wording; Linux/BSD print `LISTEN`. So on the Linux hosts where the edge actually runs, the probe matched nothing, treated 80/443 as free, and protected nothing — precisely the failure it exists to prevent. Now prefers `ss`, falls back to `netstat` matching **both** spellings.

One deliberate deviation: the review asked to *fail* when no probe is found. I warn and continue instead — a missing probe isn't evidence of a conflict, and hard-failing would block a legitimate bring-up on a minimal host. Compose still surfaces a bind error there.

**✅ Router-label reader in `edge-health`** (Minor). It grepped the whole inspect blob for the literal string, so any env var containing `pmoves-forward-auth` counted, and it printed only the container name — useless on a container serving several routers. Now parses `traefik.http.routers.<name>.middlewares` and prints container + router + chain. Positive-controlled with the same pipeline against `.rule`, which correctly yields `router=media` on pmoves-jellyfin.

**✅ Operator documentation** (Major). New `docs/operations/EDGE_TRAEFIK_SSO_RUNBOOK.md` — targets, required secrets and what breaks without each, network prerequisites and why `pmoves_app`/`pmoves_api` are deliberately *not* auto-created, port ownership and idempotent-rerun behaviour, run and rollback steps (including why not to delete the acme volume), troubleshooting table, and an explicit Known Gaps section.

Also split the preflight into a standalone `edge-preflight` target — flagged by checkmake for body length, and a prerequisite check you can run without starting anything is more useful on its own.

**⏸️ Deferred — CWE-319 on the ForwardAuth hop** (Major). The finding is fair: the hop is `http://` and `authRequestHeaders` is unset, so Traefik forwards every incoming header to the verifier. Two reasons not to change it here:

1. It's **pre-existing** config in `config/traefik/dynamic.yml` from #2221, not introduced by this PR. This PR adds a make target; folding a middleware security change into it couples an unrelated behaviour change to the bring-up.
2. **Getting the allowlist wrong breaks login, and the happy path has never run once** — Traefik has never started on this fleet. Tightening auth transport before we can observe a working login makes any breakage indistinguishable from the many other things not yet wired.

Right sequence: land the edge → prove one login end-to-end → then tighten the hop against a working baseline. Recorded in the runbook's Known Gaps so it isn't lost.

**Verified:** all four targets resolve; `edge-preflight` passes on z890 (networks present, 80/443 free, `dynamic.yml` present); `edge-health` correctly reports no protected routers **and** now explains that `external.yml` does set them while the running containers predate the labels.

PR #2440, opened 2026-08-06
fix(ui): cookie-based SSR auth for real user login

POWERFULMOVES:
## Summary

The browser client stored auth sessions in localStorage while the server middleware and API routes used cookies via `@supabase/ssr`. These two paradigms never shared session state — login appeared to succeed (GoTrue returned 200) but the server couldn't see the session, redirecting back to `/login` indefinitely.

## Changes

- **`supabaseClient.ts`**: Add `syncSessionToCookie()` — bridges localStorage sessions to cookies by writing the `sb-<host>-auth-token` cookie on auth state changes and synchronously after `signInWithPassword` (before `router.replace` runs)
- **`useSupabaseAuth.ts`**: Call `syncSessionToCookie(session)` after successful password login + `syncSessionToCookie(null)` on signOut
- **`api/rooms/route.ts`**: Switch from `ownerFromJwt` (boot-JWT-only) to cookie-based `createSupabaseRouteHandlerClient` so the route works with real user sessions
- **`proxy.ts`**: Add `/persona` to public routes so the living-doc is viewable without authentication

## Testing

Verified via Playwright: login with `pmoves@pmoves.ai` sets cookie synchronously, redirect to `/dashboard/ingest` works immediately, all 23 dashboard routes return 200, `/api/rooms` returns 200 (was 401 with boot JWT auth).

💘 Generated with Crush


POWERFULMOVES:
## Superseded — and its one remaining delta is a security regression

Rebased this branch onto `main` locally to clear the CONFLICTING state. The result was an **empty branch**: `HEAD == origin/main` exactly, zero commits, zero diff.

**Three of the four files already landed on main, byte-identical:**

```
pmoves/ui/hooks/useSupabaseAuth.ts     IDENTICAL on main
pmoves/ui/lib/supabaseClient.ts        IDENTICAL on main
pmoves/ui/app/api/rooms/route.ts       IDENTICAL on main
```

The cookie-based SSR auth work is **done and shipped**. Only `proxy.ts` conflicted, and that delta should not land:

```diff
- const MARKETING_ROUTES = ['/community'];                    # main
- const PUBLIC_PERSONA_PREFIX = '/persona/livingdoc';         # main
+ const MARKETING_ROUTES = ['/community', '/persona'];        # this PR
```

Main deliberately **narrowed** `/persona` to `/persona/livingdoc`, with the reason inline: *"P1 fix: /persona/livingdoc is declared public in persona.room.livingdoc.json (access.visibility=public, owner_only=false). Must be exempt from auth."*

This PR would re-widen it to the **entire `/persona` tree**, making every persona route public — reverting that P1.

**It also wouldn't compile.** `proxy.ts:13` (unconflicted, from main) reads:

```ts
const PUBLIC_PATH_PREFIXES = [...MARKETING_ROUTES, PUBLIC_PERSONA_PREFIX];
```

Taking this PR's side removes `PUBLIC_PERSONA_PREFIX` while line 13 still references it — a build break on top of the auth widening.

So the conflict resolution is unambiguous: **keep main.** Which leaves this branch with nothing.

**Closing as superseded** — reopenable if I've misread. Nothing is lost: the SSR auth fix is on main, and the only thing not carried forward is a change that would have re-opened a P1 and broken the build.

agent_signature (advisory, unsigned-local): `ACK::CLAUDE-OPUS-5::PR2440-SUPERSEDED-2026-08-10`

POWERFULMOVES:
Superseded — see analysis above. Content is on main; the only remaining delta reverts a P1 and breaks the build.

PR #2451, opened 2026-08-06
fix(edge): let the PMOVES-Health phone app through without opening an auth bypass

POWERFULMOVES:
## Why

Once `health.pmoves.ai` goes behind SSO, the **PMOVES-Health phone app stops working.** wger's Traefik router has no path constraint, so `/api/v2/` is guarded too — the app sends `Authorization: Token <key>`, forward-auth finds no `pmoves_session` cookie, returns 401, and `pmoves-auth-redirect` turns that into a login-page redirect. The app receives HTML instead of JSON.

Same failure class as Jellyfin's native clients, which is why Jellyfin is already excluded from forward-auth (`external.yml:210-212`).

## Why the obvious fix is dangerous

Simply exempting `/api/` is a **full authentication bypass.** wger runs `WGER_ALLOW_REMOTE_USER=True` with `WGER_REMOTE_USER_HEADER=Remote-User`, and Django's `RemoteUserMiddleware` is **not path-scoped**. The only thing preventing header forgery today is `authResponseHeaders`, which overwrites `Remote-User` — and that exists **only on routes that actually run forward-auth.**

Verified there is no second line of defence: the mounted wger nginx conf sets only `Host`/`X-Real-IP`/`X-Forwarded-*` and forwards unknown hyphenated headers; `config/traefik/dynamic.yml` had no header-strip middleware at all. So a bare exemption would let an unauthenticated internet client send `Remote-User: <anyone>` and be that user with no token.

## The change

**1. `strip-client-auth-headers` middleware** (`config/traefik/dynamic.yml`) — clears `Remote-User`, `X-Auth-Email`, `X-Auth-Subject`, `X-Forward-Auth-Secret` from **inbound** requests. Applied **first** on every router.

On forward-auth routes this changes nothing observable — strip runs, then forward-auth re-adds the trusted values. What it buys is that the invariant no longer depends on every future author remembering to attach forward-auth to a new router.

**2. A second router for the API path** (`docker-compose.external.yml`) — `health-api`, explicit `priority=100`, strip middleware, **no** forward-auth. Explicit priority rather than relying on Traefik's rule-length ordering, so intent survives a future edit to either rule.

## What this deliberately accepts

`/api/v2/` becomes internet-reachable with only wger's token auth in front of it. That is the necessary trade for native-app support and the same trade Jellyfin already makes — the API is still authenticated, just not by the gateway. The alternative (issue app tokens from `sso-auth`, drop wger's own token auth) needs the multi-client work `sso-auth` does not have yet — `oidc.py:77-80` is hardcoded to a single client.

## Template value

Every app with native clients needs this exact shape: strip inbound auth headers globally, exempt the API path, let the app's own token auth run. Getting it right here sets the pattern for Jellyfin and anything else with a mobile client.

## Verification

- `docker compose config` parses with both routers present.
- After deploy: `curl -H 'Remote-User: someone' https://health.pmoves.ai/api/v2/...` must **not** authenticate.
- Browser gets one SSO login; phone app authenticates with its token.

Full brief: `pmoves/docs/handoffs/wger-api-forward-auth-exemption-2026-08-05.md`

## Summary by CodeRabbit

* **New Features**
  * Added direct API access for the PMOVES-Health mobile app without browser-based authentication redirects.
  * Added protection that removes untrusted authentication headers before processing requests.
  * Preserved trusted authentication values returned by the authentication service.

* **Documentation**
  * Documented API authentication behavior, security considerations, reuse guidance, and verification steps.

### DIFF
--- a/pmoves/config/traefik/dynamic.yml
+++ b/pmoves/config/traefik/dynamic.yml
@@ -1,5 +1,32 @@
 http:
   middlewares:
+    # Remove client-supplied auth headers from INBOUND requests. Put this FIRST on
+    # every router, forward-auth or not.
+    #
+    # Why it must exist: apps behind this edge trust Remote-User. wger runs
+    # WGER_ALLOW_REMOTE_USER=True with WGER_REMOTE_USER_HEADER=Remote-User, and
+    # Django's RemoteUserMiddleware applies to EVERY path — there is no path scoping.
+    # Until now the only thing stopping header forgery was authResponseHeaders below,
+    # which overwrites these values — but that protection exists ONLY on routes that
+    # actually run forward-auth. Any route that skips it (an API exemption for a
+    # native mobile client, say) would let an unauthenticated client send
+    # `Remote-User: <anyone>` and be authenticated as that user with no token.
+    #
+    # Neither of the other candidate chokepoints strips it: the mounted wger nginx
+    # conf sets only Host/X-Real-IP/X-Forwarded-* and forwards unknown hyphenated
+    # headers, and nothing else in this file touched inbound headers.
+    #
+    # On forward-auth routes this is a no-op in effect: strip runs first, then
+    # pmoves-forward-auth re-adds the trusted values from the auth response. What it
+    # buys is that the invariant no longer depends on every future author remembering
+    # to attach forward-auth to a new router.
+    strip-client-auth-headers:
+      headers:
+        customRequestHeaders:
+          Remote-User: ""
+          X-Auth-Email: ""
+          X-Auth-Subject: ""
+          X-Forward-Auth-Secret: ""
     pmoves-forward-auth:
       forwardAuth:
         address: "http://pmoves-sso-auth:8080/auth/verify"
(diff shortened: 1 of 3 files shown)


POWERFULMOVES:
Closing as a duplicate — **#2419 already shipped this**, and I missed it before opening.

`origin/main` at `cb6de51bf` (`fix(traefik): strip-client-auth-headers + wger API auth-bypass protection`) already contains:

- the `strip-client-auth-headers` middleware in `config/traefik/dynamic.yml`
- the `health-api` router with `priority=100`, `service=health`, and strip-only middleware
- `strip-client-auth-headers@file` prepended to the `health` router chain
- the handoff doc `pmoves/docs/handoffs/wger-api-forward-auth-exemption-2026-08-05.md`

Diffing this branch against main shows **no functional delta** — only comment wording. Nothing here is worth a merge conflict.

One thing verified along the way that is worth recording: the **running** `pmoves-wger` container (created 2026-07-25) predates #2221/#2419 and has **zero traefik labels**, so none of this is live yet. Forging `Remote-User` against the direct `:8010` port today has no effect — confirmed behaviourally, the only response delta between a forged and an unforged request is the per-request CSRF nonce, identical to the delta between two unforged requests. `WGER_ALLOW_REMOTE_USER` is not active on the running container.

That protection becomes load-bearing the moment wger is recreated. Sequencing captured in the edge lane.

PR #2470, opened 2026-08-07
fix(evo): SSL_CERT_FILE override + up-evo target — EVO SWARM boots on :8113

POWERFULMOVES:
## Problem
`evo-controller` crash-looped on startup:
```
ssl.create_default_context(cafile=os.environ["SSL_CERT_FILE"])
FileNotFoundError: [Errno 2] No such file or directory
Application startup failed. Exiting.
```
The shared env_file leaks a **Windows** `SSL_CERT_FILE` path; httpx tries to load it inside the Linux container and dies. ~10 other services already neutralize this with an empty override — evo-controller's stanza was missing it.

## Fix
- Add `SSL_CERT_FILE=` / `REQUESTS_CA_BUNDLE=` (empty) to the evo-controller stanza (house pattern → httpx falls back to bundled certifi).
- Add `up-evo` make target (no ghcr image published → builds locally, loads `env.shared` for CHIT prod signing), mirroring `up-cipher-nobuild`.

## Verified
EVO SWARM controller now **healthy on :8113** (`/healthz → 200`), publishing `evoswarm.training.genome.v1`.

Follow-up (separate): the existing `geometry_swarm` migration (`geometry_parameter_packs`) needs applying to the running Supabase — evo currently 404s on pack upsert until then.

## Summary by CodeRabbit

- **New Features**
  - Added a streamlined command for building and starting the EVO SWARM controller with shared environment settings.

- **Bug Fixes**
  - Improved controller connectivity by preventing host certificate-path settings from interfering with container HTTP requests.
  - Aligned edge services with the primary Compose project for more consistent service management and predictable startup behavior.

### DIFF
--- a/pmoves/docker-compose.agents.yml
+++ b/pmoves/docker-compose.agents.yml
@@ -1015,6 +1015,9 @@ services:
     environment:
     # NATS_URL from env.tier-agent has credentials: nats://nats:***@nats:4222
     - PORT=8113
+    # neutralize the leaked Windows SSL_CERT_FILE / REQUESTS_CA_BUNDLE (house pattern — httpx falls back to bundled certifi)
+    - SSL_CERT_FILE=
+    - REQUESTS_CA_BUNDLE=
     - SUPA_REST_URL=${SUPA_REST_INTERNAL_URL:-${SUPA_REST_URL:-http://supabase-kong:8000/rest/v1}}
     - EVOSWARM_POLL_SECONDS=${EVOSWARM_POLL_SECONDS:-300}
     # CHIT fail-closed signing (house pattern, mirrors hi-rag-gateway*)


POWERFULMOVES:
## Review — 4090-claude

Core fix is right and well-evidenced: the leaked Windows `SSL_CERT_FILE` is a real crash cause, the empty-override is the established house pattern (~10 services already do it), and `/healthz → 200` on :8113 publishing `evoswarm.training.genome.v1` is the correct proof. **Two things to resolve before merge.**

### 1. BLOCKING — split overlays not regenerated (this is the red check)

`Validate Compose Files` is failing, and it isn't flaky — it **passes** on #2456, #2468, #2469, including one that edits `docker-compose.yt-egress.yml`.

Cause is the drift gate at `hardening-validation.yml:361`:

> The split overlays (`docker-compose.{base,core,agents,media,ui,workers,apps}.yml`) are generated from `docker-compose.yml`. Regenerate and fail on any drift.

Verified against this branch:

```
docker-compose.yml         → SSL_CERT_FILE=  present
docker-compose.agents.yml  → evo-controller stanza has NO SSL_CERT_FILE
```

`evo-controller` is an `agents`-profile service, so its stanza exists in `docker-compose.agents.yml` as well. **Anyone bringing the stack up via the `up-overlay-*` path still gets the crash-looping config** — the fix currently only applies through the monolithic compose file.

**Fix:** `make -C pmoves compose-split`, commit the regenerated overlays.

Same class as the #2358 ratchets — *"the file the developer touched ≠ the file the runner builds."* The gate did its job.

### 2. NON-BLOCKING — unrelated `EDGE_DC` change

```diff
-EDGE_DC := docker compose -p $(PROJECT)-edge ...
+EDGE_DC := docker compose -p $(PROJECT) ...
```

This moves the edge stack (Traefik + SSO) out of its own compose project into the main one. Unrelated to EVO SWARM or SSL, and it carries blast radius — a shared project name changes `down` semantics and orphan-pruning across both stacks.

**#2453 is open right now: "up-edge collided with the running stack, and Traefik…"** — same territory. Two PRs adjusting edge project scoping independently is how one silently reverts the other.

Suggest splitting it out to land with #2453's reasoning, or stating explicitly that it supersedes that half.

### Nit

The `geometry_parameter_packs` migration follow-up is well-flagged — worth an issue rather than only a PR-body line, since a 404 on pack upsert is the kind of thing that gets rediscovered as a bug.

---

Otherwise clean: `up-evo` correctly mirrors `up-cipher-nobuild`, loads `env.shared` for CHIT prod signing, builds locally since no GHCR image is published. Good to go once the overlays are regenerated.

*(Posted as a comment — GitHub blocks request-changes on one's own PR.)*

PR #2493, opened 2026-08-08
feat(nats): wire PMOVES-nats-server fork as submodule (§7a step 1)

POWERFULMOVES:
## What

Registers the `PMOVES-nats-server` fork (fork of `nats-io/nats-server`) as a git submodule at repo root, matching the other `PMOVES-*` fork convention. Gitlink at upstream `main` tip; tracks upstream for CVE currency.

## Why

First concrete step of adopting the NATS accounts + leaf-topology spec (#2492). The fork was real but **decorative** — zero PMOVES commits, not a submodule, not the running image (everything runs upstream `nats:2.11.8-alpine`). This makes it a first-class part of the tree so §7a's config/packaging can be added to it.

## Scope / gates

- **Upstream-clean in v0** (no Go patches).
- The PMOVES config + packaging per §7a (`conf/pmoves-nats.conf` memory-resolver template, thin Dockerfile, compose stanza) are added **to the fork** in a follow-up, once `nsc` mints the four SYS/CORE/EDGE/CLOUD account JWTs — that lane is gated on **nsc install + the #1901 custody decision**.

Part of the multi-slice rollout; see the runbook (follow-up) for phase sequencing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2492, opened 2026-08-08
spec(nats): accounts + leaf-topology v0 — adopt the real design (retire the WSL2-hack leaf)

POWERFULMOVES:
## Decision: adopt real NATS leaf topology + wire the fork

Operator decision 2026-08-08: **adopt this spec + wire the `PMOVES-nats-server` fork.**

## Why (forensic-grounded)

A git-history forensic established there is **no coherent leaf in production today**:
- The `nats-leaf` pattern originated 2026-03-19 (#1034) purely as a **Docker-Desktop/WSL2 routing workaround** for z890 ("containers cannot route to LAN/Tailscale IPs") — verified working *one day*, hub side reverted 3 days later (`feafffae6`), leaving a dangling stale overlay.
- The June decision (#1831/#1824) is a **hub-and-spoke of independent standalone brokers on KVM4-2**, reached by direct client URL — explicitly **NOT leaf-federated** (per `CATALOG.md`).
- Every later leaf config (`elder-melchor-leaf.conf`, jetson `nats-leaf`) copied the March hack; none ever worked (`elder-melchor` points at client port 4222, not a 7422 leaf port).
- **The fork** (`PMOVES-nats-server`) is real but decorative: zero PMOVES commits, not a submodule, not the running image (upstream `nats:2.11.8-alpine`).

This spec is the **first and only** real design (SYS/CORE/EDGE/CLOUD trust-zone accounts, nsc/JWT, memory resolver, a real 7422 leafnode listener, CHIT signing as an orthogonal payload-trust layer).

## This PR

The design doc only. Implementation is phased (§9) and mostly **operator-gated**:
- **Phase 1 (mint)** needs `nsc` installed + the custody decision per #1901 (Docker file-mounted secret v0 → CHIT-vault follow-up).
- Compose/Dockerfile/config edits are **Known-Road (operator-authorized)**.
- `secrets_manifest.yaml` entries (§8) are zero-access — operator adds via the example→manifest→funnel pipeline.
- EDGE/CLOUD leaf attach (§9.4–5) needs cross-node access.

Rollout runbook + submodule wiring land in follow-up PRs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Added a rollout runbook for NATS account and leaf topology deployment.
  * Documented trust-zone accounts, credential management, secrets handling, and hub/leaf configuration.
  * Added validation criteria for connectivity, isolation, exports, message signing, monitoring, and credential rotation.
[...]

PR #2496, opened 2026-08-08
feat(nats-mcp): CORE .creds auth + CHIT-signed publish (§7b)

POWERFULMOVES:
## What

Upgrades `pmoves-nats-mcp` from a shim to a first-class two-layer-trust citizen (spec §7b):

- **CORE-account auth** — `NATS_CREDS` env → `nats.connect(user_credentials=…)`; binds to the CORE account instead of the legacy `nats:pmoves` plaintext. Back-compat: falls back to `NATS_URL` creds when unset.
- **CHIT-signed publish** — `nats_publish` gains `sign?`; CHIT-aware subjects (`chit.*`, `tokenism.prosodic.*`, `geometry.*`) or `sign=true` are signed via the **canonical** signer `pmoves.tools.chit_security.sign_cgp` — **imported, never re-typed** (4090's "name it, don't re-type it"). **No fake CHIT**: signs only when the real signer succeeds; otherwise marks `X-CHIT-Signed: false` (unsigned-local, honest). `nats_subscribe` surfaces the header so the consumer verifies the two independent gates.
- **Packaging** — `Dockerfile` (uv, reproducible) + `examples/README.md` (the `.claude/mcp.json` wiring with `NATS_CREDS` + `PYTHONPATH` for the canonical import).

## Verification

`py_compile` clean. The signing seam is honest by construction (transport gate = account `.creds`; provenance gate = canonical signature; unsigned-local is marked, never faked).

## Gates / next

- Real signing needs the canonical signer importable (`PYTHONPATH=<repo root>`) + a signing key — else honest unsigned-local. CORE `.creds` come from Phase 1 mint (done, in vault). Wiring `.creds` into `.claude/mcp.json` + the compose swap is the deploy step.

Part of the NATS leaf-topology adoption (spec #2492, submodule #2493, fork PMOVES-nats-server#1).


POWERFULMOVES:
## Trim: both defects fixed (1a61782b1)

### P1 — `chit_signed` inverted the security signal it exists to carry

Publish returned a **bool**; subscribe returned the **raw NATS header**, which is always a string. The honest verdict for an unsigned message is `"false"` — and that string is truthy:

```
X-CHIT-Signed='true'   old='true'  truthy=True  | new=True
X-CHIT-Signed='false'  old='false' truthy=True  | new=False    <-- the bug
X-CHIT-Signed=None     old=None    truthy=False | new=False
```

`bool("false")` is `True` in Python and `Boolean("false")` is `true` in JS, so any consumer writing the obvious `if msg["chit_signed"]:` read **every unsigned message as signed**. An agent auditing the bus would have concluded it was fully signed while nothing was — the failure mode is silent and points the wrong way, which is the worst shape for a security signal.

Normalised at the subscribe boundary to `== "true"`, so subscribe reports the same type publish does. `X-CHIT-Reason` is now surfaced as `chit_note` for parity with the publish path, so the *reason* a message is unsigned survives too. `examples/README.md` documented the string comparison (`chit_signed == "true" | "false"`), so it is corrected as well — otherwise the doc would have kept teaching the bug.

### P2 — `_publish` would have TypeError'd the integration check (Codex, correct)

`_publish` took `headers` and `sign` as mandatory, but `.claude/hooks/test/run_integration.sh:74` calls it as `_publish("$SUBJECT", '{"hello":"world"}')`. Every integration run with NATS actually available would raise `TypeError` before opening a connection and report the round trip as failed — i.e. the one check most likely to catch a regression on this path would have been the first thing to break.

Defaults restored (`headers=None`, `sign=False`), with a docstring naming the caller so they don't get dropped again. Required args are now exactly `['subject', 'payload']`, verified by parsing the signature rather than eyeballing it.

`handle_publish` already had these defaults — only the internal helper had lost them.

## Not fixed here, on purpose

The **CHIT-aware prefix set** finding (`chit.`, `tokenism.prosodic.`, `geometry.`) — missing `tokenism.cgp.ready.v1`, `tokenism.cgp.weekly.v1`, `tokenism.geometry.event.v1`, while sweeping in non-CGP `geometry.health.v1` — is a **spec-level** issue and belongs on **#2492 §7b**. Changing the prefix set here would fork the definition from the spec that governs it and make both harder to reconcile. It should land there first, then be picked up here.

## Red check

`emit lifecycle trail`, and not this PR's doing: `actions/checkout` runs `git submodule foreach --recursive` even with `submodules: false` and dies with exit 128 on the `PMOVES-Archon` gitlinks that have no `.gitmodules` entries. Fix is `POWERFULMOVES/PMOVES-Archon#24`, awaiting operator approval.

PR #2514, opened 2026-08-10
docs(ops): JuiceFS media MinIO-reformat + whole-drive storage runbooks

POWERFULMOVES:
Two operator storage runbooks captured from local working-tree state. Docs only, no literal IPs.

- **JUICEFS_MEDIA_MINIO_REFORMAT_RUNBOOK.md** — reformat the `pmoves-media` JuiceFS volume onto tailnet MinIO (the fix for the file-backed single-node blocker documented in the cross-node storage handoff).
- **WHOLE_DRIVE_STORAGE_OPTION_A_RUNBOOK.md** — Option A: SMB media → Jellyfin + JuiceFS-on-`N:` for new content.

Part of the fleet-pass local-work landing; sibling to the cache-guard/runbook PRs #2501/#2502.


POWERFULMOVES:
## Trim: all 5 threads were correct, all 5 fixed (272016e48)

Every finding verified against the tree before acting. Codex was right on all five, and the red check was **not** what it looked like.

### The red was not `emit lifecycle trail`

It was **`validate-command-anchors-ratchet`**, and it was this PR's own content — 7 new anchor findings. Reproduced locally rather than inferred:

```
Anchor findings: 442 total, 435 baselined, 7 new
  GHOST_TARGET  JUICEFS_MEDIA_MINIO_REFORMAT_RUNBOOK.md -> juicefs-mount-pg
  UNKNOWN_HOST  ... -> pmoves-b850-ai-top   (x5)
  UNKNOWN_HOST  ... -> pmoves-nano-1
```

Now: **`PASS — 442 baselined, 0 new`**.

### What was wrong

**1. Step 2 formatted the wrong database (P1).** Step 1 closes its `ssh` block, so Step 2's `docker run` executed on whatever node the operator was sitting on — while `supabase-db` is a node-local Docker alias on `pmoves_data`. Run from z890 it formats an unrelated local database, or fails, *after* Step 1 has already destroyed b850's volume. A destructive runbook that cannot recover what it destroyed. Steps 2 and 2a are now explicitly on b850; the MinIO endpoint stays z890 by Tailscale hostname.

**2. `juicefs-mount-pg` does not exist (P1).** Confirmed by grep across `pmoves/Makefile` + `pmoves/mk/*.mk`, and independently by the failing ratchet. b850 holds the metadata DB locally, so the correct target is **`juicefs-mount-local`** — it is the one that actually implements the `META_PASSWORD` form the surrounding prose promises (`mk/egress.mk:360`).

**3. The jetson mount could not have worked (P1).** `sudo juicefs mount` against `supabase-db` is a native client outside b850's Compose network — the alias is not DNS and cannot resolve — and the DSN carried no password. Both remote nodes now go through `juicefs-cross-node-setup`, whose documented contract is exactly this case: its `JUICEFS_HOST` **defaults to `pmoves-b850-ai-top`**, and it passes the secret via `META_PASSWORD` so it never reaches argv or `ps`.

**4. `up-jellyfin` cannot show the Startup Wizard (P2).** Its recipe is only `up -d jellyfin-bridge` (`Makefile:3186`). Corrected to `up-jellyfin-ai`; `up-external` also works.

**5. Raw `docker compose up` for MinIO (P1).** Blocked by the damage-control hooks, and it skips env-file and restart handling at the one moment the object store is mid-move. Routed through `up-minio`.

### Cross-references, not duplicated work

- **JuiceFS cache bounds belong to #2501, not here.** I did not touch them. Note the fix above *removes* this doc's hand-rolled `juicefs mount --enable-xattr` entirely by routing through the canonical target, so the unbounded-call-site inventory on #2501 is unaffected — the remaining one is still `docker-compose.juicefs-mount.yml:34`.
- `pmoves-b850-ai-top` and `pmoves-nano-1` are real hosts the fleet config does not declare, so they are recorded in `_known_gaps.yaml` — the mechanism this repo already uses for `pmoves-b850` (HERMES_AGENT_INTEGRATION.md) and `pmoves-nano` (FLEET_INVENTORY_LIVE.md).

### One thing I found but did not change

`pmoves/config/operator_nodes.yaml` registers the B850/Knuckles node as `reach: "knuckles"`, but its documented Tailscale hostname is `pmoves-b850-ai-top` — that is what `.claude/context/runner-topology.md:12` states, what 26 places in the repo use, and what `juicefs-cross-node-setup.sh` defaults to. The registry's own header says `reach = Tailscale hostname`, so this looks like real drift, and it is the underlying reason the anchor validator does not recognise the host.

I left it alone: changing a `reach` value could break whatever resolves `knuckles`, I cannot verify from here which names actually resolve on the tailnet, and it is not this PR's concern. Worth its own issue.

PR #2516, opened 2026-08-10
spec(value-engine): value-engine domino v0

POWERFULMOVES:
v0 spec for the value engine, captured from local working-tree state. Docs only.

Value = future-error-reduction; verified work → Archon-mints → domino cascade → Known Roads pave → ToKenism → Wealth/DoX/BoTZ, with trails as victory stories.


POWERFULMOVES:
## Trim: all 4 threads correct, spec corrected (71cb48d83)

Each finding verified against the tree. All four hold, and two of them would have stopped an implementer cold.

**1. "Trigger (exists today)" — it does not (P1).** The three `archon.mint.*` subjects are registered in `contracts/topics.json` and have schemas, but **nothing publishes them**: searching `*.py`/`*.ts`/`*.js`/`*.go` returns zero hits, and every match in the repo is a doc, a slash command, or a schema. `.claude/PATTERNS.md` places the flow between the slash commands and `archon-qa-agent`; `model-lifecycle-pipeline-2026-07-30.md` marks the stage a stub. The Archon NATS bridge is now stated as a v0 **prerequisite**, in the header, the prose, and the components table.

**2. The trigger event cannot carry the scoring inputs (P1).** `archon/mint.confirmed.v1.schema.json` is `additionalProperties: false` over exactly two fields:

```jsonc
{ "agent_id": "<uuid>", "confirmed_at": "<date-time>" }
```

No `error_class`, `estimated_reduction`, `contributors`, or `trail_ref` — and `mint.agent.v1` does not supply them either. An implementer following the old text **could not have produced a Domino Record from the event the spec tells them to subscribe to.** The spec now requires either a correlated domino-candidate contract or a named lookup source per field, and notes that widening `mint.confirmed.v1` is the wrong lever — it is a QA-gate signal, not a value record.

Codex also caught that this spec contradicts itself: its own "first real domino" is **#2464, a PR recovery** — not an agent mint, and it would never arrive on `archon.mint.confirmed.v1`. Recorded.

**3. The value metric inflated itself (P1).** Step 1 derives `error_weight` by counting `error_class` in `known-roads.jsonl`; Step 2 appends a domino line carrying that same `error_class` to that same file. So the second domino for a class counts the first one's **paved road** as another occurrence of the error. Weight and reported value grow every time a preventive pattern is scored, with no new error having happened — a metric defined as "reduces FUTURE error" that pays out for scoring. Now counts typed incident records only, and the acceptance test pins it: score two dominos with the same `error_class`, assert the first one's weight is unchanged.

**4. `tokenism.value.recorded.v1` was unregistered (P2).** Absent from `contracts/topics.json`, no `*.v1.schema.json`. Canonical envelope validation and subject auditing could not recognise it. Registration and a versioned schema are now v0 deliverables *and* acceptance criteria.

Spec text only — no implementation, status stays `draft for operator review`. `validate-command-anchors-ratchet` still passes (435 baselined, 0 new).

## Cross-PR follow-up (not actioned here)

This spec is the document **#2481** was missing. The mesh plan there cited `value-engine-domino-v0-spec-2026-08-07.md` when it did not exist, and that section was rewritten to say writing it was Stage C's first task. It exists now.

**If #2516 lands, #2481's mesh plan should be updated to cite it.** Flagging rather than editing — #2481 should not be changed out from under itself while it is in review, and the citation is only correct once this merges.

PR #2511, opened 2026-08-10
chore(submodule): promote PMOVES-hermes-agent to hardened tip (42 upstream commits)

POWERFULMOVES:
## Summary

Promotes the `PMOVES-hermes-agent` gitlink from `863e3131` → `40765f769c` on `PMOVES.AI-Edition-Hardened`.

Pure upstream sync — 42 commits advance on the hardened branch. Zero PMOVES-authored commits in the delta. Upstream = `NousResearch/hermes-agent@main`, merged via fork-sync [PR #3](https://github.com/POWERFULMOVES/PMOVES-hermes-agent/pull/3) (admin-merged; the only failing check was upstream contributor attribution which does not apply to a pure upstream sync).

## Why

Unblocks:
- **Mavis join on SPARK** (was waiting on the Hermes fork to be current)
- **Evolution Fabric Wave 0 three-body review** (Crush + Hermes + Archon) per `pmoves/docs/handoffs/SPARK_EVOLUTION_FABRIC_HANDOFF_2026-08-04.md`
- All Hermes-dependent SPARK lanes (Jetson combiner fleet, Crush-in-Pinokio, HERMES_AGENT_INTEGRATION.md v0.18.2 profile)

## Verification

- `git ls-files --stage PMOVES-hermes-agent` → `160000 40765f769c...` ✓
- Target SHA confirmed on `PMOVES.AI-Edition-Hardened` branch tip
- Fork-sync PR #3 state = MERGED (rebase, 2026-08-10T03:57:32Z)
- Single gitlink-only commit; no code/config changes in parent

## Testing

- `git diff --cached` shows exactly 1 line change (subproject commit SHA)
- No parent-repo code touched; submodule-integrity gate will verify the gitlink is reachable

## Summary by CodeRabbit

* **Chores**
  * Updated the Hermes Agent component to a newer revision.
  * No user-facing functionality or behavior changes.

PR #2519, opened 2026-08-10
fix: three defects on main surfaced by the #2511 review (scoring contract, orphan gitlink, branch policy)

POWERFULMOVES:
## Summary

Three defects that already existed on `main`, surfaced by the review of #2511. None were introduced by that PR — its threads were outdated-by-rebase, but the findings underneath them were real, so they are fixed here where they actually live.

Grouped as one concern: *defects on main surfaced by the #2511 review*. The fourth finding from that review (a symlink-predictable temp path in `beats_to_voice.py`) is deliberately **not** here — it is security-shaped and gets its own PR so it receives its own review attention.

One commit per fix.

## 1. `hf-research-agent` — the scoring scale was two numbers that disagreed

`_evaluate_model` returned `"max_score": max_score + 10` (= 110) while `test_hf_services.py:187` asserted `max_score == 100`. A deterministic failure the moment the HF suite's imports resolve — **currently masked**: the class is `skipif`-guarded on import, `aiohttp` is absent, and all 21 tests skip rather than fail.

**The code was right; the test was stale.** Criteria 1-4 award 40+25+20+15 = 100 and Criterion 5 adds a genuine +10 compatibility bonus, so 110 is the attainable maximum. Verified empirically rather than by reading — a model that hits every criterion scores exactly 110:

```
BASE_MAX_SCORE = 100 | COMPAT_BONUS_POINTS = 10 | MAX_SCORE = 110
attainable maximum: score=110 == MAX_SCORE=110   <- 110 is real, not cosmetic
compat bonus: plain=80 bonused=90 delta=10
```

Nothing keys off `max_score` functionally: `MIN_SCORE` is an absolute threshold (50), not a percentage; there is no `hf.model.evaluated.v1` schema pinning the field; the only other reader is a log denominator. So rather than flip the assertion, the scale is now stated once and both the bonus and the reported denominator read those constants, so they can no longer drift apart.

New coverage for the bonus measures it as a **delta between two otherwise-identical models**, so it holds regardless of the env-derived `PREFERRED_TAGS`/`AVOID_TAGS` sets.

> Note: because of the `skipif` guard, CI green here does **not** prove these tests ran. I exercised the scoring path directly with a stubbed `aiohttp`; output above. Installing `aiohttp` into the test image so this suite actually executes is a separate, larger question.

## 2. `PMOVES-ollama` — registered but no gitlink

`.gitmodules` registered the path but `main` carried no mode-`160000` entry: 65 registered paths against 64 gitlinks, with `PMOVES-ollama` the only orphan.

Not a lost pin — a **split landing**. #2365 merged the `.gitmodules` entry; the commit that adds the gitlink (`5047cecc6`) is still unmerged on the local branch `feat/ollama-submodule` with no open PR. This restores the stranded half, pinned to a SHA verified against the remote rather than copied from that commit:

```
git ls-remote .../PMOVES-ollama.git refs/heads/PMOVES.AI-Edition-Hardened
  -> 8d8c701d6a8a7f4129488d7698e3ae02d24e223c
```

which matches the branch recorded in `.gitmodules`. Registered paths and gitlinks are now 65/65.

## 3. Branch policy — the obvious fix would have repeated the #2511 bug

`submodule-branch-policy-check` was **already failing on main**: 5 submodules with no `branch` field and 5 more pinned to a branch the allow-list did not cover. Missing `branch` is not cosmetic — `validate_submodule_gitlinks.sh` skips branch/ancestry validation for those entries and `gitlink-promoter.yml` never discovers them, so they sit outside the hardened update workflow entirely.

**Applying `SUBMODULE_BRANCH_DEFAULT` here would have been wrong.** None of the five maker forks has a `PMOVES.AI-Edition-Hardened` branch, so that would have written five branch names that do not exist — precisely the defect class that #2511 was fixed for. Each fork has exactly one branch, the upstream default, confirmed by `git ls-remote`:

| submodule | branch | gitlink on main |
|---|---|---|
| `PMOVES-obico-server` | `release` | matches tip |
| `PMOVES-moonraker-obico` | `master` | matches tip |
| `PMOVES-OrcaSlicer` | `main` | matches tip |
| `PMOVES-OctoPrint-Obico` | `master` | matches tip |
| `PMOVES-fluidd` | `develop` | matches tip |

Every gitlink already points at exactly that branch's tip, so recording these describes reality rather than changing what is checked out. Non-hardened branches go in `SUBMODULE_BRANCH_ALLOW`, the mechanism already used for `PMOVES-ClawZ=main`.

**Scope note:** the five `skills/*` forks legitimately track `main` and are added the same way. They are strictly beyond "add the missing branch fields", but without them the gate stays red and the branch fields above cannot be verified by running it. Flagging explicitly rather than burying it.

```
make -C pmoves submodule-branch-policy-check
  before: FAIL (5 missing branch, 5 policy mismatches)
  after:  PASS - checked: 65
```

## Do not fix twice

**#2493** (open) adds `PMOVES-nats-server` with no `branch` key. It will reintroduce a missing-branch failure when it lands — its fork tracks `main`. Cross-referenced so this does not get half-fixed from both ends.

## Two silent-false-negative traps hit while verifying this

Recording both; the next agent will hit them, and each returns **empty output that looks like a clean result**.

1. **`git config --file .gitmodules --get-regexp 'submodule\.NAME\.'` returns nothing for an entry that demonstrably exists.** In the #2511 worktree this returned empty for `PMOVES-hermes-agent` while the section sat at line 380. `git config --file .gitmodules --list | grep -i hermes` returns all three keys. This is a *different surface* from the known `git show origin/main:path` path-conversion trap on Windows — use `--list | grep`, or `grep` the file.
2. **`git diff` / `git show --stat` silently omit a submodule that has `ignore = all`.** Adding the `PMOVES-ollama` gitlink produced *no* entry in `git diff --name-status origin/main HEAD` and an empty `git show --stat` for its own commit, even though `git ls-tree HEAD PMOVES-ollama` showed it committed. Use `--ignore-submodules=none` or compare trees with `git ls-tree`.

## Verification

- `make -C pmoves submodule-branch-policy-check` -> **PASS, checked: 65** (FAIL on main)
- registered paths vs gitlinks -> **65/65, no orphans** (65/64 on main)
- all 5 maker branches + the ollama pin confirmed present on their remotes via `git ls-remote`
- scoring path exercised directly (stubbed `aiohttp`): max is attainable at 110, bonus delta = 10
- `python -m py_compile` clean; `.gitmodules` parses (65 sections)
- `ruff`: one pre-existing `typing.Any` unused-import warning in the test file, present identically on `main` and left alone

`submodule_integrity.py` reports uninitialized submodules in a fresh worktree — environmental, unrelated, identical before the change.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -388,15 +388,20 @@
 [submodule "PMOVES-obico-server"]
 	path = PMOVES-obico-server
 	url = https://github.com/POWERFULMOVES/PMOVES-obico-server.git
+	branch = release
 [submodule "PMOVES-moonraker-obico"]
 	path = PMOVES-moonraker-obico
 	url = https://github.com/POWERFULMOVES/PMOVES-moonraker-obico.git
+	branch = master
 [submodule "PMOVES-OrcaSlicer"]
 	path = PMOVES-OrcaSlicer
 	url = https://github.com/POWERFULMOVES/PMOVES-OrcaSlicer.git
+	branch = main
 [submodule "PMOVES-OctoPrint-Obico"]
 	path = PMOVES-OctoPrint-Obico
 	url = https://github.com/POWERFULMOVES/PMOVES-OctoPrint-Obico.git
+	branch = master
 [submodule "PMOVES-fluidd"]
 	path = PMOVES-fluidd
 	url = https://github.com/POWERFULMOVES/PMOVES-fluidd.git
+	branch = develop
--- a/pmoves/mk/preflight.mk
+++ b/pmoves/mk/preflight.mk
@@ -9,7 +9,20 @@ RETRO_FLAGS ?=
 RUNNER_PHASE ?= local-certification
 SUBMODULE_LAYER_MANIFEST ?= configs/submodule_layer_validation_manifest.json
 SUBMODULE_BRANCH_DEFAULT ?= PMOVES.AI-Edition-Hardened
-SUBMODULE_BRANCH_ALLOW ?= PMOVES-DoX=PMOVES.AI-Edition-Hardened-DoX,PMOVES-ClawZ=main
+# Submodules that legitimately track an upstream default branch instead of a
+# hardened branch. Every branch here is verified to exist on its remote.
+SUBMODULE_BRANCH_ALLOW ?= PMOVES-DoX=PMOVES.AI-Edition-Hardened-DoX,\
+  PMOVES-ClawZ=main,\
+  PMOVES-obico-server=release,\
+  PMOVES-moonraker-obico=master,\
+  PMOVES-OrcaSlicer=main,\
+  PMOVES-OctoPrint-Obico=master,\
+  PMOVES-fluidd=develop,\
+  skills/Pmoves-skills=main,\
+  skills/PMOVES-awesome-agent-skills=main,\
+  skills/pmoves-fork-repository-skill=main,\
+  skills/PMOVES-agent-sandbox-skill=main,\
+  skills/Pmoves-claude-d3js-skill=main
 AUDIT_RUNTIME_GPU ?= 0
 PRECHECK_VENV_WIN ?= .venv-pmoves/Scripts/python.exe
 PRECHECK_VENV_UNIX ?= .venv-pmoves/bin/python


POWERFULMOVES:
Verification pass — checked the one judgement call rather than accepting it.

**The scoring contract resolves correctly, and for the right reason.** The question was which side is authoritative, not just making code and test agree. Derived from the rubric in `main.py:56-62`:

```
BASE_MAX_SCORE = 100          # Criteria 1-4: 40 + 25 + 20 + 15
COMPAT_BONUS_POINTS           # Criterion 5, on top
MAX_SCORE = BASE + BONUS      # = 110
```

So the **code was right and the test was stale** — `assert == 100` predated Criterion 5. Naming `MAX_SCORE` as a constant, asserting `== hf_research_mod.MAX_SCORE == 110`, and adding `test_max_score_matches_the_documented_rubric` means the next drift fails loudly at the contract rather than silently at a magic number. That is the right disposition.

Worth noting this was only ever reachable because a `skipif` import guard was masking it — the suite would have failed deterministically the moment the HF imports resolved. A test that cannot run is not a test that passes.

The other two are bookkeeping and verify clean: `PMOVES-ollama` was the sole orphan (65 `.gitmodules` entries vs 64 gitlinks on main — I counted independently), and the branch-policy gap is real.

**Cross-reference for whoever lands this:** the missing-`branch`-field problem is not only the five maker submodules. **#2493** adds `PMOVES-nats-server` with no `branch` key either, so the same gap is actively being widened on another open PR. Worth a line in `.gitmodules` policy rather than two independent fixes that each miss the other.

agent_signature (advisory, unsigned-local): `ACK::CLAUDE-OPUS-5::PR2519-VERIFICATION-2026-08-10`

POWERFULMOVES:
## Same caveat as #2520: `python-tests` green is not evidence

The `python-tests` check passed on this PR while running **zero** tests (`collected 455 items / 4 errors / 2 skipped` -> `2 skipped, 4 errors`). Cause is `.github/workflows/merge-gate.yml:41`, which caps discovery at `head -20` files and ends in `|| true`, so it cannot fail. Full evidence: https://github.com/POWERFULMOVES/PMOVES.AI/pull/2520#issuecomment-5243502588

This matters specifically for finding 1 here: the hf-research-agent suite is *also* `skipif`-guarded on a missing `aiohttp`, so even when selected it skips rather than runs. That is exactly how a `max_score == 100` assertion sat contradicting code returning 110 without anyone noticing.

The evidence for this PR is therefore the direct runs recorded in the description — `submodule-branch-policy-check` going FAIL -> PASS, 65/65 gitlink parity, and the scoring path exercised with a stubbed `aiohttp` — not the green check.

PR #2538, opened 2026-08-14
fix(launcher): claude-pmoves via PATH symlink resolved ROOT=$HOME — MCP creds silently empty

POWERFULMOVES:
## What broke

`install-claude-pmoves-command.sh` installs the command by **two** routes:

- an rc function holding the launcher's absolute path, and
- a `~/.local/bin/claude-pmoves` symlink, for shells that don't source the rc.

They behaved differently, because of one line:

```bash
ROOT="$(cd "$(dirname "${BASH_SOURCE[0]:-$0}")/../.." && pwd)"
```

| Route | `BASH_SOURCE[0]` | derived `ROOT` |
|---|---|---|
| rc function | the real launcher path | the repo ✅ |
| PATH symlink | `~/.local/bin/claude-pmoves` | `$HOME` ❌ |

With `ROOT=$HOME`, the shared env file and `.claude/mcp.json` both miss.

**It failed as a `WARN`, not an error.** The launcher printed the warning and
`exec`'d claude anyway — producing a session that looks completely healthy with
every cred-dependent MCP dark. Same failure class as #2484, different entrypoint.

Surfaced on B850 (Knuckles), where the rc block had never been installed, so the
symlink was the only live route.

## What changed

**1. Walk symlinks before deriving ROOT.** Pure-bash loop handling relative link
targets — no `readlink -f` dependency, so it stays portable.

**2. Validate ROOT against a marker file rather than trusting the computation.**

This is the part worth reviewing. The sibling launchers *already* do this —
`crush-pmoves` tests `$candidate/pmoves/Makefile`, `pmoves-mini` tests
`$candidate/pmoves/tools/mini_cli.py`, and both reject a candidate that fails.
Both compute the same wrong candidate from `dirname "$0"`; **validation, not path
arithmetic, is why they survived the same trap.** `claude-pmoves.sh` was the only
one that computed a root and trusted it.

A launcher whose entire job is loading repo-relative config must not proceed when
it cannot find the repo, so it now hard-exits with a fix hint. `PMOVES_REPO_ROOT`
is the escape hatch for deliberate overrides.

## Verification (B850 / Knuckles)

| Route | Result |
|---|---|
| PATH symlink | `loaded 373 vars` — was **0**, with two WARNs |
| real path (rc function) | `loaded 373 vars` |
| `PMOVES_REPO_ROOT` override | honored |
| bogus root (`PMOVES_REPO_ROOT=/tmp`) | **exit 1** + actionable error |

`bash -n` clean. (`shellcheck` is not installed on this node, so that lint did not run.)

## Audit note

The surrounding launcher design was checked against the Claude Code docs in the
same pass, expecting to find more. It's correct as written:

- `--mcp-config` is variadic (`<configs...>` per `claude --help`), so the `=` form
  is required to stop it swallowing a trailing prompt — the #2243 P1 fix was right.
- Omitting `--strict-mcp-config` is what keeps the per-node `.mcp.json` merged
  (`--strict-mcp-config` = *"Only use MCP servers from --mcp-config, ignoring all
  other MCP configurations"*).
- `${VAR}` in the roster does resolve from Claude Code's **own** process env, which
  is why sourcing the shared env before `exec` is the correct mechanism.
- `.claude/mcp.json` is not one of the scopes Claude reads on its own (those are
  `~/.claude.json` and a project-root `.mcp.json`), so the explicit `--mcp-config`
  really is required.

ROOT derivation was the only defect.

## Blast radius

Any node that installed `claude-pmoves` via the symlink route has been running with
an empty MCP roster. Worth a spot-check across the fleet after merge:
`readlink ~/.local/bin/claude-pmoves && claude-pmoves --version` should report
`loaded <N> vars`, not a WARN.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved launcher reliability when run through symbolic links.
  * Supports explicitly configured repository locations.
  * Added validation and clear diagnostics when the repository cannot be located.

### DIFF
--- a/deploy/provision/claude-pmoves.sh
+++ b/deploy/provision/claude-pmoves.sh
@@ -14,35 +14,59 @@
 # Usage: run this instead of `claude` (alias it: alias claude=/path/to/claude-pmoves.sh).
 set -u
 
-# Resolve through symlinks before deriving ROOT. install-claude-pmoves-command.sh
-# drops a ~/.local/bin/claude-pmoves symlink on PATH as the route for shells that
-# don't source the rc function. Taking dirname of the SYMLINK made ROOT=$HOME, so
-# env.shared and .claude/mcp.json both missed and every cred-dependent MCP started
-# empty — silently, with only a WARN. Same failure class as #2484, other entrypoint.
+# ---------------------------------------------------------------------------
+# REPO-ROOT RESOLUTION — keep byte-identical across the three launchers that
+# carry it (this file, crush-pmoves.sh, pmoves/scripts/claude-pmoves.sh).
+# Enforced by deploy/provision/tests/test-launcher-root-resolution.sh, which
+# fails if one is fixed and the others are not — the drift that produced this
+# bug in the first place.
+#
+# WHY THE WALK: install-claude-pmoves-command.sh drops a ~/.local/bin/<name>
+# symlink on PATH for shells that don't source the rc function. Taking dirname
+# of the SYMLINK made ROOT=$HOME, so env.shared and .claude/mcp.json both missed
+# and every cred-dependent MCP started empty — silently, with only a WARN.
+#
+# WHY `CDPATH= cd -P --`: dirname yields a bare relative path when the script is
+# invoked relatively; `cd` consults CDPATH for such arguments, which both jumps
+# elsewhere AND echoes the destination, embedding a newline in the captured path.
+# ---------------------------------------------------------------------------
 SELF="${BASH_SOURCE[0]:-$0}"
 while [ -L "$SELF" ]; do
-  link_dir="$(cd -P "$(dirname "$SELF")" && pwd)"
-  SELF="$(readlink "$SELF")"
+  link_dir="$(CDPATH= cd -P -- "$(dirname -- "$SELF")" && pwd)"
+  SELF="$(readlink -- "$SELF")"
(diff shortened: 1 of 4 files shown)


POWERFULMOVES:
Reworked after review — the first cut fixed **one of three** launchers carrying the same code.

- `deploy/provision/crush-pmoves.sh` and `pmoves/scripts/claude-pmoves.sh` had the identical unvalidated root derivation. All three now carry a byte-identical walk.
- `PMOVES_REPO_ROOT` collided with an existing consumer (`pmoves/services/creator-operator/config.py:15`) → renamed `PMOVES_LAUNCHER_ROOT`.
- Hard exit removed the documented `alias claude=...` degraded path and printed an empty root → an explicit `PMOVES_ENV_SHARED` is now honoured with a warning; unresolved roots print `<unresolved>`.
- `CDPATH` could corrupt the resolved path → `CDPATH= cd -P --` throughout.

**New guard:** `deploy/provision/tests/test-launcher-root-resolution.sh` — asserts the walk is identical across all three, that none reintroduces the colliding variable, and that each resolves correctly through both an absolute symlink and a relative symlink chain. 12/12 pass. Negative control: reverting one launcher to the old walk fails with *"walk DIFFERS ... fix all three together"*, so the guard is proven to catch what it exists for.

Live on B850: 373 vars via the installed symlink, still 373 under hostile `CDPATH`, still 373 with `PMOVES_REPO_ROOT=/tmp` exported.

PR #2544, opened 2026-08-14
fix(ci): branch-trail-emit failed half its runs on stale runner workspaces

POWERFULMOVES:
## Not a repo bug

`main` pins `pmoves/integrations/archon` at **`944d31da`**, which carries **zero gitlinks and no `.gitmodules`**. The repo is clean.

The failing jobs log `submodules: false`, yet `git submodule status` reports **`1e02907a`** — a pre-fix commit that *does* carry `external/*` gitlinks with no `.gitmodules`. `actions/checkout` still runs `git submodule foreach --recursive` for its sshCommand cleanup **regardless of that setting**, and it aborts:

```
fatal: No url found for submodule path
  'pmoves/integrations/archon/external/PMOVES-Agent-Zero' in .gitmodules
```

`runs-on: [self-hosted, Linux]` matches **five** registered runners (2× b850, 2× kvm4, 1× spark). Only some are parked on the stale commit, so the job failed **4 of its last 8 runs, alternating** — which reads as flaky CI rather than one stale directory. #2519 fixed the repo; nothing reset the runners.

## Scope — checked, not assumed

A naive grep for `self-hosted` suggested 15 exposed workflows. That was wrong:

| workflow | runs-on | exposed |
|---|---|---|
| `hardening-validation` *(required)* | `ubuntu-latest` | no — ephemeral |
| `submodule-gitlink-gate` *(required)* | `ubuntu-latest` | no — ephemeral |
| `village-gate` | `ubuntu-latest` | no — ephemeral |
| `branch-trail-emit` | `[self-hosted, Linux]` | **yes** |

Run history confirms it: `hardening-validation` 7/8 success, `submodule-gitlink-gate` 8/8, `branch-trail-emit` 4 failures in 8.

**No required check is affected**, so this is noise rather than a merge blocker. Noise that alternates is still expensive — it teaches people that red means nothing, which is how the two silently-broken bump workflows in #2542 went unnoticed.

## Fix

Deinit before checkout so stale state cannot survive a run. This job needs no submodules, so dropping them costs nothing, and it is best-effort because a clean or first-run workspace has nothing to deinit.

## This is a mitigation, not the cure

The five runners still hold a stale archon checkout, and **any future job that recurses will hit it**. Cleaning those workspaces is an operator action and is deliberately not attempted from here.

### DIFF
--- a/.github/workflows/branch-trail-emit.yml
+++ b/.github/workflows/branch-trail-emit.yml
@@ -45,6 +45,30 @@ jobs:
     runs-on: [self-hosted, Linux]
     timeout-minutes: 5
     steps:
+      # These runners keep a PERSISTENT workspace. `submodules: false` means
+      # checkout never updates the submodules, so whatever they were left at
+      # survives forever -- and actions/checkout still runs
+      # `git submodule foreach --recursive` for its sshCommand cleanup,
+      # regardless of that setting. Some runners are parked on an archon commit
+      # (1e02907a) that carries external/* gitlinks with no .gitmodules, so the
+      # foreach aborts:
+      #
+      #   fatal: No url found for submodule path
+      #     'pmoves/integrations/archon/external/PMOVES-Agent-Zero' in .gitmodules
+      #
+      # The REPO is clean -- main pins archon at 944d31da, which has zero
+      # gitlinks. #2519 fixed the repo; nothing reset the runners. Because
+      # `runs-on: [self-hosted, Linux]` matches five runners and only some are
+      # dirty, this failed 4 of the last 8 runs, alternating, which reads as
+      # flaky CI rather than one stale directory.
+      #
+      # Deinit before checkout so the state cannot persist. This job needs no
+      # submodules at all, so dropping them is free. Best-effort by design: on a
+      # clean or first-run workspace there is nothing to deinit.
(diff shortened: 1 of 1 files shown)


POWERFULMOVES:
## Verified causally, not by luck

The alternating 4-of-8 failure pattern makes a single green run meaningless — it could just be landing on a clean runner. So here is the controlled comparison across four PRs open right now:

| PR | branch-trail-emit.yml version | `emit lifecycle trail` |
|---|---|---|
| #2541 | main's | **fail** |
| #2542 | main's | **fail** |
| #2546 | main's | **fail** |
| **#2544** | **this fix** | **pass ×2** |

Two passes on this branch, not one — landing on a clean runner twice in a row while three concurrent PRs all fail is not a plausible coincidence.

Step-level evidence from run `31817856218`:

```
Set up job: success
Clear stale submodule state (persistent self-hosted workspace): success
Checkout: success        <- this is where the other three abort
Ensure uv: success
```

The new step runs, then the checkout that was aborting on `git submodule foreach --recursive` completes.

**Still a mitigation, not the cure.** The five runners continue to hold the stale archon checkout at `1e02907a`; this only stops it surviving *within* this one workflow. Any other job that recurses will still hit it. Cleaning those workspaces remains an operator action.

PR #2545, opened 2026-08-14
fix(b850): rocm-smi exporter served empty scrapes; node profile diverged from hardware

POWERFULMOVES:
Two symptoms on B850/Knuckles that turned out to be one investigation: the Grafana GPU
panel was blank, and the node profile described hardware that isn't installed.

## 1. The exporter served HTTP 200 with a zero-byte body

`rocm-smi-http@.service` inlined shell directly in `ExecStart`:

```
printf "...Content-Length: %d..." "${#body}"
```

systemd expands **both** `%` specifiers and `${VAR}` in `ExecStart` before `/bin/sh` ever
sees the string. What systemd actually resolved:

```
Content-Length: /run/credentials/rocm-smi-http@1.service      <- %d = credentials-dir specifier
journal: "Invalid environment variable name evaluates to an empty string: #body"
```

`%d` became a path, `%s` became the user's shell, `${#body}` became `""`. The printf format
was corrupted and the length argument vanished.

**Every scrape returned 200 with nothing in it** — while the socket unit reported
`active (listening)`, the exporter reported `active (running)`, and
`/run/rocm-smi-metrics.prom` held perfectly good samples. Nothing in the stack looked wrong.

**Fix:** move the responder into a real script systemd never parses. Escaping (`%%` / `$$`)
would also work, but reintroduces the bug on the next edit.

**Caught while fixing:** the responder used `body=$(cat file)`, and command substitution
strips trailing newlines — `Content-Length` came out one byte short (458 vs 459) and the
exposition text lost the final newline the Prometheus format requires. Now takes the length
via `wc -c` and streams the file, so wire bytes == disk bytes (verified with `cmp`).

## 2. The memory metric never emitted a sample

The collector read `rocm-smi --showmemuse` and selected `"VRAM Total Used Memory (B)"`.
That flag doesn't return that key. Verified live on ROCm 7.1.0:

| Flag | Returns |
|---|---|
| `--showmemuse` | `GPU Memory Allocated (VRAM%)`, `Memory Activity`, … |
| `--showmeminfo vram` | `VRAM Total Memory (B)`, **`VRAM Total Used Memory (B)`** |

`select()` never matched, so `HELP`/`TYPE` printed with zero samples — invisible unless you
counted them. Now reads `--showmeminfo vram`, and adds `rocm_gpu_memory_total_bytes` so used
can render as a percentage.

### Verification on this node

| | Before | After |
|---|---|---|
| Body bytes | **0** | 759 |
| `Content-Length` vs file | corrupted | 759 = 759 |
| Body vs file | — | `cmp`: IDENTICAL |
| `rocm_` samples | 4 | **8** |
| Missing-file path | — | 200, `Content-Length: 0`, exit 0 |
| Responder journal | `Invalid environment variable name…` every scrape | clean |

`install-rocm-exporter-fix.sh` backs up both files, verifies, and prints rollback commands
on failure. Already applied to B850.

## 3. Profile diverged from measured hardware

`ram_gb: 128` → **32** (`MemTotal 31939336 kB`). A second GPU doesn't add system RAM, so
that one was simply wrong.

**The dual-R9700 spec is deliberately NOT changed.** The operator confirms two cards are
physically installed; exactly one enumerates. That's a hardware/firmware fault, not a spec
error — so `total_vram_gb: 64` and `dual_gpu_flags` stay as the correct target.

Added a `probe:` block recording measured state, **8 divergences and 3 open questions**:

- Above-4G / ReBAR are already **ON** (card 1 holds a 32 G BAR at `0xf000000000`), so the
  usual dual-large-VRAM culprit is ruled out.
- CPU root ports are fully committed (`00:01.1` = card 1, `00:01.2` = NVMe); the chipset has
  **5 empty downstream ports** and card 2 is on none of them.
- `compose_overrides` names `docker-compose.rocm.override.yml` — **no such file** in the repo.
- `pmoves/docs/P7_SKILL_B850.md` starts llama-server on `:8080`, contradicting this profile's
  `:8090` and its stated reason (8080 reserved for Agent Zero fleet-wide).
- `AGNOTE4482_SITREP.md:243` calls this node *"64GB RAM; no discrete GPU"* — wrong on both
  counts. **A gfx1201 R9700 with 32 GB VRAM is present and idle**, so this node has probably
  been getting no GPU work at all.

The repo's generator (`json-to-profile.py`) emits a flat `hardware:` block with no
declared-vs-observed concept, which is why this drift had nowhere to be recorded.

## Deliberately not included

- **amd-smi migration.** `rocm-smi` is deprecated upstream in favour of `amd-smi` (already
  installed here: AMDSMI 26.1.0 / ROCm 7.1.0), and AMD's official Prometheus path is now
  [ROCm/device-metrics-exporter](https://github.com/ROCm/device-metrics-exporter) — the older
  [amd_smi_exporter](https://github.com/amd/amd_smi_exporter) is EOL. Migrating renames every
  metric, so bundling it with a repair would make a blank panel impossible to attribute.

## Known: this exporter still isn't scraped

Fixing the endpoint is necessary but not sufficient. Tracked separately:

1. **Prometheus has no job for `:9835`** — its config covers `cadvisor`, `loki`,
   `flute-gateway`, `tensorzero-gateway`, `deepresearch`, `supaserch`, `neo4j`,
   `gpu-orchestrator`, `otel-collector`. Never wired.
2. **UFW default-deny blocks all container→host traffic** on this node (~1,341 drops/6 h; no
   rule for any Docker bridge subnet). Prometheus can reach neither
   the exporter neither via `host.docker.internal` nor via its bridge gateway. Running the exporter as a container
   would sidestep this entirely — which is also the natural shape of the
   device-metrics-exporter migration.

---

Release Note: B850/Knuckles node profile (`workstation-9850x3d-dual-r9700.yaml`, `node_id: pmoves-b850`) corrected — `ram_gb` 128 → 32 (measured). The dual-R9700 target spec (`total_vram_gb: 64`, `dual_gpu_flags`) is intentionally unchanged: both cards are physically installed but only one enumerates, which is a hardware/firmware fault rather than a spec error. A new `probe:` block records measured state with 8 divergences and 3 open questions, including that this node has a live gfx1201 R9700 with 32 GB VRAM despite `AGNOTE4482_SITREP.md:243` describing it as having no discrete GPU. Schedulers or routing that read this profile should expect single-GPU capacity on B850 until the second card enumerates.

### DIFF
--- a/deploy/provision/install-rocm-exporter-fix.sh
+++ b/deploy/provision/install-rocm-exporter-fix.sh
@@ -75,9 +75,14 @@ echo "==> VERIFY"
 # fallback would concatenate and report a nonsense "000000". Default after capture.
 code="$(curl -s -o /dev/null -w '%{http_code}' --max-time 5 http://127.0.0.1:9835/metrics 2>/dev/null)" || true
 code="${code:-000}"
-body="$(curl -s --max-time 5 http://127.0.0.1:9835/metrics 2>/dev/null || true)"
-bytes="$(printf '%s' "$body" | wc -c | tr -d ' ')"
-samples="$(printf '%s' "$body" | grep -c '^rocm_' || true)"
+# Measure bytes with curl's own counter, NOT `$(curl ...) | wc -c`: command
+# substitution strips trailing newlines, so the report was one byte short of what
+# the endpoint actually served (755 vs 756). A verify step that misreports the
+# number it is verifying is the same class of defect this script exists to fix.
+_tmp="$(mktemp)"; trap 'rm -f "$_tmp"' EXIT
+bytes="$(curl -s --max-time 5 -o "$_tmp" -w '%{size_download}' http://127.0.0.1:9835/metrics 2>/dev/null)" || true
+bytes="${bytes:-0}"
+samples="$(grep -c '^rocm_' "$_tmp" 2>/dev/null || true)"
 cards="$(rocm-smi --showid --json 2>/dev/null | grep -o '"card[0-9]*"' | sort -u | wc -l | tr -d ' ' || echo '?')"
 
 echo "    HTTP          : $code    (want 200)"


POWERFULMOVES:
Reworked after review — the first cut fixed the installed files and left **the provisioner that generates them** broken.

`deploy/provision/rdna4-gpu-install.sh` (invoked by `hostinger-kvm-setup.sh:489`) still heredoc-ed the broken `ExecStart` with `%d`/`${#body}` and the `--showmemuse` collector. Every fresh RDNA4 node reproduced both bugs, a routine ROCm reinstall would silently revert B850, and this PR was becoming a second divergent copy. The exporter now has **one definition** — the provisioner installs the repo files and fails loudly if absent.

Also fixed in my own new code:
- responder: missing file returned **200 + 0 bytes** (the same invisibility this PR kills) → 503; dead collector served frozen metrics forever → 503 past 60s; `wc -c` + `cat` could straddle the collector`s `mv` → read once; request now drained (unread data → RST can discard the response).
- installer: `set -euo pipefail` + unguarded `curl | wc -c` aborted **before printing its own rollback commands**; success gate hardcoded `samples -ge 8` (B850`s discrete+iGPU count — a single-GPU node would be told "STILL BROKEN" and roll back a working fix); `cp -a` assumed targets existed; never enabled the units it shipped.

Verified: 200/756==756 + trailing newline · missing→503 · stale→503 (`600s old`) · fresh→200 · unreachable endpoint reaches the rollback branch.

Profile `probe:` updated — the two exporter divergences now read FIXED, and two new entries name the rest of the chain: Prometheus has no job for `:9835`, and UFW default-deny drops all container→host traffic, so a host-bound exporter is unscrapeable here regardless of correctness.

PR #2547, opened 2026-08-14
fix(config): DGX Spark GB10 is sm_121, not sm_110 — read from the hardware

POWERFULMOVES:
## The repo held two contradicting answers for the same GPU

| file | value | |
|---|---|---|
| `config/profiles/dgx-spark-grace-blackwell.yaml:20` | `compute_capability: "12.1"` | ✅ already correct |
| `config/gpu-models.yaml:401` | `compute_capability: "11.0"  # SM_110` | ❌ |
| `config/rooms/creator-studio.room.collab.json:29` | `gpu_arch: [sm_120, sm_110]` | ❌ |

Someone already corrected the profile. The other two were never updated — so a reader landing on `gpu-models.yaml` still got `sm_110`.

**I was that reader.** I wrote `CUDA_ARCHS="87;110"` into the arm64 build override on #2446. That build arg **cannot compile**: `compute_110` does not exist in CUDA 12.x. `sm_110` is Jetson **THOR**, a different part.

## Ground truth — from NVML on the node

Not from the model name, not from reasoning about Blackwell generations. Spark's Ollama logs what NVML reports, and its Loki is readable over the tailnet:

```
msg="inference compute" library=CUDA compute=12.1 name=CUDA0
  description="NVIDIA GB10" driver=13.0 type=iGPU total="121.6 GiB"
```

Consistent across every restart logged **2026-08-07 → 2026-08-13**. Queried independently twice.

## The same log shows the bug already happening in production

```
msg="skipping CUDA device — compute capability not in compiled architectures"
  device="NVIDIA GB10" cc=1210 archs="[500 520 600 610 700 750 800 860 890 900 1200]"
```

Ollama's CUDA-12 build has `sm_120` but **not** `sm_121`, so it **skipped the GPU** and fell back to its bundled `cuda_v13` libdir. sm_121 breaking a CUDA 12.x binary on this exact hardware is **observed, not theorised**.

It is also the same silent-degradation class as the rest of this lane: it runs, it just doesn't use the 121 GiB accelerator, and nothing says so.

**For anyone compiling for this node:** sm_121 needs **CUDA ≥ 12.9**. The node ships driver 13.0 / CUDA 13.0.2 — target 13.x, not 12.8.

## What was NOT obtained — stated, not papered over

- **Literal `nvidia-smi` output.** The tailnet ACL denies this node as an SSH *source* — `tailnet policy does not permit you to SSH as user "..."` — for every login and every destination tried. NVML is the same library `nvidia-smi` reads, so the number is a hardware readout, but **the ACL gap is real and worth closing** if you want direct fleet introspection.
- **Jetson Orin's `8.7` is still unverified.** The nano nodes are online but expose no readable surface and hit the same ACL denial. It is consistent with Orin's known part — but nobody has read it off that hardware. That is exactly the standing `sm_110` had until today, so it is flagged rather than assumed.

## Follow-up

#2446's `CUDA_ARCHS="87;110"` still needs correcting, and one arch list cannot serve both arm64 nodes — Jetson Orin wants `87` on CUDA 12.8, Spark wants `121` on CUDA 13.x. Handled separately on that PR.

### DIFF
--- a/pmoves/configs/pinokio-apps/curated/ace-step.yaml
+++ b/pmoves/configs/pinokio-apps/curated/ace-step.yaml
@@ -9,7 +9,7 @@ runtime:
   autostart: true
   gpu_required: true
   min_vram_mb: 8192
-  gpu_arch: [sm_120, sm_110]
+  gpu_arch: [sm_120, sm_121]
   gpu_reservation_mb: 8192
   gpu_reservation_mode: concurrent
   dependencies: []
--- a/pmoves/configs/pinokio-apps/curated/comfyui-desktop.yaml
+++ b/pmoves/configs/pinokio-apps/curated/comfyui-desktop.yaml
@@ -9,7 +9,7 @@ runtime:
   autostart: true
   gpu_required: true
   min_vram_mb: 16384
-  gpu_arch: [sm_120, sm_110]
+  gpu_arch: [sm_120, sm_121]
   gpu_reservation_mb: 16384
   gpu_reservation_mode: concurrent
   dependencies: []
--- a/pmoves/configs/pinokio-apps/curated/lightonocr-2-1b.yaml
+++ b/pmoves/configs/pinokio-apps/curated/lightonocr-2-1b.yaml
@@ -9,7 +9,7 @@ runtime:
   autostart: true
   gpu_required: true
   min_vram_mb: 2048
-  gpu_arch: [sm_120, sm_110, sm_89]
+  gpu_arch: [sm_120, sm_121, sm_89]
   gpu_reservation_mb: 2048
   gpu_reservation_mode: concurrent
   dependencies: []
--- a/pmoves/configs/pinokio-apps/curated/qwen3-tts.yaml
+++ b/pmoves/configs/pinokio-apps/curated/qwen3-tts.yaml
@@ -9,7 +9,7 @@ runtime:
   autostart: false
   gpu_required: true
   min_vram_mb: 4096
-  gpu_arch: [sm_120, sm_110]
+  gpu_arch: [sm_120, sm_121]
   gpu_reservation_mb: 4096
   gpu_reservation_mode: concurrent
   dependencies: []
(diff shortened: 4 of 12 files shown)


POWERFULMOVES:
## Scope correction — it was 14 files, not 2

The PR body above says two files. That was wrong when I wrote it, and the correction is worth recording because of *how* it surfaced.

I searched `pmoves/config/`. There is also **`pmoves/configs/`** — a different directory, plural — and the bad arch had spread through it. I only found out because a repo-wide grep I had started earlier, hit a 2-minute timeout on, and mentally written off as too slow, **finished in the background and disagreed with me**.

Twelve more:

| count | path | form |
|---|---|---|
| 10 | `pmoves/configs/pinokio-apps/curated/*.yaml` | `gpu_arch: [sm_120, sm_110]` |
| 1 | `pmoves/configs/tac_trees/pinokio-p8.tac.yaml` | probe `?gpu_arch=sm_120,sm_110` |
| 1 | `pmoves/configs/pinokio-apps/schema/pinokio-app.v1.schema.json` | **enum** |

### The schema is the one that mattered

Its `gpu_arch` enum **legalised `sm_110` and contained no `sm_121` at all**. So any manifest naming the *real* architecture would have **failed validation**, while the fictional one passed.

Fixing the manifests without the schema would have converted a wrong-value bug into a broken-gate bug — the gate would have started rejecting the truth.

`sm_121` added. **`sm_110` deliberately retained**: Jetson THOR is a real NVIDIA architecture, and the enum is a vocabulary, not a fleet inventory. Nothing in PMOVES is a Thor node today, but that is a fact about the fleet, not about which arch strings exist.

### Verified after the change

- every `gpu_arch` value across all curated manifests validates against the schema enum
- no `sm_110` survives outside the schema vocabulary
- schema still parses as JSON; all manifests still parse as YAML

### Running total for one unverified number

2 config files → 12 configs files → a Docker build arg on #2446 → a JSON-schema enum that would have rejected the correct value. Written once, inherited by every later reader, including me.

PR #2559, opened 2026-08-15
fix(damage-control): readOnlyPaths were writable from Bash via any interpreter

POWERFULMOVES:
## The gap

Every existing pattern keys on a **shell verb** — `>`, `tee`, `sed -i`, `cp`, `mv`, `truncate`. An interpreter uses none of them:

```bash
python - <<'PY'
import pathlib
pathlib.Path('.claude/context/mcp-api.md').write_text('x')
PY
```

matched **nothing**. Every `readOnlyPath` was writable from Bash while the `Edit`/`Write` tools were correctly blocked — so the guard looked effective from the tool side and was open from the shell side.

## How it surfaced, and the audit

An agent applying doc fixes wrote to four `.claude/context/` files with `perl -i` before hitting the Write guard, noticed, and reverted them.

I then audited **my own** session, since I'd used the python-heredoc form throughout: all six branches, **64 files**, against the **56** `readOnlyPaths` entries — **zero writes to a protected path**. Nothing improper landed. But the capability was there and neither of us would have been told.

## The fix

The hook receives the whole command string, heredoc body included, so the path *is* present — just on a later line than the interpreter. Hence `[\s\S]*` where every other pattern is single-line. Order-independent lookaheads assert *"an interpreter runs"* **and** *"a write API is used"*, then require the protected path.

**Reads stay allowed, deliberately.** read-only means read-only, not no-access. The obvious over-broad fix — *interpreter + path* — would block `python -c "print(open(P).read())"`, and a guard that blocks legitimate reads gets disabled by whoever it inconveniences.

| blocked (6) | allowed (5) |
|---|---|
| `write_text`, `write_bytes`, `open(...,'w')` | `python -c "print(open(P).read())"` |
| `shutil.copy`, `os.remove` | `cat`, `grep` |
| `node writeFileSync` | interpreter write to an **unprotected** path |
| `python3.12` variants | path named with no interpreter |

## Residual gap — stated, not papered over

This **cannot** resolve indirection:

```bash
for f in <paths>; do perl -i -pe '...' "$f"; done
```

The path hides behind `$f` and no regex over command text can expand it. **That is the hole the agent actually fell through.** Closing it needs interpretation, not pattern matching. The defence there is that `Edit`/`Write` are blocked, so it only matters for a deliberate bypass — a policy question, not a technical one.

## The test also guards a failure mode I hit myself

`check_path_patterns()` catches `re.error`, warns, and continues — so a **malformed pattern doesn't crash, it silently protects nothing**, while the hook still exits 0 and the write still lands.

My first version of these patterns was over-escaped and did exactly that. The unit test caught it; a live run would not have. Same shape as everything else in this lane: the check ran, reported success, and measured nothing.

So the test asserts **every read-only pattern compiles**, not just that these behave.

## Verification

- 20 read-only patterns compile
- 11 behaviour checks pass
- all **four** existing damage-control tests still pass — no regression

### DIFF
--- a/.claude/hooks/damage-control/bash-tool-damage-control.py
+++ b/.claude/hooks/damage-control/bash-tool-damage-control.py
@@ -100,39 +100,82 @@ TRUNCATE_PATTERNS = [
 # ---------------------------------------------------------------------------
 # INTERPRETER WRITE PATTERNS
 # ---------------------------------------------------------------------------
-# Every pattern above keys on a SHELL verb: >, tee, sed -i, cp, mv, truncate.
-# An interpreter uses none of them. `python - <<'PY'` followed by
-# pathlib.write_text() matched NOTHING, so readOnlyPaths were writable from
-# Bash the whole time, while the Edit/Write tools were correctly blocked.
+# Every other pattern keys on a SHELL verb (>, tee, sed -i, cp, mv, truncate).
+# An interpreter uses none of them, so `python - <<'PY'` + pathlib.write_text()
+# matched nothing and readOnlyPaths were writable from Bash while Edit/Write
+# were correctly blocked.
 #
-# The hook receives the entire command string, heredoc body included, so the
-# path IS present here -- it is simply on a later line than the interpreter.
-# Hence [\s\S]* rather than .*, since every pattern above is single-line.
+# THE PATH MUST BE QUOTED AND BOUND TO THE WRITE. An earlier revision asserted
+# 'an interpreter runs' AND 'a write verb appears' independently, then only
+# required the path to occur somewhere in the command. That was wrong three ways:
 #
-# Order-independent by construction: the write verb may precede the path
-# (open(p, 'w')) or follow it (p = Path(x); p.write_text(...)). The lookaheads
-# assert 'an interpreter runs' AND 'a write API is used' anywhere in the
-# command, then require the protected path to appear.
+#   * readOnlyPaths includes /usr/, /bin/, build/, .venv/, node_modules/, so
+#     `/usr/bin/python3 -c "...write_text('a.txt')..."` was BLOCKED on /usr/,
+#     and `source .venv/bin/activate && python -c "open('out.txt','w')"` on /bin/.
+#   * reading a protected file and writing elsewhere was blocked, contradicting
+#     the read-only contract -- the canonical regenerate-a-doc script.
+#   * unanchored [\s\S]* re-scanned from every offset, x56 paths x2 forms, in a
+#     BLOCKING PreToolUse hook: a 5 KB heredoc stalled every Bash call 22s.
 #
-# READS STAY ALLOWED, deliberately: read-only means read-only, not no-access.
-# A bare python -c "print(open(P).read())" has no write verb and does not match.
+# Requiring the path inside a quoted string fixes the first: /usr/ and .venv/
+# appear as bare command fragments, a real target is always a literal. Binding
+# it to the write operation fixes the second: `Path('P').read_text()` has a READ
+# verb after the path. Dropping the lookaheads and the leading wildcard fixes
+# the third -- re.search already scans, so these stay linear.
 #
-# RESIDUAL GAP, stated rather than papered over: this cannot resolve indirection.
-#   for f in <paths>; do perl -i -pe '...' "$f"; done
-# hides the path behind $f, and no regex over the command text can expand it.
-# Closing that needs interpretation, not pattern matching. The defence there is
-# that Edit/Write are blocked, so it only matters for a deliberate bypass --
-# a policy question, not a technical one.
-_INTERP = r"(?:python[0-9.]*|node|ruby|perl|php|deno|bun)"
-_WRITE_API = (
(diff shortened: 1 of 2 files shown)


POWERFULMOVES:
## Converting to draft — the review is right, this is a stop-ship

Seven findings, all empirically verified against the live `patterns.yaml` (56 `readOnlyPaths`). Two are severe enough that merging would be worse than the gap this PR closes.

**1. Quadratic scan on every Bash command.** `(?=[\s\S]*…)(?=[\s\S]*…)[\s\S]*{path}` is unanchored, so a failed lookahead re-scans from every offset — × 56 paths × 2 forms, in a blocking `PreToolUse` hook:

| command length | before | after |
|---|---|---|
| 1000 | 0.102s | 1.110s |
| 2000 | 0.120s | 3.815s |
| 5000 | ~0.15s | **22.43s** |

A 5 KB heredoc is exactly the shape this PR defends against. And the probe contained no interpreter at all — so that is the *common* path, not a worst case.

**2. Naming a read-only path blocks an unrelated write.** Nothing binds the write verb to the path. Since `readOnlyPaths` includes `/usr/`, `/bin/`, `build/`, `.venv/`, `node_modules/`:

- `/usr/bin/python3 -c "…write_text('a.txt')…"` → **blocked** on `/usr/`
- `source .venv/bin/activate && python -c "open('out.txt','w')…"` → **blocked** on `/bin/`

That bricks a large class of ordinary commands the moment it lands.

**3. It violates its own stated contract.** Reading a protected file and writing elsewhere is blocked — the canonical "regenerate a doc from context" script. My allow-case tests never exercised read-and-write-elsewhere, so the suite reported PASS on a guard that breaks the contract its own comment states.

**4. `\.write\(` matches `sys.stdout.write(`** — blocking a pure read that differs from my passing test only by `print()` vs `sys.stdout.write()`.

**5. The class is not actually closed.** `Path(p).unlink()`, `shutil.rmtree()`, `os.truncate()` all still pass — the two most natural ways to delete from Python.

**6. `NO_DELETE_BLOCKED` was never extended**, so `python -c "os.remove('CLAUDE.md')"` remains allowed.

**7. No `powershell`/`pwsh`** on a Windows-primary fleet.

## The shape was wrong, not just the details

Findings 1-4 share one root: the pattern asserts *"an interpreter runs"* and *"a write verb appears"* independently, then only requires the path to exist somewhere. Binding the path to the write operation — and requiring it to be **quoted** — dissolves all four, because `/usr/`, `.venv/` and `build/` appear unquoted as command fragments, while a real target is always a string literal.

Reworking it that way now.

POWERFULMOVES:
## Reworked — and the review surfaced a worse, pre-existing bug

All seven findings addressed. Reworking also exposed something that was already there and is more serious than the gap this PR set out to close.

### The pre-existing directory bypass

Configured directory paths end with `/`, and `re.escape()` kept that slash **mandatory**. Dropping one character walked past the **entire** guard:

| command | result |
|---|---|
| recursive delete of `pmoves/tools/` | **BLOCKED** |
| recursive delete of `pmoves/tools` | **ALLOWED** |

Same directory, same destruction, one character apart. This applied to **every** pattern class — delete, write, move, truncate — because they all interpolate `{path}`. Every `readOnlyPath` and `noDeletePath` with a trailing slash was open.

`_escape_path()` now makes the slash optional, fixing all classes at once. The boundary is `(?:/|(?![A-Za-z0-9._-]))` — not `/?`, which also matches `pmoves/toolsmith`, and not ``, which treats `tools` followed by `-` as a boundary and so matches `pmoves/tools-old`. Both are different directories. All four cases verified.

### The seven findings

| # | fix |
|---|---|
| 1 | **22s -> 0.001s.** Lookaheads and leading wildcard removed; `re.search` already scans |
| 2 | Path must be **quoted** — `/usr/`, `.venv/`, `build/` appear as bare command fragments; a real target is always a string literal |
| 3 | Verb **bound** to the path, so a read-then-write-elsewhere script stays allowed |
| 4 | The bare write-method token no longer matches `sys.stdout.write(` |
| 5 | Added `Path.unlink`, `shutil.rmtree`, `os.truncate` — the class genuinely was not closed |
| 6 | `NO_DELETE_BLOCKED` now fed by `INTERPRETER_DELETE_PATTERNS` |
| 7 | `powershell`/`pwsh` plus `Set-Content` / `Out-File` / `Remove-Item` added |

Findings 1-4 shared one root: asserting *"an interpreter runs"* and *"a write verb appears"* **independently**, then only requiring the path to exist somewhere. The shape was wrong, not the details — which is why this is a rewrite rather than a patch.

### Residual gap, unchanged

Indirection remains out of reach: a variable holding the path, or a shell loop passing it as `$f`. No regex over command text can expand a variable.

### Verification

- **24 checks pass**, including the three incidental-path allow-cases the first version failed
- a **performance assertion** that fails if the quadratic scan returns
- all **four** existing damage-control tests still pass, despite `_escape_path` being shared by every pattern class

The review is why this is worth merging rather than why it was reverted — the first version would have made every Bash command take seconds and blocked `/usr/bin/python3`.

PR #2561, opened 2026-08-15
docs(a2ui): static integration audit — protocol, pretext, HTML5, Remotion

POWERFULMOVES:
Static audit of the A2UI surface. **Assessment only — no behaviour changed.**

Every finding carries the command that produced it, re-verified against `f27d43ed6`.

## Nine findings

- **F1** — one `a2ui` prefix spans **three unrelated schemas**. `a2ui-renderer` cannot consume what `a2ui-nats-bridge` publishes (zero `beginRendering`/`surfaceUpdate` refs in its `src/`).
- **F2** — `a2ui-v0.1.md` names the vendored Lit bundle as its renderer, twice. `stage.js:145` sets `enableCustomElements = false`, which guards the only branch that could instantiate a `pm-*` element.
- **F3** — pretext has **zero functional use** on the HTML5 surface while the video path depends on it, so web and video can disagree on line breaks. Its `resolveTextLayout()` is already browser-compatible by construction.
- **F4** — `Pmoves-pretext` is uninitialized, and measured: **0 commits ahead, 32 behind** upstream. The fork carries no customization at all.
- **F5** — subject-registry drift, including two subjects that have **never existed in any code file, in any commit**, yet sit in the canonical catalog.
- **F6** — the renderer README documents `comfy.collab.*` inputs it never subscribes to, and an output subject it never publishes.
- **F7** — no CHIT coverage entry for the renderer, whose output feeds a provenance path.
- **F8** — a room manifest pins an "active" app whose route is absent from the static tree (caveated — the OpenRoom shell may resolve it; **not verified**).
- **F9** — `PMOVES-DoX` advertises the A2UI **v0.9** extension and emits **v0.8**.

## Sequencing

Ranked by `(unblocks × risk) ÷ effort`. Items 1–5 are docs/config only. The protocol→animation adapter is flagged **possible YAGNI** pending a use-case answer, and v0.9 migration is explicitly **not** recommended — everything that moves bytes is already v0.8, and `@a2ui/lit` ships only a `0.8` implementation.

## Verification

No services were started; nothing here is live-behaviour evidence. Section 16 lists exactly what was and was not verified, including the limits of a static read.

### DIFF
--- a/pmoves/docs/audit/A2UI_INTEGRATION_AUDIT_2026-08-14.md
+++ b/pmoves/docs/audit/A2UI_INTEGRATION_AUDIT_2026-08-14.md
@@ -27,7 +27,7 @@ individual pieces are, on the whole, well built — the CHIT gate on the bridge 
 tested, the vendored web bundle is genuinely self-contained and reproducibly built, and the
 Remotion renderer is better instrumented than most services in the repo.
 
-The cheapest, highest-leverage work is **not** engineering. Five of the ten findings are
+The cheapest, highest-leverage work is **not** engineering. Five of the nine standing findings (F10 was retracted) are
 documentation and registry corrections that cost hours and prevent the next contributor from
 building against a seam that isn't there. The version split in **F9** — which looks like the
 most alarming finding — turns out to be two wrong strings in one file, not a migration.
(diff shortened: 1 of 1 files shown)


POWERFULMOVES:
**Correction posted to the branch: F10 was added after this PR opened, then retracted.**

I briefly added an F10 claiming the `a2ui-renderer` Docker build could not succeed. It was wrong, and I've retracted it in place (`7e18eeba2`) rather than deleting it, since how it failed is the useful part.

**Refuted by execution:**
```
docker build --target builder -f pmoves/services/a2ui-renderer/Dockerfile pmoves/
# exit 0
```

**Cause:** I read that Dockerfile from the main working tree, which is checked out on `feat/voice-gotime-convergence`, while every other check in this audit ran against `f27d43ed6`. On `main` the Dockerfile is already context-relative and correct. There is no build defect on main.

The nine standing findings (F1–F9) were each verified in a clean worktree at `f27d43ed6` and are unaffected.

Two things did survive the retraction and are recorded:
- Consuming `Pmoves-pretext` from the renderer is blocked on `main` — the submodule sits at the repo root, outside the `pmoves/` build context, and `--ignore-scripts` skips its `prepack` so `dist/` is never produced.
- `feat/voice-gotime-convergence` has no `a2ui-renderer` compose service at all (`no such service`), where `main` defines it at `docker-compose.yml:3766`.

PR #2593, opened 2026-08-18
fix(supabase): Kong crash-loop (3,924 restarts) — declarative config needs the expressions router

POWERFULMOVES:
Supabase Kong had crash-looped **3,924 times**. Kong is the REST gateway on `:8000`, so everything behind it was dark — including **Archon, stuck reporting `credentials loading… ready:false` for three days** while still showing `healthy` to Docker.

**Three stacked faults, each hiding the next.** Two are repo defects fixed here; one was live-env data.

## Fault 1 — `DASHBOARD_PASSWORD` never existed *(fixed live, env-only)*
Absent from `env.shared` **and every tier file**, so compose passed empty and Kong refused the basicauth entry. It was **not even in the funnel's `Missing secrets (non-fatal)` warning** — which is why it survived thousands of restarts unnoticed. Minted via the `secrets-rotate` Known Road.

## Fault 2 — expressions-router syntax the container can't parse *(fixed here)*
Upstream `9777f051d6` (#45462, "restrict rest root anon") added:
```yaml
- name: rest-v1-openapi-root
  expression: 'http.path == "/rest/v1/"'
```
Under the default traditional router `expression` is an unknown field — and **one bad route fails the entire `kong.yml` parse**, taking all 21 services down.

Upstream sets `KONG_ROUTER_FLAVOR: expressions` beside that route (`PMOVES-supabase/docker/docker-compose.yml:95`), but **PMOVES defines its own `supabase-kong` service and never carried the var over**. The submodule's config is *live-mounted*, so upstream's change reached us the moment the submodule bumped — the config outran our service definition. Counts: **1 route uses `expression`, 20 use `paths`.**

*Not* rewriting the route to `paths:` — that's a **prefix** match where upstream means **exact**, which would widen the restricted-anon scope to all REST traffic, and would diverge the fork on an auth-relevant file the next sync would re-break.

## Fault 3 — new opaque API keys aliased to the legacy JWTs *(fixed at 3 sources)*
`SUPABASE_SECRET_KEY`/`SUPABASE_PUBLISHABLE_KEY` are Supabase's **new opaque key model** (`sb_secret_…`), not JWT aliases. They were aliased, so each consumer declared the same key twice → Kong rejected on a uniqueness violation.

Proven by sha256 fingerprint from the live container (values never printed): `PUBLISHABLE == ANON`, `SECRET == SERVICE`.

**Upstream ships both EMPTY** and its entrypoint strips blank entries (*"Remove empty key-auth credentials (unconfigured opaque keys)"*, `kong-entrypoint.sh:47`) — empty is the **supported** state. Fixed so it cannot recur:
- `bootstrap/registry.json` — listed `SUPABASE_SECRET_KEY` as a canonical **alias of `SERVICE_ROLE_KEY`** (the seeding origin)
- `scripts/supabase/apply_env_profile.py:93` — rewrote it on every profile apply
- pinned empty in the kong service so the stack boots regardless of live-env residue

## Bonus — `secrets-audit` "hang" root-caused *(fixed here)*
`make secrets-funnel` appeared to hang forever. It calls `secrets-audit` as step 6 of 7; steps 1–5 (**including the env writes**) had already completed, so operators saw a command that never returned and reasonably assumed total failure.

**Not a worktree/branch artifact.** `candidate_files()` used `REPO_ROOT.rglob("*")` and filtered *after*; rglob can't prune, so it descended into every `node_modules`/`.venv`/`vendor` across ~50 submodules and then **read + regex-scanned** them. Measured: **400,000 paths in 6.2s, still unfinished.** Replaced with pruning `os.walk` → walk **3.8s**, audit **completes in 394s** instead of never.

Also a **correctness** fix found while profiling: worktrees live in **two** places — `.claude/worktrees` (excluded) and top-level `.worktrees/` (**not**) — so **13,398 files / 235 MB** of duplicate repo had every finding double-counted.

## Verification
| Check | Result |
|---|---|
| Supabase stack | **13/13 healthy** |
| Kong | `running`, `healthy`, **restarts=0** |
| REST via Kong | **401** without key, **403** with anon on restricted root (upstream's intended behavior) |
| Archon | recovered to `{"status":"ok"}` — was `ready:false` for 3 days |
| `secrets-audit` | completes in 394s (was: never) |

## ⚠️ Reviewer notes
- **Known follow-up, not in scope here:** `secrets-audit` exits **non-zero on findings**, and there are live ones (placeholder `POSTGRES_PASSWORD`/`SUPABASE_DB_PASSWORD` in tier files). So `make secrets-funnel` **still fails at step 6** even now that it terminates — tracked as lane **L4**.
- **Operator-only residue:** live `env.shared:119-120` still carry the stale aliases. `env.shared` is agent-zero-access by design and `bootstrap_env.py --rotate` **refuses an empty value**, so the pipeline has no clear-a-key operation. The compose pin makes this harmless meanwhile.
- Archon runs its **own** postgres (`pmoves-archon-postgres`) which **name-conflicts on `--force-recreate`** — bring it up without forcing.

### DIFF
--- a/pmoves/docker-compose.core.yml
+++ b/pmoves/docker-compose.core.yml
@@ -480,6 +480,19 @@ services:
     # makes the route table a reviewable file instead of unmanaged DB state.
     - KONG_DATABASE=off
     - KONG_DECLARATIVE_CONFIG=/usr/local/kong/kong.yml
+    # REQUIRED by the declarative config this mounts. Upstream Supabase commit
+    # 9777f051d6 (#45462, "restrict rest root anon") added a route using the
+    # EXPRESSIONS router syntax (expression: 'http.path == "/rest/v1/"').
+    # Under the default traditional router that is an unknown field and the ENTIRE
+    # kong.yml fails to parse, so Kong never starts - one route takes down all 21
+    # services (observed: 3,924 restarts). Upstream sets this same var in
+    # PMOVES-supabase/docker/docker-compose.yml:95; PMOVES defines its own
+    # supabase-kong service and had not carried it over, so the submodule's config
+    # outran this definition. Do NOT 'fix' the route to paths: instead - that is a
+    # PREFIX match where upstream means an EXACT one, which would widen the
+    # restricted-anon scope to all REST traffic.
+    # See docs/handoffs/supabase-kong-declarative-config-boot-failure-2026-08-18.md
+    - KONG_ROUTER_FLAVOR=expressions
     # Supabase's CLI issue #14: force LAST so service short-names resolve.
     - KONG_DNS_ORDER=LAST,A,CNAME
     - KONG_DNS_NOT_FOUND_TTL=1
--- a/pmoves/docker-compose.yml
+++ b/pmoves/docker-compose.yml
@@ -753,6 +753,19 @@ services:
     # makes the route table a reviewable file instead of unmanaged DB state.
     - KONG_DATABASE=off
     - KONG_DECLARATIVE_CONFIG=/usr/local/kong/kong.yml
+    # REQUIRED by the declarative config this mounts. Upstream Supabase commit
+    # 9777f051d6 (#45462, "restrict rest root anon") added a route using the
+    # EXPRESSIONS router syntax (expression: 'http.path == "/rest/v1/"').
+    # Under the default traditional router that is an unknown field and the ENTIRE
+    # kong.yml fails to parse, so Kong never starts - one route takes down all 21
+    # services (observed: 3,924 restarts). Upstream sets this same var in
+    # PMOVES-supabase/docker/docker-compose.yml:95; PMOVES defines its own
+    # supabase-kong service and had not carried it over, so the submodule's config
+    # outran this definition. Do NOT 'fix' the route to paths: instead - that is a
+    # PREFIX match where upstream means an EXACT one, which would widen the
+    # restricted-anon scope to all REST traffic.
+    # See docs/handoffs/supabase-kong-declarative-config-boot-failure-2026-08-18.md
+    - KONG_ROUTER_FLAVOR=expressions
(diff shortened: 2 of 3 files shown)


POWERFULMOVES:
**Split applied** — per your own suggestion, which the operator agreed with.

This PR is now the Kong router-flavor fix alone: 3 files, +198/-0. The two `SUPABASE_SECRET_KEY` commits (`42605b3b7`, `44f510339`) moved to #2595, cherry-picked unmodified with your authorship preserved. Nothing was rewritten or dropped — `git diff` between the old head and the two PRs' combined trees is empty.

Both are independently revertable now, which was the point.

**On the second pushback — you were right, and it's now located.** You noted the residue is masked and the pipeline has no clear-a-key operation. `rotate_secret()` (`scripts/bootstrap_env.py:109`) opens with:

```python
if not value:
    raise ValueError(f"no value to rotate {key!r}: pass --value, or use a generatable --gen-type")
```

`--rotate SUPABASE_SECRET_KEY --value ""` is rejected outright. Since upstream's intended state for `SUPABASE_SECRET_KEY` and `SUPABASE_PUBLISHABLE_KEY` *is* empty, **the secrets pipeline currently cannot express the correct configuration.** That's a capability gap worth its own PR — an explicit `--clear KEY`, or `--value "" --allow-empty` so emptiness is always deliberate.

One correction to how I first wrote this up: I described the remediation as an operator edit to `env.shared`. That was wrong — `env.shared` is machine-written and nobody hand-edits it. The fix goes through an operator-authorized pipeline operation, which is exactly why the missing clear-a-key op is the blocker rather than a footnote.

Also FYI: `#2574` and `#2576` from your suggested order were already merged before this handover — including the squash-dropped re-land you flagged as live-on-main urgent. That one is fixed.

PR #2598, opened 2026-08-18
feat(secrets): --clear KEY — the pipeline could not express an empty value

POWERFULMOVES:
Closes the gap Z890-CLAUDE identified while triaging the Supabase Kong outage, and which the operator asked to have fixed properly rather than worked around.

## The gap

`rotate_secret()` in `pmoves/scripts/bootstrap_env.py` — the only supported writer for `env.shared` — opened with:

```python
if not value:
    raise ValueError(f"no value to rotate {key!r}: pass --value, or use a generatable --gen-type")
```

So `--rotate KEY --value ""` was rejected. `env.shared` is machine-written and nobody hand-edits it, so **"set this key back to empty" had no path through the pipeline at all.**

## Why that matters — it is not hypothetical

Upstream Supabase ships `SUPABASE_SECRET_KEY` and `SUPABASE_PUBLISHABLE_KEY` **empty on purpose** (`PMOVES-supabase/docker/.env.example:47,49`) and its Kong entrypoint strips blank key entries (`kong-entrypoint.sh:47`). A *populated* one is what breaks the deployment.

Aliasing the legacy service-role JWT into `SUPABASE_SECRET_KEY` made Kong declare the same key twice for one consumer, violating `keyauth_credentials` uniqueness, which crash-looped Kong **3,924 times** and took every `/rest/v1`, `/auth/v1` and `/storage/v1` route down (#2593, #2595).

#2595 stopped the pipeline from *writing* the bad value. It could not clear the value already sitting in the live env — item 3 of that handoff, still OPEN. This is the missing half.

## Design

A separate `--clear KEY` flag, **not** permission to pass `--value ""`.

An unset shell variable expands to the empty string. `--value "$MAYBE_UNSET"` silently blanking a live credential is precisely the failure mode worth designing against, so blanking has to be spelled out. `rotate_secret` grows `allow_empty=False`, and the default-path error now names `--clear` so the next person doesn't rediscover this from scratch.

`--clear` and `--rotate` are mutually exclusive (exit 2).

## Verification

12 tests in `pmoves/tests/unit/test_bootstrap_env_clear.py`, every one against a `tmp_path` env file — **none touches the real `env.shared`**, which is zero-access under damage-control.

Falsified rather than assumed: reverting the `allow_empty` guard fails 4 of the 12. Coverage includes surgical replacement (comments and neighbouring keys byte-identical), idempotency, append-when-absent, that empty is *still* refused without the flag, and that the generation / explicit-value / multi-line-refusal / invalid-key paths are unchanged.

Full unit suite: **218 passed**, no regressions (the 1 failure + 7 errors on my box are pre-existing and local — a POSIX-only `chmod` assertion on Windows, and `aiohttp` absent from my system python; CI installs it via `.github/requirements-tests.txt`).

## Follow-up this enables

`SUPABASE_SECRET_KEY` and `SUPABASE_PUBLISHABLE_KEY` can now be cleared through the pipeline:

```
python scripts/bootstrap_env.py --clear SUPABASE_SECRET_KEY
python scripts/bootstrap_env.py --clear SUPABASE_PUBLISHABLE_KEY
make -C pmoves chit-export && make -C pmoves secrets-funnel
```

That is an operator-authorized pipeline operation, which is what closing item 3 actually requires — not an edit.

### DIFF
--- a/pmoves/configs/secrets_cleared.yaml
+++ b/pmoves/configs/secrets_cleared.yaml
@@ -1,0 +1,34 @@
+# Keys deliberately held EMPTY in env.shared — the clear-key tombstone.
+#
+# Why this file has to exist
+# --------------------------
+# `secrets-funnel` runs `secrets-local-hydrate` as its second step, and
+# `hydrate()` overlays a value from local.env whenever the env.shared value is a
+# *placeholder*. The empty string is the first entry in PLACEHOLDER_VALUES
+# (pmoves/services/showtime-api/updater.py), so a key that was deliberately
+# cleared is indistinguishable from one that was never set — and the stale
+# local.env value gets written straight back on the next funnel run.
+#
+# That silently reverses `bootstrap_env.py --clear KEY`. Without this list the
+# clear appears to succeed and the old credential is restored moments later,
+# which is worse than the operation failing outright.
+#
+# Empty is a real configured state, not an absence: upstream Supabase ships
+# SUPABASE_SECRET_KEY and SUPABASE_PUBLISHABLE_KEY empty on purpose and its Kong
+# entrypoint strips blank key entries, so a POPULATED one is what breaks the
+# deployment (#2593 / #2595 — a duplicate key crash-looped Kong 3,924 times).
+#
+# Semantics
+# ---------
+# A key listed here is skipped by `secrets_local_hydrate.hydrate()`, so local.env
+# can never re-populate it. `--force` does NOT override this: forcing exists to
+# push a rotated GH Secret over a stale local value, which is a different intent
+# from "this key is supposed to be empty".
+#
+# `bootstrap_env.py --clear KEY` appends here automatically. Rotating a real
+# value into the key with `--rotate` removes the entry, because the key is no
+# longer deliberately empty.
+#
+# This file holds key NAMES only — never values. It is safe to commit.
+
+cleared_keys: []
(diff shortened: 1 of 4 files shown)


POWERFULMOVES:
**P1 confirmed and fixed** — verified against source rather than accepted, and the finding is correct in every particular.

`secrets-funnel` (codex.mk) runs `secrets-local-hydrate` as its **second** step. `hydrate()` overlays a value from the local overlay file whenever the env-shared value is a placeholder, and the empty string is the **first entry** in `PLACEHOLDER_VALUES` (`pmoves/services/showtime-api/updater.py:37`). So a deliberately-cleared key was indistinguishable from a never-set one, and the stale value came straight back on the next funnel run — making this PR's own documented recipe silently self-reversing. Reporting success and then restoring the credential is worse than failing outright.

**Fix:** `pmoves/configs/secrets_cleared.yaml` records keys that are deliberately empty; `hydrate()` skips them. `--clear` appends, and `--rotate` **removes** the entry, so a clear-then-rotate cannot leave a latent trap that blocks that key forever.

`--force` deliberately does *not* override the tombstone. Forcing means "push a rotated GH Secret over a stale local value" — a different intent from "this key is supposed to be empty".

The list is line-parsed rather than YAML-parsed, matching `pytest_ratchet.py` and `hardening_ratchet.py`: a bootstrap script should not take a pyyaml dependency to do its job, and the header comment is load-bearing documentation a JSON file could not carry. Key **names** only, never values.

**8 new tests (20 total).** One is a deliberate baseline asserting the *original* hazard — `hydrate()` restoring the stale value when no tombstone is present — so the regression cannot return unnoticed. Full unit suite 226 passed, no regressions.

Your suggested alternative ("remove/tombstone the upstream value") is what this implements. I chose the tombstone over mutating the local overlay because that file is machine-synced by `sync-secrets-local.yml` and would simply be repopulated on the next run — moving the problem rather than fixing it.

Thanks for this one. It would have shipped a feature that undid itself.

PR #2601, opened 2026-08-18
fix(fleet): knuckles is reachable at pmoves-b850-ai-top, not "knuckles"

POWERFULMOVES:
Opened while triaging why #2577 failed `validate-command-anchors-ratchet`. The PR was not wrong — the fleet registry was.

## The bug that exists regardless of CI

`pmoves/config/operator_nodes.yaml` gave knuckles `reach: "knuckles"`. Its actual Tailscale hostname is **`pmoves-b850-ai-top`** (operator-confirmed). That file documents `reach` as *"Tailscale hostname (NEVER a raw 100.x IP)"*, so anything resolving knuckles through the capacity registry was aimed at a name that does not resolve. That is a live defect with or without a linter.

Also registers the Jetson edge node `pmoves-nano-1`.

## Why the anchor check was flagging correct documentation

`known_hosts()` in `validate_command_anchors.py` collects `pmoves-[a-z0-9-]+` tokens from `config/fleet-map.yaml` and `config/operator_nodes.yaml`. `pmoves-laptop`, `pmoves-5090`, `pmoves-spark`, `pmoves-z890` are recognised because they appear there as `reach` values.

**The KVMs and Jetsons are absent from the registry entirely, and knuckles' reach was unprefixed.** So every doc that used the *real* hostname tripped `UNKNOWN_HOST` and was recorded as a gap.

The baseline had accumulated **12 such entries — every one a real machine**:

```
pmoves-kvm2   pmoves-kvm4-1   pmoves-kvm4-2   pmoves-b850   pmoves-b850-ai-top
pmoves-nano   pmoves-nano-1   pmoves-9850x3d-r9700   pmoves-pve-01
```

The gate was absorbing correct documentation as defects, because the registry was wrong and the docs were right. That is the failure mode a ratchet is most vulnerable to: the baseline quietly becomes a record of the wrong thing.

## What this changes

Fixes the two hosts with confirmed names and drops their now-stale baseline entries. The ratchet caught the staleness itself — declaring the hosts turned both entries STALE, exactly as designed:

```
507 -> 505 findings.  0 new, 0 stale.  PASS
```

## Deliberately not done

The other **ten** entries need their `reach` values confirmed before the same correction applies. I have not guessed them. An invented hostname is worse than an absent one: it would satisfy the anchor check for docs naming a machine nobody can reach — converting a visible gap into an invisible one.

Same reason only one of the three Jetsons is added.

## Scopes review (requested, reported not acted on)

`pmoves/configs/claws/scopes/*.json` declares 9 scopes. Two observations worth a second opinion:

| finding | detail |
|---|---|
| **hostname convention is split** | `4090/5090/spark` use `pmoves-`-prefixed hostnames; `z890`, `kvm2`, `kvm4-1`, `kvm4-2`, `nemoclaw`, `nemotron-claw` use bare names. The prefixed form is the one that actually resolves. |
| **`nemotron-claw` declares `node: "5090"`** | the same `node` value as the `5090` scope. If it is a profile on the 5090 that is fine; if it is a separate machine, the node id is wrong. |
| **coverage** | no scope exists for knuckles/b850, and only one Jetson (`nemoclaw`, `node: jetson`) is declared out of three. |

Unblocks #2577, whose two NEW findings were exactly these two hosts.

### DIFF
--- a/pmoves/config/operator_nodes.yaml
+++ b/pmoves/config/operator_nodes.yaml
@@ -30,3 +30,11 @@ nodes:
     reach: "pmoves-nano-1"
     vram_gb: 8
     caps: ["cuda", "edge"]
+    # NOT schedulable: this node is not compose-bootstrapped — no /opt/pmoves and
+    # no arm64 worker images — so nothing can consume a job assigned to it. It is
+    # listed anyway so its hostname resolves for topology and `ssh` recipes
+    # naming it stop being recorded as UNKNOWN_HOST gaps. Because select_node()
+    # picks the LOWEST-VRAM match, omitting this flag would make nano-1 win every
+    # job with needs ["cuda"] and min_vram_gb <= 8. Flip to true once the worker
+    # stack is deployed.
+    schedulable: false
--- a/pmoves/services/creator-operator/router.py
+++ b/pmoves/services/creator-operator/router.py
@@ -16,12 +16,28 @@ def load_nodes(path: Path) -> list:
 
 
 def select_node(node_caps: dict, nodes: list):
-    """Lowest-VRAM node that satisfies min_vram_gb and all required caps."""
+    """Lowest-VRAM node that satisfies min_vram_gb and all required caps.
+
+    Nodes marked ``schedulable: false`` are excluded. The capacity registry has
+    to name every machine on the fleet — the anchor validator derives known
+    hostnames from it, so a node absent from the registry makes correct
+    documentation look like a typo. But "this box exists" and "a worker can
+    consume a job on it" are different claims, and the registry could previously
+    only make the second one.
+
+    That gap is not cosmetic here, because this function picks the LOWEST-VRAM
+    match: an un-bootstrapped edge node is not a harmless extra row, it *wins*
+    every small job it nominally qualifies for and those jobs get dispatched to
+    a host with no worker to run them. Defaults to True, so every existing
+    registry entry keeps its current behaviour.
+    """
     need = set(node_caps.get("needs", []))
     min_vram = node_caps.get("min_vram_gb", 0)
     candidates = [
         n for n in nodes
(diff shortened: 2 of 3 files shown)


POWERFULMOVES:
**P2 confirmed and fixed** — verified against source rather than accepted, and the finding is correct.

`creator-operator/config.py:16` loads `pmoves/config/operator_nodes.yaml` as `NODES_PATH`, and `router.select_node()` returns the **lowest-VRAM** node satisfying `min_vram_gb` and caps. So adding nano-1 (8 GB, `cuda`) did not merely record a hostname — it made nano-1 **win** every work order with `needs: ["cuda"]` and `min_vram_gb <= 8`, dispatching those jobs to a host with no `/opt/pmoves` and no arm64 worker images. Nothing could consume them, and because it is the lowest-VRAM entry it would win *preferentially*, not occasionally.

**Fix:** `select_node()` now skips entries with `schedulable: false`, and nano-1 carries that flag. It defaults to `True`, so every existing registry entry keeps its current behaviour.

On your "record the hostname in a topology-only source" alternative: there isn't one the validator reads. `known_hosts()` derives fleet hostnames from exactly two files — `config/fleet-map.yaml` (the Hostinger **VPS** map, semantically wrong for a local Jetson) and `config/operator_nodes.yaml`. Introducing a third file would mean changing the validator too. Marking the node non-schedulable — your second option — keeps one registry that can express both claims, which the fleet needs anyway: the other two Jetsons and any future un-bootstrapped node land in the same state.

**4 tests.** One deliberately flips the flag back to `true` and asserts nano-1 then *does* win — without that, the exclusion test would pass even if the filter did nothing. Router suite 23 passed; anchors ratchet still 0 new / 0 stale.

Good catch. I was solving a hostname-resolution problem and did not check what else consumes that file.

PR #2605, opened 2026-08-18
fix(secrets): host-side CHIT key was never delivered + a blank value beat a populated alias — and the audit that missed it, grounded

POWERFULMOVES:
Two things that fail by silently succeeding, plus the grounding pass that found them.

## The fix (needed on every node)

Every compose file writes `CHIT_PASSPHRASE=${CHIT_PROD_PASSPHRASE:?...}` — the **container-side** name and the **host-side** name differ, and only the container-side one was ever delivered. 26 refs / 5 compose files / 12 services. Because compose interpolates the whole file before running anything, one unresolved `:?` gated **every** `up-*` target on the node, not just those services.

`gh secret list` verified: no `CHIT_PROD_*` secret exists in repo or prod scope, only `CHIT_PASSPHRASE`. So this maps the one real secret to the name the runtime reads rather than minting a duplicate — same precedent as `KIMI_CODING_API` / `MOONSHOT_API_KEY`.

Second, `build_outputs()` resolved a source by key **presence** alone, so a canonical label present-but-**empty** beat a populated alias: written as `KEY=` into every target, and kept out of `missing` even when required. Blank is worse than absent because compose `${KEY:?}` rejects empty while `${KEY?}` accepts it, and anything that *sources* an env file and exports it re-exports the blank — where shell environment beats every `--env-file`.

Third, `SECRETS_DISTRIBUTION_PATTERNS.md` told Pattern-B operators to run `make secrets-funnel` right after pulling a CI bundle. That is the Pattern-A funnel: its `secrets-funnel-sync` depends on `chit-export` (`mk/codex.mk:111`), which re-encodes the node's **local** `env.shared` over the exact file the pull just installed. The doc already contradicted itself — its status line names the correct targets while the copy-pasteable block below carried the destructive one, so **the wrong half was the half operators use**. Artifact name and a stale trade-off corrected in the same block.

## The grounding pass

The instrument-trust audit argued that confident reports need their underlying values read, then published a `157/60/6/4` sweep funnel from a script that had been thrown away, and seven findings with almost no citations. Every claim true; none checkable.

- `pmoves/tools/silent_handler_sweep.py` + `make -C pmoves silent-handler-sweep` — the durable replacement. Stages 1–3 mechanical; **stage 4 is deliberately not automated** and emits `null` in `--json`, because automating that judgement is the error the audit documents.
- The original script was recovered and preserved as `--legacy`, so the historical column stays regenerable. All three columns reconcile: **157 → 160** stage-1, and **6 → 2** pass-only is exactly the four sites fixed in #2572.
- A citation had **rotted into pointing at its own opposite** (`geometry.py:583` → `:587` landed on the correct-discipline counterexample).
- The `merge-gate` paragraph named the wrong thing as repaired. **The job literally named `merge-gate` is still vacuous on today's main** — `PASSED=true`, three echoes, never read — even after #2592. `merge-decision` (`:80`) is the actual enforcer, and it fails only on `"failure"`, so a cancelled or skipped required job passes.

## Provenance, including my own errors

Both sweep predicates were wrong about silence, in opposite directions. The original substring-scanned `ast.dump(handler)` — a text grep wearing an AST costume, inside a sweep whose headline is that text greps cannot see handler shape. The replacement's first draft counted `Return` as audible, which excludes `except Exception: return _FALLBACK` — finding #7 itself. Both pinned as named regression tests.

And three wrong diagnoses of the secrets blocker are recorded in the register claim, including one briefly committed to a source comment as fact — finding #6, by the audit's own author. **The real cause was an empty export in one session's shell shadowing correct tier files.** With it removed, `docker compose config` is clean across the whole stack; the tier files were right all along. The remaining residue is the blank key in `env.shared` — lane **L5**, operator-only, since no pipeline operation clears a key.

## Verification

- `secrets-funnel` run under explicit operator authorization after verifying targets: rc=0, 19s, all 7 steps.
- **100 passed** in `tests/tools/` (2 pre-existing failures unrelated: one baselined at `_known_failures.yaml:166`, the other the same shell-export class via `NATS_URL`).
- Every line citation re-derived against main after 49 commits; 13 unchanged, 4 drifted and corrected.
- Register claim is **append-only** — 3 insertions, 0 deletions — and CHIT-signed (`HMAC-SHA256`, kid `chit-signing-v01`), identity resolving to the registered glyph rather than the fallback: **finding #7's fix verified in production**, not in a test.

## Found but not claimed

`make chit-manifest-register ARGS='--check'` — the documented gate — **cannot work**. Command-line `ARGS` is exported to sub-makes and `env-bootstrap-lite` forwards it to a tool with no such flag (`mk/preflight.mk:39`). **7 targets** share that prerequisite (`codex.mk:84,91,98`; `preflight.mk:52,59,240,247`). Fix is `$(MAKE) ARGS= env-bootstrap-lite`. Left for whoever holds the Make lane.

Also: L4 on the #2594 board (`secrets-audit` fails at step 6) **does not reproduce on B850** — 12s, exit 0, 0 errors / 5 warnings. `secrets_hardening_audit.py:345` is `return 1 if errors else 0`, so it is conditional on ERROR-level findings, not universal.

### DIFF
--- a/pmoves/tools/silent_handler_sweep.py
+++ b/pmoves/tools/silent_handler_sweep.py
@@ -313,7 +313,11 @@ def sweep(paths: Iterable[str], legacy: bool = False) -> List[Site]:
             # rather than letting it vanish into a clean-looking total.
             unparsed += 1
             continue
-        rel = str(path.relative_to(_REPO_ROOT))
+        # .as_posix(), not str(): on Windows str() yields backslashes, so every
+        # emitted path — and anything derived from it, including a baseline or a
+        # cross-referenced report — differs by platform for the same file. Forward
+        # slashes are what the rest of this repo's tooling records.
+        rel = path.relative_to(_REPO_ROOT).as_posix()
         for node in ast.walk(tree):
             if not isinstance(node, ast.Try):
                 continue


POWERFULMOVES:
**Node-side verification from SPARK (review lane):** your diagnosis reproduces on a live node, with one additional wrinkle the alias mapping alone won't fix.

SPARK's `env.shared` carries **both** keys with **different values** — `CHIT_PROD_PASSPHRASE` (43 chars) and `CHIT_PASSPHRASE` (64 chars). Compose interpolation reads `--env-file env.shared`, so the 12 services get the 43-char value; host-side `make sign-trail` resolves via the `CHIT_SIGNING_KEY` > `CHIT_PASSPHRASE` priority chain (no `CHIT_SIGNING_KEY` set here) and signs with the 64-char value. Result: host receipts and any container-internal signing use **different key material** — signatures can't cross-verify.

The manifest alias (`CHIT_PASSPHRASE` → both targets in `.env.generated`) is correct for generated outputs, but a hand-set divergent `CHIT_PROD_PASSPHRASE` in `env.shared` silently wins for every compose interpolation, since env.shared feeds `COMPOSE_ENV_FILES` directly. Suggest the PR's node checklist include: **diff the two values before adopting the mapping, keep the one your existing receipts signed with, delete the divergent one** — otherwise the alias propagates whichever value the funnel happens to resolve first.

Also confirming the `build_outputs()` blank-beats-alias finding from the consumer side: SPARK's funnel run today reported `errors=0 warnings=5` with the tier-agent gap — consistent with presence-only resolution masking partially-empty sources.

POWERFULMOVES:
Confirmed and addressed in `b49641227` — thank you, this is a real gap my change does not close.

You're right that the manifest alias only governs **generated** outputs. A hand-set divergent value in `env.shared` wins for every compose interpolation regardless, because it feeds `COMPOSE_ENV_FILES` directly. The alias mapping cannot fix that, and neither can the funnel: which value is authoritative is a question only someone who knows what their receipts were signed with can answer.

You asked for a node checklist. I've made it a **check** instead — a checklist is a documentation fix for an instrument problem, which is the thing this lane keeps arguing against. Audit section 8c walks the v2 manifest's alias map and reports any entry whose label and aliases are both set with more than one distinct non-blank value.

**It found a divergence on its first run, and not the one it was written for:**

```
[WARN] pmoves/env.shared: alias divergence: KIMI_CODING_API (len 72), MOONSHOT_API_KEY (len 51)
  are the same manifest entry 'kimi_coding_api' but hold DIFFERENT values.
```

B850 has had both Kimi names set to different keys, silently. Whichever the funnel resolves first wins, and TensorZero's gateway reads the legacy name — so the provider may be authenticating with a key nobody chose. Same shape as your CHIT finding, different entry, and it would have kept passing every per-key check because both keys are populated.

Design points worth flagging for your node:

- **WARN, not ERROR.** Erroring exits non-zero at funnel step 6 and would strand precisely the nodes that have the problem — SPARK included.
- **Lengths, never values.** Your 43-vs-64 is exactly the right diagnostic granularity, so that is what it emits.
- **A blank on one side is not divergence.** That is the blank-beats-alias case fixed in `62ead6404`, and it is why the CHIT pair stays quiet on B850, where the host-side name is zero-length rather than divergent. Your node should trip on the CHIT entry where B850 does not.

The remediation text the finding prints is your checklist, verbatim: decide which value your existing receipts were signed with, delete the other, then re-funnel.

One thing I cannot verify from here and you can: whether SPARK's 43-char value or its 64-char value is the one your existing receipts signed with. The check deliberately will not guess.

Also confirming your second point from this side — the `build_outputs()` fix is `62ead6404`, and `errors=0 warnings=5` with a tier-agent gap is the same signature I saw here before the fix.

PR #2609, opened 2026-08-19
fix(docker): correct non-root stanza for Alpine and distroless stages (#2604 follow-up)

POWERFULMOVES:
## Summary
Follow-up to #2604, found during live bring-up of the agents overlay on SPARK (the same run that validated #2604's restores — `mai-ui-agent` built and started):

1. **notebooklm-agent** (`node:20-alpine`): #2604's restored `RUN useradd ... || true` silently no-ops on Alpine (no `useradd`), leaving `USER pmoves` pointing at a user that was never created — the daemon refused to start the container (`unable to find user pmoves: no matching entries in passwd file`). Now uses `adduser -D -s /bin/sh pmoves` with **no silent fallback**.
2. **agentgym-rl-coordinator** (final stage `gcr.io/distroless/python3-debian12:nonroot`): distroless has no shell — the appended `RUN useradd` cannot execute and would break the build outright. Switched to the stage's built-in `nonroot` user, matching the pre-existing `HF_HOME=/home/nonroot` the image was designed around.

## Testing
- [x] `docker build` notebooklm with compose's real context (`./services/notebooklm-agent`) → image carries `pmoves:x:1001`, `id` runs non-root
- [x] `python3 tools/hardening_ratchet.py` (with #2606's NO_FROM check) → 12/12 baselined, 0 new
- [ ] full `overlay-up-agents` re-run (SPARK, next)

### DIFF
--- a/pmoves/services/agentgym-rl-coordinator/Dockerfile
+++ b/pmoves/services/agentgym-rl-coordinator/Dockerfile
@@ -24,6 +24,7 @@ ENV HF_HOME=/home/nonroot/.cache/huggingface
 EXPOSE 8114
 CMD ["python", "-m", "uvicorn", "app:app", "--host", "0.0.0.0", "--port", "8114"]
 
-# Non-root runtime (house pattern; restores #2285 intent without truncating)
-RUN useradd --create-home --shell /bin/bash pmoves 2>/dev/null || true
-USER pmoves
+# Non-root runtime: the final stage is distroless (no shell, no useradd, no RUN).
+# It ships a built-in `nonroot` user (uid 65532) — and the HF_HOME above already
+# points at /home/nonroot, so this is the user the image was designed around.
+USER nonroot
--- a/pmoves/services/notebooklm-agent/Dockerfile
+++ b/pmoves/services/notebooklm-agent/Dockerfile
@@ -30,6 +30,9 @@ ENV GOOGLE_REFRESH_TOKEN=""
 # (a detached  as PID 1 would hit EOF on stdin and exit-loop.)
 CMD ["tail", "-f", "/dev/null"]
 
-# Non-root runtime (house pattern; restores #2285 intent without truncating)
-RUN useradd --create-home --shell /bin/bash pmoves 2>/dev/null || true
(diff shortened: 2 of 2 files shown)


POWERFULMOVES:
Reviewed and verified — both fixes are correct and the scope is complete.

**Scope check.** #2604 restored 9 Dockerfiles and appended the same `RUN useradd ... || true` to each. Enumerated the final `FROM` of all nine:

```
agentgym-rl-coordinator   gcr.io/distroless/python3-debian12:nonroot   <- broken
notebooklm-agent          node:20-alpine                               <- broken
audio-reprocess           python:3.11-slim
flute-gateway .bpm        python:3.11-slim
flute-gateway .minimal    python:3.11-slim
hf-mcp-server             python:3.12-slim
kokoro-tts .kokoro        python:3.11-slim
mesh_exposure             python:3.12-slim
owner-presence            python:3.11-slim
```

Exactly the two you fixed. The other seven are Debian-based where `useradd` exists, so 2-for-2 is the whole set — not a partial sweep.

**distroless.** Confirmed the final stage contains no `RUN` at all (only `COPY`/`ENV`/`WORKDIR`/`EXPOSE`/`CMD`), so an appended `RUN` would have failed the build outright rather than degrading. And `ENV HF_HOME=/home/nonroot/.cache/huggingface` was already there, so `USER nonroot` matches what the image was built around instead of inventing a second home.

**alpine.** `adduser -D -s /bin/sh pmoves` is the right busybox form, and dropping `|| true` is the important half — that swallow is exactly what turned a missing user into a runtime failure (`no matching entries in passwd file`) instead of a build failure.

**Ratchet:** `hardening_ratchet.py` on the rebased branch → `Findings: 12 (12 baselined, 0 new)`. `USER nonroot` passes the last-USER rule as expected.

**One observation, not a blocker.** The remaining seven still carry `|| true`. On Debian bases `useradd` exists so it will not fire, but the construct is what made this class of failure silent in the first place — a swallowed failure leaves `USER pmoves` dangling and the container dies at start rather than at build. Worth removing on the next pass through those files; deliberately out of scope here.

PR #2617, opened 2026-08-19
chore(supabase): image currency — kong 3.9.3, studio 2026.08.03 (low-risk L2 slice)

POWERFULMOVES:
**Applied and verified live on z890.** L2 currency, split by blast radius rather than done as one sweep.

## Drift, measured against the submodule (not assumed)
| service | upstream | pmoves | action |
|---|---|---|---|
| kong | 3.9.3 | 3.9.1 | **bump** |
| studio | 2026.08.03 | 2026.06.03 | **bump** |
| db | 17.6.1.136 | 17.6.1.108 | deferred — data tier |
| analytics (logflare) | 1.43.1 | 1.30.3 | deferred — 13 minor versions |
| vector | 0.53.0 | 0.28.1 | deferred — ~25 releases |
| **gotrue** | v2.189.0 | **v2.191.0** | ⚠️ **DO NOT TOUCH — ours is NEWER** |

### ⚠️ The gotrue trap
PMOVES runs a **newer** gotrue than upstream pins. "Syncing to upstream" would **downgrade the auth service**. Any bulk-bump tooling that assumes upstream-is-newer would silently regress it. Left alone deliberately, and **verified still v2.191.0** after this change.

## Why only these two
Kong is patch-level and is the component most recently proven out (#2593); studio is UI-only with no data path. The deferred items each need their own window — a postgres image bump on a live volume isn't a "currency" change, and logflare owns the `_analytics` schema inside `_supabase`, where a 13-version jump can carry migrations.

**On vector specifically:** its `vector.yml` is **live-mounted from the submodule**, so the config is already written for upstream's newer binary. Static analysis shows only long-stable component types (`docker_logs`, `remap`, `route`, `filter`) — which is why it runs healthy despite the gap. So that drift is **currency/CVE hygiene, not a live breakage**. Worth bumping deliberately, not as a drive-by.

## Verified after `make -C pmoves up-supabase`
| Check | Result |
|---|---|
| kong | `kong/kong:3.9.3` — **healthy, restarts=0** |
| studio | `2026.08.03-sha-022b374` — healthy |
| gotrue | `v2.191.0` — **unchanged** |
| stack | **12/13 healthy**, 0 unhealthy, 0 restarting |
| Kong declarative config | **0 parse errors** (the #2593 failure mode) |
| REST via Kong | **401** without apikey, **403** with anon on restricted root |

That last row is the point of verifying a *gateway* bump: routing and key-auth still behave exactly as #2593 established.

> One service read `UNHEALTHY` mid-rollout — it was startup lag; storage settled to healthy with the server listening and no errors. Checked rather than assumed either way.

Compose edit made under the `compose:` Known Road with the handoff on disk; bypass recorded in `known-roads.jsonl`.

### DIFF
--- a/pmoves/docker-compose.core.yml
+++ b/pmoves/docker-compose.core.yml
@@ -463,7 +463,12 @@ services:
           memory: 512M
   supabase-kong:
     <<: *tier-supabase-hardened
-    image: kong/kong:3.9.1
+    # Matches upstream PMOVES-supabase/docker/docker-compose.yml. Kong is the REST
+    # gateway; keeping it aligned with the submodule matters because its declarative
+    # config (kong.yml) is LIVE-MOUNTED from that submodule, so upstream config
+    # changes reach this container the moment the gitlink bumps - which is exactly
+    # how the expressions-router outage (#2593) arrived. Patch-level bump.
+    image: kong/kong:3.9.3
     profiles:
     - supabase-local
     volumes:
@@ -797,7 +802,9 @@ services:
   supabase-studio:
     <<: *tier-supabase-hardened
     # Pinned 2025-02-04: Latest stable from Supabase
-    image: supabase/studio:2026.06.03-sha-0bca601
+    # Matches upstream. Dashboard UI only - no data path and nothing depends on it,
+    # so this is the lowest-risk half of the L2 currency slice.
+    image: supabase/studio:2026.08.03-sha-022b374
     restart: unless-stopped
     profiles:
     - supabase-local
--- a/pmoves/docker-compose.yml
+++ b/pmoves/docker-compose.yml
@@ -752,7 +752,12 @@ services:
   #  strips comments trailing the `environment:` sequence on its round-trip.)
   supabase-kong:
     <<: *tier-supabase-hardened
-    image: kong/kong:3.9.1
+    # Matches upstream PMOVES-supabase/docker/docker-compose.yml. Kong is the REST
+    # gateway; keeping it aligned with the submodule matters because its declarative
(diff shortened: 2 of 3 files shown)


POWERFULMOVES:
Reviewed — verified independently and merging.

**Tags exist.** The load-bearing check for an image bump is that the target actually pulls; a bad tag is an unpullable image and a dead bring-up. Queried Docker Hub directly:

```
kong/kong:3.9.3                        HTTP 200  EXISTS
supabase/studio:2026.08.03-sha-022b374 HTTP 200  EXISTS
supabase/gotrue:v2.191.0               HTTP 200  EXISTS
```

**The gotrue trap holds, and by a wider margin than the table shows.** My submodule checkout is older than yours — it pins `gotrue:v2.185.0` and `studio:2026.01.27` where your table reads `v2.189.0` / `2026.08.03`. So I could not confirm your upstream column from here, and I am flagging that rather than implying I did. What the direction check does confirm is the part that matters: PMOVES runs **v2.191.0**, newer than either upstream snapshot, so a bulk "sync to upstream" would downgrade the auth service from two different starting points. Confirmed the compose diff touches **no** gotrue image line — the only two `gotrue` mentions in the diff are in the handoff doc.

**Overlay consistency.** `docker-compose.core.yml` and `docker-compose.yml` show identical diff shape (11 insertions / 4 deletions each), which is what a source edit plus `compose-split` regenerate looks like rather than two hand-edits drifting. `Validate Compose Files` agrees.

All five required checks green. Splitting by blast radius — patch-level gateway and a no-data-path dashboard now, data tier and the 13/25-release laggards deferred — is the right call on a stack that was crash-looping on Kong three days ago.

PR #2635, opened 2026-08-19
fix(consciousness): copy the whole services/common package, not two files

POWERFULMOVES:
## Summary
`consciousness-service` crash-loops at import on current images:

```
File "/app/chr_algorithm.py", line 28
from pmoves.services.common.env import get_secret
ModuleNotFoundError: No module named 'pmoves.services'
```

Chain: `chr_algorithm` tries `from services.common.env import get_secret` → executing the package `__init__` → which eagerly star-imports `.telemetry` and `.model_nexus` (added in c186c1bd4) → neither is in the image (Dockerfile copied only `__init__.py` + `env.py`) → ImportError → the `pmoves.services` fallback isn't packaged either → dead before uvicorn binds. Reproduced live on SPARK (container Restarting).

Fix: whole-package COPY — the house pattern every other consumer already uses (pdf-ingest, nats_event_bus):
```dockerfile
COPY services/common /app/services/common
```

## Testing
- [x] Rebuilt image on SPARK: `from services.common.env import get_secret` → OK, callable
- [x] Confirmed broken image (`ghcr pmoves-consciousness:pmoves-latest`) fails the same import with `services.common.telemetry` missing

## Also observed this pass (routed, not this PR)
- `hi-rag-gateway` (v1, legacy): lock pins torch 2.0.1 but unpinned transformers resolved to a release whose `flex_attention` integration needs `torch.compiler` → import crash. v2 (active) healthy. Needs a transformers constraint in the v1 lock — Hi-RAG lane.

Generated with Crush


POWERFULMOVES:
**Correction to the routed finding in this PR's body (SPARK-KIMI, post-audit):** I wrote "lock pins torch 2.0.1 but unpinned transformers resolved to a release needing torch.compiler" — **wrong on mechanism**. Research audit of `pmoves/services/hi-rag-gateway/requirements.lock` shows it pins `torch==2.9.1` + `transformers==4.57.3` — byte-identical to v2's known-good combination.

The actual defect: **`Dockerfile:16` re-installs torch/torchvision/torchaudio UNPINNED from the cu-index after the lock install** (`pip install --index-url .../whl/${TORCH_CUDA_VERSION} torch torchvision torchaudio || true`) — the CUDA wheel index resolved an old torch (2.0.1 observed in the built image on SPARK), clobbering the locked 2.9.1 and breaking the locked transformers/tokenizers ABI. The `|| true` hid it. v2's GPU Dockerfile pins `TORCH_VERSION`/`TORCHAUDIO_VERSION` build args — v1 just needs the same pin (or retirement).

**Additional HiRAG findings for the lane owner (from the same audit):**
- v1 has **no profile gate** despite its README claiming `profiles: legacy` — it starts on every default `up`. No code consumers dial 8089/8187 (only archon env exports + `HIRAG_V1_HOST_PORT`). Retirement candidate.
- `session-context-worker` gets `HIRAG_INGEST_URL=.../ingest` — **v2 has no `/ingest` route** (only `/hirag/upsert-batch` / `/hirag/upsert-provenance`); already documented as a trap in the Fordham ballot doc.
- Wrong health path `/healthz` on 8086/8087 in several configs (skill-pairings.yaml, mini_cli, agent_registry `hirag_v2` block) — v2 liveness is `/`.
- #2646's mcp bridge sends `top_k`/`rerank` but v2's `QueryReq` uses `k`/`use_rerank` — Pydantic silently drops them (defaults apply).
- CHIT: signature **verification** live + fail-closed at v2's geometry boundary; **Neo4j persistence still unbuilt** (Postgres-gated only), as CHIT_INTEGRATION_STATUS §10 says.

Issue #2639, opened 2026-08-19
submodule-branch-policy-check has been exiting non-zero on main with 7 mismatches


POWERFULMOVES:
## Measured state

`make -C pmoves submodule-branch-policy-check` exits 1 on `origin/main` with 7 mismatches. Measured at pure `origin/main`, not on any PR branch:

```
- PMOVES-nats-server                 -> main    (expected: PMOVES.AI-Edition-Hardened)
- skills/PMOVES-skills               -> main    (expected: PMOVES.AI-Edition-Hardened)
- PMOVES-MiniMax-MCP                 -> main    (expected: PMOVES.AI-Edition-Hardened)
- Pmoves-minimax-cli                 -> main    (expected: PMOVES.AI-Edition-Hardened)
- Pmoves-MiniMax-Provider-Verifier   -> main    (expected: PMOVES.AI-Edition-Hardened)
- PMOVES-agents.md                   -> main    (expected: PMOVES.AI-Edition-Hardened)
- pmoves-keygen                      -> master  (expected: PMOVES.AI-Edition-Hardened)
```

These are not drift in the "gitlink points somewhere unexpected" sense. Each of these submodules genuinely tracks the branch it declares; the checker's default is `SUBMODULE_BRANCH_DEFAULT = PMOVES.AI-Edition-Hardened` and these are simply not on it.

## Triage

**One is not a real mismatch at all.** `skills/PMOVES-skills` is a duplicate registration of `skills/Pmoves-skills` — same repository, same commit, two case-differing paths. Tracked separately in #2638. Resolving that one removes this row.

**Five are first-party or upstream-tracking forks** where `main` is plausibly correct and the fix is an allowlist entry, not a re-pin: `PMOVES-nats-server`, `PMOVES-MiniMax-MCP`, `Pmoves-minimax-cli`, `Pmoves-MiniMax-Provider-Verifier`, `PMOVES-agents.md`. Each needs the same check that was applied to the three in #2632 — confirm the declared branch actually exists on the fork, then allowlist it. That check is cheap (`gh api repos/POWERFULMOVES/<repo>/branches`) and deterministic.

**One is intentional and should stay red until it clears on its own.** `pmoves-keygen` is pinned to the head of open PR `POWERFULMOVES/pmoves-keygen#1`; per #2632 it resolves when that merges and the gitlink is re-pinned to `master`.

## Why this sat unnoticed

[...]

Issue #2638, opened 2026-08-19
skills/Pmoves-skills and skills/PMOVES-skills are the same repo registered twice — paths collide on Windows/macOS


POWERFULMOVES:
## The finding

`.gitmodules` registers the same repository twice, at two paths that differ only in case:

```
[submodule "skills/Pmoves-skills"]
	path = skills/Pmoves-skills
	url = https://github.com/POWERFULMOVES/Pmoves-skills.git
	branch = main

[submodule "skills/PMOVES-skills"]
	path = skills/PMOVES-skills
	url = https://github.com/POWERFULMOVES/PMOVES-skills.git
	branch = main
```

Those two URLs are **the same repository**. GitHub resolves both to a repo whose current name is `PMOVES-skills` — identical `parent`, `pushed_at` and `size` on both:

```
POWERFULMOVES/Pmoves-skills  -> name=PMOVES-skills  fork=true  parent=vercel-labs/skills  pushed=2026-08-14T20:53:33Z  size=1482
POWERFULMOVES/PMOVES-skills  -> name=PMOVES-skills  fork=true  parent=vercel-labs/skills  pushed=2026-08-14T20:53:33Z  size=1482
```

The repo was renamed `Pmoves-skills` -> `PMOVES-skills`; the old URL still redirects, so both entries clone successfully and nothing complains at fetch time.

Both gitlinks point at the **identical commit**:

```
$ git ls-tree origin/main skills/
160000 commit c6f69c631292444cc541ac6d91e2226b0ff247da	skills/PMOVES-skills
160000 commit c6f69c631292444cc541ac6d91e2226b0ff247da	skills/Pmoves-skills
```

## Why it matters

**It collides on case-insensitive filesystems.** Windows and macOS cannot hold both `skills/Pmoves-skills` and `skills/PMOVES-skills` as distinct directories. Two submodule entries want to occupy one path.

It also produces downstream noise that reads as unrelated drift:

- `submodule-branch-policy-check` reports `skills/PMOVES-skills` as a mismatch, because `SUBMODULE_BRANCH_ALLOW` in `pmoves/mk/preflight.mk` lists only `skills/Pmoves-skills=main`. It looks like a casing bug in the allowlist. It is not — the allowlist is correct for the entry it names, and the second entry is the actual problem.
- `submodule-layer-validate` flags both paths with `errors=2` each, where every other submodule reports 1.

[...]

Issue #2637, opened 2026-08-19
audit-layers-static is documented as a release gate but is referenced by zero workflows


POWERFULMOVES:
## What was measured

`audit-layers-static` is documented as a certification pass — `CODEX_CLAUDE_PARITY_MAP.md` maps `/deploy:audit-layers` to it, and ~70 files under `pmoves/docs/evidence/submodule_layer/` instruct operators to run it. It is **referenced by zero GitHub Actions workflows**:

```
$ grep -rln "audit-layers-static" .github/workflows/
(no matches)
$ grep -rln "submodule-branch-policy-check\|submodule_branch_policy_check" .github/workflows/
(no matches)
```

That is the whole answer to why it never fails: **it never runs.** It is an operator-invoked make target only.

## Why that matters right now

`submodule-branch-policy-check` — the third target in the `audit-layers-static` chain — has been exiting non-zero on `main`:

```
$ make -C pmoves submodule-branch-policy-check
FAIL: submodule branch policy check failed
Branch policy mismatches:
  - PMOVES-nats-server ... - skills/PMOVES-skills ... - PMOVES-MiniMax-MCP
  - Pmoves-minimax-cli ... - Pmoves-MiniMax-Provider-Verifier
  - PMOVES-agents.md  ... - pmoves-keygen
exit 1  (7 mismatches)
```

Measured at pure `origin/main`, so it is not the product of any open PR. It has been red and blocking nothing.

This is the same shape as two defects already fixed this week — the `merge-gate` job that was `echo "passed=true"` (removed in #2630) and the `verifier-gate` that could never report (fixed in #2631). A gate that cannot say NO reads, in every dashboard, exactly like a gate that said YES.

## Also worth deciding

The chain's last line is `@$(MAKE) --no-print-directory docs-reconcile-check || true` — a fail-open inside a certification pass. If `docs-reconcile-check` is advisory that is fine, but it should be stated rather than expressed as `|| true`.

`make` stops the chain at the first failing target, so whichever gate fails first suppresses every gate after it. With 12 targets in the chain that is a meaningful amount of unrun certification.

## What would resolve this

A decision, then the wiring to match it. Either:

[...]

PR #2590, opened 2026-08-17
feat(submodule): 6-repo fold-in PR3 - PMOVES-agents.md (agentsmd/agents.md open format)

POWERFULMOVES:
## 6-repo fold-in — PR3 (PMOVES-agents.md)

**Scope:** PR3 of 3 in the 6-repo fold-in slice. Operator chose 3 separate PRs for revert granularity.

| PR | Submodule(s) | Status | LEARNINGS file |
|----|--------------|--------|----------------|
| PR1 | `skills/PMOVES-skills/` (vercel-labs/skills post-recenter home) | **MERGED** as #2586 at `d4af79f325` | `skills_foldin_pr1_LEARNINGS.md` |
| PR2 | `PMOVES-MiniMax-MCP` + `Pmoves-minimax-cli` + `Pmoves-MiniMax-Provider-Verifier` | parallel worktree | `mcpcli_foldin_pr2_LEARNINGS.md` |
| **PR3 (this)** | **`PMOVES-agents.md/`** (agentsmd/agents.md open format) | **OPEN** | `agents_md_foldin_pr3_LEARNINGS.md` |

## What lands (1 new top-level submodule)

- `.gitmodules` — new `[submodule "PMOVES-agents.md"]` entry
  - `path = PMOVES-agents.md`
  - `url = https://github.com/POWERFULMOVES/PMOVES-agents.md.git`
  - `branch = main`
- `PMOVES-agents.md/` — new gitlink pointer at the upstream HEAD `d1ac7f063d20e70015ed6732664049ae4ba9d74e` (POWERFULMOVES fork of `agentsmd/agents.md` main)

**Why top-level (not under `skills/`):** `PMOVES-agents.md` is a **format/taxonomy reference** (the agents.md open spec + agent taxonomy + persona schema), not a runtime-invokable agent skill. Agents *load* the format spec on Tier-2; they don't *invoke* it. The `skills/` tree is reserved for runtime skills (claude-d3js, agent-sandbox, etc.).

## Files touched (PR3 net additions only)

| File | Change | Notes |
|------|--------|-------|
| `.gitmodules` | +4 lines | new `[submodule "PMOVES-agents.md"]` block (path/url/branch) |
| `PMOVES-agents.md` | new gitlink | mode 160000, points to `d1ac7f0...` |
[...]

PR #2642, opened 2026-08-19
fix(submodules): allowlist the 5 forks that have no hardened branch

POWERFULMOVES:
Resolves the deterministic part of #2639.

`submodule-branch-policy-check` has been exiting non-zero on `main` with 7 mismatches. Five are the same class #2632 fixed for three others: the submodule's declared branch is correct, and `SUBMODULE_BRANCH_ALLOW` simply does not list it, so it falls through to `SUBMODULE_BRANCH_DEFAULT`.

## Verified before adding

`pmoves/mk/preflight.mk` states the invariant for itself — *"Every branch here is verified to exist on its remote."* Checked each against the live branch list:

| submodule | declares | branches on the fork |
|---|---|---|
| PMOVES-nats-server | main | `feat/pmoves-config`, `main` |
| PMOVES-MiniMax-MCP | main | `main` |
| Pmoves-minimax-cli | main | `main` |
| Pmoves-MiniMax-Provider-Verifier | main | `main` |
| PMOVES-agents.md | main | `main` |

**None of the five has a `PMOVES.AI-Edition-Hardened` branch at all**, so the default could never have been satisfiable for them. Four have exactly one branch.

```
before  7 mismatches, exit 1
after   2 mismatches, exit 1
```

## What is deliberately left

**`skills/PMOVES-skills` is not a real mismatch.** It is a duplicate registration of `skills/Pmoves-skills` — same repository (the old name redirects), same gitlink commit `c6f69c63`, two paths differing only in case. Tracked as #2638. Resolving it *removes* this row; allowlisting it would paper over a path collision that breaks on Windows and macOS.

**`pmoves-keygen` is intentional.** Pinned to the head of open PR `POWERFULMOVES/pmoves-keygen#1`; it clears when that merges and the gitlink is re-pinned to `master`.

## This does not turn the gate green, and is not meant to

It removes every row removable by evidence, leaving two that each need a *decision* rather than a *measurement*. Per #2637, the gate cannot be wired into CI until both close out — wiring it sooner would install a permanently-red required check.

### DIFF
--- a/pmoves/mk/preflight.mk
+++ b/pmoves/mk/preflight.mk
@@ -25,7 +25,12 @@ SUBMODULE_BRANCH_ALLOW ?= PMOVES-DoX=PMOVES.AI-Edition-Hardened-DoX,\
   skills/Pmoves-claude-d3js-skill=main,\
   PMOVES-jcodemunch-mcp=main,\
   pmoves-hirag-mcp=main,\
-  PMOVES-Spark-VSS=main
+  PMOVES-Spark-VSS=main,\
+  PMOVES-nats-server=main,\
+  PMOVES-MiniMax-MCP=main,\
+  Pmoves-minimax-cli=main,\
+  Pmoves-MiniMax-Provider-Verifier=main,\
+  PMOVES-agents.md=main
 AUDIT_RUNTIME_GPU ?= 0
 PRECHECK_VENV_WIN ?= .venv-pmoves/Scripts/python.exe
 PRECHECK_VENV_UNIX ?= .venv-pmoves/bin/python


POWERFULMOVES:
**Holding this — #2643 supersedes its approach, and I think #2643 is right.**

#2643 makes the argument directly: #2632 (which I merged) cleared three gitlink FAILs by declaring `main` instead of `PMOVES.AI-Edition-Hardened`, and *"that fixed the symptom by moving three submodules off the fleet branch strategy. The right fix is the branch."* It then creates the hardened branch on each fork at exactly the commit the gitlink already pinned — zero content divergence, reversible by deleting three refs.

This PR does the same thing #2643 is undoing, for five more submodules. Merging both would have the repo arguing with itself in one file.

#2643 also surfaces something that undercuts the mechanism this PR relies on:

> `submodule_branch_policy_check.py` computes `accepted = {default} | allowed[name]`, so **`SUBMODULE_BRANCH_ALLOW` is additive** — an entry naming the wrong branch can only ever permit more, never fail.

So an allowlist entry cannot be wrong in a way the gate detects. That makes "add an exception" the weaker of the two fixes on its own terms, independent of which branch is correct. #2643's new `test_submodule_branch_parity.py` closes that hole by failing on a *stale exemption* — an entry that outlives the disagreement it documents. The five entries here would pass that test today (they agree with `.gitmodules`), but they would become stale the moment anyone creates a hardened branch on those forks.

## What is still true and worth keeping

The measurement holds: none of these five has a `PMOVES.AI-Edition-Hardened` branch, so the default is currently unsatisfiable for all of them.

| submodule | declares | branches that exist |
|---|---|---|
| PMOVES-nats-server | main | `feat/pmoves-config`, `main` |
| PMOVES-MiniMax-MCP | main | `main` |
| Pmoves-minimax-cli | main | `main` |
| Pmoves-MiniMax-Provider-Verifier | main | `main` |
| PMOVES-agents.md | main | `main` |

## The decision this actually needs

Four of the five were folded in **three days ago** (#2589 model cascade, #2590 agents.md). Whether freshly-adopted upstream trackers belong on the fleet hardened-branch strategy at all is a design question about the fold-in, not a gate-tidying question — and it is the same judgement #2643 deliberately declined to make for `PMOVES-Danger-infra`, leaving it to that submodule's owner.

I am not going to answer it by merging an allowlist entry. Proposed order:

1. Land **#2643** and adopt its policy.
2. Decide, per fork, whether these five get hardened branches created (consistent with #2643) or are declared legitimate upstream trackers.
3. Reopen/rework this PR to whichever the answer is.

Leaving it open rather than closing it so the measurement and the branch listing stay attached to the question. Cross-referenced from #2639.

POWERFULMOVES:
Superseded by #2673 (merged 2026-08-22).

Per the operator call — *"all this hooplah when they could have been just added"* — the five forks got **hardened branches** rather than standing `=main` exemptions. Created at each fork's current HEAD, so divergence is zero and `main` stays byte-identical to upstream.

#2673 covered **six**, not five: `pmoves-keygen` was the mismatch this PR could not reach, and it became resolvable once keygen#1 merged and #2663 landed the gitlink.

Two things this PR could not have known, both found while doing it:

**The branch is declared in four places, not one.** `.gitmodules`, `configs/branch_protection/pmoves_standard.json`, `config/fork_registry.json`, and `mk/preflight.mk`'s `SUBMODULE_BRANCH_ALLOW`. The branch-protection spec **takes precedence over `.gitmodules`** in `resolve_branch()`, so an allowlist-only change would have left the drift audit and any ruleset apply still targeting `main`.

**The allowlist would have been the wrong shape here.** An exemption is right for a fork that genuinely cannot carry an overlay. These five had no PMOVES changes and no reason not to have the branch — so the exemption would have been permanent debt taken on to make a red gate green, which is the failure mode the gate exists to prevent.

Verified on merged main: `submodule-branch-policy-check` **exit 0**, and still exit 2 when one entry is reverted, so the green is not vacuous. Issue #2639 closed.

This PR was also DIRTY and contended `preflight.mk` with the already-merged #2660.

PR #2665, opened 2026-08-21
feat(mcp): gateway federating end to end — 23 tools verified through the secrets pipeline

POWERFULMOVES:
Follow-up to #2656 (merged). The operator sequence ran on B850 and **the previously-unverified gap is closed**. Four more defects surfaced by running it.

## End to end, no shell token

```
chit-manifest-register   +2 entries (MCP_GATEWAY_AUTH_TOKEN, host CHIT key)
bootstrap_env --rotate   minted the token into env.shared
secrets-funnel           env.tier-agent 46 -> 47 entries
up-mcp-gateway           started with `env -u MCP_GATEWAY_AUTH_TOKEN`
mcp-gateway-verify       PASS — 23 tools, exit 0
```

The **first `up` failed, correctly**: the manifest registers a *key*, not a *value*, so `${MCP_GATEWAY_AUTH_TOKEN:?}` refused to start an unauthenticated fleet-wide gateway. That guard doing its job is why the token now comes from the pipeline rather than a shell export.

## Four defects found by running it

**1. `--static` is required to run without the Docker socket.** Removing the socket crash-looped on `guessing network: Cannot connect to the Docker daemon`. `run.go:326` guards `pullAndVerify()` and `guessNetworks()` behind `if !g.Static`, and the latter runs whenever `DOCKER_MCP_IN_CONTAINER=1` — so a non-static containerised gateway dials the socket even when every entry is `remote`. My earlier "remove the socket" change was right for the wrong reason; this is what makes it work.

**2. The verifier could not speak MCP.** It POSTed `tools/list` directly: *"method \"tools/list\" is invalid during session initialization"*. Streamable HTTP requires `initialize` → `notifications/initialized` first, echoing the server's `Mcp-Session-Id`. Implemented properly.

**3. The verify target had no credentials** — 401, exit 3 (unmeasured; correct behaviour, useless result). Now prefixed with `$(LOAD_ENV_SHARED)` like other env-consuming targets.

**4. Docker's own config test is not a gate.** Added `make mcp-gateway-config-check` using the documented `--dry-run`. It correctly reported cipher failing to connect **and exited 0 anyway**. Recorded in the target: it validates configuration, it does not prove federation. `docker mcp tools ls` is also not a substitute — no URL flag, it spawns its own gateway rather than querying the deployed one.

`compose run` *replaces* the service command, so the flags are read back out of the compose file rather than restated in the Makefile — restating them is the drift that made `pmoves-daemon-log-rotation.sh` disagree with `deploy/provision/daemon.json`.

## cipher is disabled, not deleted

It fails the handshake with `initialize: failed to write: 400 Bad Request`, and **cipher is behaving correctly**:

```
GET  /mcp/sse  -> 200, "event: endpoint", "data: /mcp/messages?sessionId=<uuid>"
POST /mcp/sse  -> "Cannot POST /mcp/sse"          (stream endpoint is GET-only)
POST /mcp/messages?sessionId=<uuid> once closed
               -> {"error":"Unknown session"} 400
```

The session is bound to a held-open SSE stream. The gateway uses the official Go SDK's `mcp.SSEClientTransport`, so this needs transport-level debugging, not a catalog edit. The entry stays in `catalog.yaml` commented with that diagnosis, and cipher is removed from `--servers` so **catalog, running set and verifier all agree**. Shipping it enabled would have been a known-broken entry dressed as configuration.

## Also in here

Best-practice review against `docs/security.md` corrected five deviations — healthcheck that could never pass (POSTed to an authenticated endpoint), an unnecessary Docker socket mount, a floating `:latest` tag on a security boundary (now digest-pinned, resolution verified), an under-documented insecure-URL opt-out, and an unexplained permanent warning. Plus two constraints recorded: the localhost-only `Origin` rule, and the digest requirement for future `type: server` entries.

### DIFF
--- a/.claude/PATTERNS.md
+++ b/.claude/PATTERNS.md
@@ -889,3 +889,106 @@ automatic cross-container sharing.
 **Doc-path exhaustiveness:** A path change (e.g. `/sse` → `/mcp/sse`) must be
 grepped across `README.md`, service `CLAUDE.md`, `.claude/CATALOG.md`, handoff
 notes, and `agent_registry.yaml` endpoint fields in the same commit set.
+
+## Blank Is Not Absent — the recurring hazard class (2026-08-21)
+
+**Five independent instances surfaced in a single session.** Not five coincidences —
+one failure mode that this stack reproduces at every layer, because almost every
+"is it configured?" check tests *presence* rather than *usability*.
+
+| # | Where | Empty value | How it presented |
+|---|---|---|---|
+| 1 | `docker-compose.sso.yml` | `SUPABASE_JWT_SECRET:-${JWT_SECRET}`, `SSO_FORWARD_AUTH_SECRET:-` | a **running** sso-auth container with broken auth; edge 401s |
+| 2 | `keygen-cli` | `KEYGEN_PASSPHRASE=""` | an **unencrypted private key** written silently, output said nothing |
+| 3 | `kong.yml` | `DASHBOARD_PASSWORD` | `password: length must be at least 1`, kong crash-loop, 33 restarts |
+| 4 | Docker Desktop | `HTTP_PROXY=` injected into **every** container | `supabase-vector`: "Failed to build Proxy connector: empty string" |
+| 5 | `secrets_sync.py` (#2661 lane) | canonical key present-but-empty | beat a **populated** alias, written as `KEY=` into every target |
+
+### Why it keeps winning
+
+- **`-n` / `if [ -n "$x" ]` / `os.Getenv(k) != ""` all collapse absent and empty into
+  one branch.** The distinction that matters is *three*-valued: absent, empty, set.
+  `os.LookupEnv` / `${VAR?}` / an explicit `is None` check preserve it; the common
+  idioms do not.
+- **Empty passes upstream validation.** Compose `${KEY:?}` rejects empty but
+  `${KEY?}` accepts it. A secrets funnel that has not run exports the name with no
+  value — so the variable *exists*, and every presence check says yes.
+- **The symptom lands far from the cause.** Instance 4 is the clearest: a Windows
+  proxy setting nobody configured broke a log shipper inside a Linux container, via
+  a daemon-level env injection that appears in no compose file and no env file.
+- **The good outcomes are worse than the bad ones.** #1 and #2 did not fail — they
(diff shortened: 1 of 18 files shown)


POWERFULMOVES:
@codex — both P1s confirmed and fixed in `983fb2657`. Both were "works on my node" defects, and you caught them precisely.

**P1 #1 — registry-only addition had no effect elsewhere.** Exactly right. I ran `chit-manifest-register` locally, which wrote both manifests, and **never committed them**:

```
git show origin/main:pmoves/chit/secrets_manifest_v2.yaml | grep -c MCP_GATEWAY_AUTH_TOKEN
0
```

So the token existed in the code-level REGISTRY and in my node's generated manifest, and in neither file on main. Any other node would funnel without it and die on `${MCP_GATEWAY_AUTH_TOKEN:?}`. My end-to-end verification was real but node-local. Both generated manifests are now committed.

**P1 #2 — dependency on state this repo never creates.** Also right, and it exposed a bigger gap: **`up-botz-mcp-bridge` did not exist**. `up-bots` starts `botz-gateway` — the local work-item shim on 8054 — which as you noted is a different service, and *nothing* canonical started the bridge. That's why the dependency was invisible.

Added both the missing target and `mcp-gateway-preflight`, now a prerequisite of `up-mcp-gateway`. Falsification-tested by stopping the bridge:

```
NOT RUNNING: pmoves-botz-mcp-bridge (the only catalogued server)

Fix: make -C pmoves up-botz-mcp-bridge
then: make -C pmoves up-mcp-gateway
exit 1
```

`up-mcp-gateway` now refuses rather than starting a gateway that federates nothing.

Two things the fix shook out, worth recording:

- My first remediation string embedded a raw `docker compose up -d` and the deploy guard blocked the commit as a pipeline bypass. It was right — I reached for raw compose because BoTZ is a separate project, without checking for a canonical target. There wasn't one; now there is, routed through `COMPOSE_ENV_FILES` like every other stack.
- `BOTZ_STACK_DC` first used `-p pmoves` and `up` died on a container name conflict. The running container's labels say `project=pmoves-botz`, `working_dir=<repo>/PMOVES-BoTZ`. Compose identity must match what created the containers or it treats them as foreign.

Verified after: preflight ok, `mcp-gateway-verify` **PASS — 23 tools**.

PR #2698, opened 2026-08-23
docs(acl): the Funnel grant is live — say so where it reads as dead

POWERFULMOVES:
**This PR now does the opposite of what it originally did.** Codex was right; I was wrong.

## What I claimed

That the `tag:exit` Funnel grant was dead, and removed it. My evidence was a sweep showing "every funnel hit in the tree is `secrets-funnel`."

## Why that was wrong

The sweep used `--include=*.yml`, **which does not match `*.yaml`**. Both pieces of contrary evidence live in `.yaml` files, so the search that produced "nothing uses Funnel" never looked where Funnel is used:

| file | says |
|---|---|
| `configs/pinokio-network-inventory.yaml:268-272` | kvm2 (`tailscale_tag: tag:exit`) runs service **`Tailscale-Funnel`**, `class: intentional` — "Tailscale Funnel for public ingress" |
| `configs/tac_trees/public-tunnel.tac.yaml:51-57` | TAC check `l4.kvm2_up.tailscale_funnel` runs `tailscale funnel status`, **expects active listeners**. Cloudflare-Tunnel is the *primary* public ingress; **Funnel is the alternate**, for the tailnet-managed cert chain |

Merging the original change would have broken kvm2's documented Cloudflare-failover ingress.

## What made it look dead

`Traefik + Cloudflare DNS-01` genuinely does serve the seven `*.pmoves.ai` routers — so Funnel isn't what fronts those, and the grant looks orphaned. Then a grep returns a wall of **`secrets-funnel`** (the credential pipeline, an unrelated name), which reads like confirmation rather than noise. Two independent things pointing the wrong way.

## What this PR does now

Keeps the grant. Replaces the one-line comment with the evidence, so the next person hits the citations instead of the same wall.

```diff
-    // Exit nodes can use Tailscale Funnel for public ingress.
+    // Funnel on tag:exit is LIVE, not vestigial. ...
+    //   configs/pinokio-network-inventory.yaml:268-272  (class: intentional)
+    //   configs/tac_trees/public-tunnel.tac.yaml:51-57  (TAC expects active listeners)
+    // Removing this grant breaks kvm2's documented Cloudflare-failover ingress.
```

## Blast radius: none

Net diff vs `main` is **comment-only** — the `nodeAttrs` entry is byte-identical to what's already deployed. `deploy-tailscale-acl.yml` will apply a semantically identical policy.

## Verified

- The workflow's own lint (`strip //` → `json.loads`) parses it
- `nodeAttrs` restored to its original single entry; `tagOwners`/`acls`/`ssh`/`autoApprovers` untouched


POWERFULMOVES:
Verified and confirmed — you're right, and the change is reversed.

Both citations check out. The failure was mechanical: my sweep used `--include=*.yml`, which does not match `*.yaml`. Both `pinokio-network-inventory.yaml` and `public-tunnel.tac.yaml` are `.yaml`, so the search that told me "nothing uses Funnel" never looked where Funnel is used — and what it did return was a tree full of `secrets-funnel`, which reads like confirmation.

The grant stays. The PR now only documents *why* it exists, with your two citations inline, so the next grep lands on the evidence instead of the same wall. Net diff vs main is comment-only.

PR #2701, opened 2026-08-23
docs(auth): the localhost auth URL is right, and here is the trigger that changes it

POWERFULMOVES:
## Why

I flagged `API_EXTERNAL_URL=http://localhost:8000` as a misconfiguration — twice, escalating each time — and was wrong both times. Comment-only PR so the next reader doesn't spend the same hour.

## What I claimed, and what's actually true

| Claim | Reality |
|---|---|
| `issuer: ""` is a live defect in the SSO stack | The OAuth2 server feature is **not enabled** (no `GOTRUE_*OAUTH*` vars). v2.191.0 serves the discovery endpoint anyway, as a skeleton. Nothing consumes it. |
| `localhost` breaks confirmation-email links | `auth.users` is **empty (0 rows)** and `GOTRUE_SMTP_HOST` is **unset**. That path has never run. |
| It should point at the public edge | **There is no public Supabase endpoint.** |

## The load-bearing fact

```
$ docker inspect pmoves-supabase-kong-1 --format '{{json .NetworkSettings.Ports}}'
"8000/tcp": [{"HostIp": "127.0.0.1", "HostPort": "8000"}]
"8001/tcp": [{"HostIp": "127.0.0.1", "HostPort": "8001"}]
```

Loopback, not `0.0.0.0`. And no Traefik router targets Kong — `auth.pmoves.ai` goes to `service=sso-auth@docker` (`docker-compose.traefik.yml:71`), the forward-auth service, not GoTrue.

[Supabase's reverse-proxy guide](https://supabase.com/docs/guides/self-hosting/self-hosted-proxy-https) says `SUPABASE_PUBLIC_URL` / `API_EXTERNAL_URL` must name the URL **end users reach the API at**. Today that is loopback. The value is accurate.

## Why a comment and not a fix

The existing header note settles the *scheme* (base + Kong routes that add `/auth/v1`) and forbids piecemeal changes. It never addressed the *value*, which is why the value kept looking wrong on inspection. This adds that half.

It records the **trigger**, not just the conclusion — the day a Traefik router points at Kong, the value becomes wrong, and `SUPABASE_PUBLIC_URL` + `SUPABASE_SITE_URL` + `GOTRUE_URI_ALLOW_LIST` (empty today, so `SITE_URL` is the only permitted `redirect_to`) + Kong routes + issuer all move together under the existing SET rule.

## Verified

- Compose parses, 106 services
- Comment-only: `19 insertions(+), 0 deletions(-)`


POWERFULMOVES:
Confirmed and fixed. `KONG_PROXY_BIND` is a real knob — `${KONG_PROXY_BIND:-127.0.0.1}` at `docker-compose.yml:866` and `docker-compose.core.yml:562` — so "loopback-only" was a **default** I wrote up as an invariant.

Two changes:
- Reason 1 now says the bind is a default, names `KONG_PROXY_BIND`, and tells the reader to check the live bind rather than trust the line.
- The trigger is now **either** condition: a Traefik router pointing at Kong, **or** any non-loopback bind. Plus the consequence you named — a LAN client handed a GoTrue-generated `localhost:8000` URL resolves it against itself and fails.

That's the sharper failure mode, and it doesn't need Traefik at all. Thanks.

PR #2702, opened 2026-08-23
feat(supabase): the scoped role never stopped a superuser from logging in

POWERFULMOVES:
Step 3.5 of the juicefs cross-node lane — a step the handoff doesn't have, and Step 4 shouldn't ship without.

## The gap

The lane orders its steps around one stated property:

> "The scoped role below is what keeps this from being a superuser surface."

It doesn't. A scoped role changes which credential JuiceFS **uses**. It does nothing to stop a superuser credential from being **accepted**. What governs that is `pg_hba.conf`, which the handoff never mentions:

```
host  all  all  0.0.0.0/0  scram-sha-256      ← every role, from anywhere
```

Once `:5432` is tailnet-published, that admits both superusers as password logins over the tailnet:

| role | rolsuper | rolbypassrls | live connections |
|---|---|---|---|
| `pmoves` | ✅ | ✅ | **35** |
| `supabase_admin` | ✅ | ✅ | 1 |
| `juicefs_meta` | — | — | 1 |

`pmoves` is the initdb superuser (`POSTGRES_USER=pmoves`) and carries nearly all traffic. **Rotating only `supabase_admin`, as the lane specifies, would have left the busier superuser untouched and the surface open.**

## The change

Four rules above the catch-alls — first match wins:

```
host  all  juicefs_meta  100.64.0.0/10        scram-sha-256
host  all  all           100.64.0.0/10        reject
host  all  juicefs_meta  fd7a:115c:a1e0::/48  scram-sha-256
host  all  all           fd7a:115c:a1e0::/48  reject
```

In-cluster traffic is untouched: it arrives on `172.30.x` and still matches the existing `172.16.0.0/12` rule, which sits below these.

## Why an IP-scoped rule is sound here

Checked rather than assumed. Docker's only MASQUERADE rules are `-s <bridge-subnet> ! -o <bridge>` — container→outside. The `DOCKER` chain does pure `DNAT --to-destination` with **no inbound source rewrite**, so a remote peer arrives as its real `100.x`. Had it been SNAT'd to the bridge gateway, tailnet traffic would be indistinguishable from in-cluster traffic and pg_hba could not have been the control at all.

## Why repo-managed instead of edited in place

The image ships `pg_hba.conf` at `/etc/postgresql/` — the container's **writable layer**, not the data volume. An in-container edit survives `restart` and is destroyed by `recreate`. Step 4 of this same lane *requires* a recreate (multi-homing onto `pmoves_external` to publish the port).

**A live edit would have been silently reverted by the very step it exists to protect.**

## Verified

- Diff vs. the running file: **additions only, 0 removals**
- `supabase/postgres:17.6.1.108` loads the file, starts, and matches rules from it (`Connection matched file "/tmp/hba/pg_hba.conf" line 83`) with **no parse errors** — Postgres refuses to start on a malformed hba
- Compose parses; `supabase-db` gains exactly one volume entry

## Activation

Takes effect on the next `supabase-db` recreate. **Step 4 supplies that**, so it costs no extra disruption — and the cross-node confirmation (5090 reaching `:5432` as `juicefs_meta`, and *not* as a superuser) lands naturally as Step 4's acceptance test.


POWERFULMOVES:
This one is stale — it reviewed `869f16ca8`, the branch's first commit. The overlay was regenerated in the follow-up commit `2fa9866d8` after `Validate Compose Files` failed for exactly this reason.

Current branch head:

```
$ git show origin/feat/pg-hba-tailnet-role-scope:pmoves/docker-compose.core.yml | grep -c 'supabase/config/pg_hba.conf'
1
```

Regenerated with the Makefile's pinned toolchain (`uv run --with ruamel.yaml==0.19.1`) so the rendering is byte-stable; the only delta is the `supabase-db` pg_hba mount. `Validate Compose Files` and `validate-composes-ratchet` both pass on the current head.

Your underlying point was right, though — `overlay-up-core` is the path that matters, and the source-only edit would have left the control absent there.

PR #2705, opened 2026-08-24
feat(secrets): juicefs_meta's password had no route to any other node

POWERFULMOVES:
## The gap

B850 mounts `pmoves-media` as the scoped `juicefs_meta` role. Role exists, cutover happened, mount is healthy. The **credential** never entered the pipeline:

```
/home/pmoves/.pmoves-secrets/juicefs_meta_pw  →  bind  →  /run/secrets/jfs_meta_pw
```

```
drwx------ pmoves pmoves  /home/pmoves/.pmoves-secrets/
-rw------- pmoves pmoves  juicefs_meta_pw   41 bytes
```

A hand-placed file, under a *different user's* home, referenced **nowhere** in this repo, with **no CHIT registry entry**. B850 works; no second node can mount without hand-copying a secret.

That's the blocker the 5090 is parked on — they asked for the credential "via the pipeline (never chat)", and there was no pipeline route to ask for.

## Why it's load-bearing, not tidy

[#2702](https://github.com/POWERFULMOVES/PMOVES.AI/pull/2702) made `pg_hba` admit **only** `juicefs_meta` from the tailnet and reject every other role there. A remote mount has no fallback credential — it authenticates as this role or not at all.

## The change

Sanctioned path only — REGISTRY entry → `chit-manifest-register` → `chit-manifest-sync`. **No manifest hand-edited.**

```python
"JUICEFS_META_PASSWORD": {"tier": "data", "required": False},
```

- **`tier: data`** — `juicefs-mount` and `juicefs-gateway` both read `env.tier-data`
- **`required: False`**, unlike its neighbours — only nodes that actually mount `pmoves-media` need it, and under `--merge` (strict) a required slot fails the whole funnel on every node that legitimately lacks it. The usual objection (required=False buys silence) doesn't apply: an absent value fails **loudly** at mount time with an auth error.

v2 also emits `docker_secret: pmoves_juicefs_meta_password` — the exact shape B850's hand-bound secret already has, so the pipeline can materialize what's currently placed by hand.

## Necessary but not sufficient

In #2700's sense. This routes the label; it does not conjure a value. Two follow-ons, tracked separately:

1. The value has to be put into CHIT (operator).
2. `docker-compose.juicefs.yml:48` and `:87` still default `JUICEFS_META_URL` to **`supabase_admin`** with an inline password — so a compose-driven recreate of the mount would silently revert the cutover to a superuser.

## Verified

| gate | result |
|---|---|
| `chit-manifest-check` | in sync |
| `chit-manifest-register --check` | complete, no constraint drift |
| re-run `chit-manifest-sync` | byte-stable (diff unchanged) |
| manifest diffs | **+23, 0 deletions** — one entry each |

### DIFF
--- a/.github/workflows/sync-secrets-local.yml
+++ b/.github/workflows/sync-secrets-local.yml
@@ -125,6 +125,13 @@ jobs:
           SUPABASE_REALTIME_KEY: ${{ secrets.SUPABASE_REALTIME_KEY }}
           SUPABASE_REALTIME_SECRET: ${{ secrets.SUPABASE_REALTIME_SECRET }}
           # ── Data Stores ──
+          # JuiceFS scoped metadata role. Same trap as CLAUDE_CODE_OAUTH_TOKEN
+          # above: the manifest slot alone routes nothing to a node hydrated from
+          # the bundle, because this map IS the bundle. The nodes that need this
+          # are precisely the runnerless ones -- 5090 has no Actions runner and
+          # hydrates via `secrets-funnel-from-prod` -- so omitting it here would
+          # leave the route terminating one hop short of its only consumer.
+          JUICEFS_META_PASSWORD: ${{ secrets.JUICEFS_META_PASSWORD }}
           MEILI_MASTER_KEY: ${{ secrets.MEILI_MASTER_KEY }}
           SURREAL_ADDRESS: ${{ secrets.SURREAL_ADDRESS }}
           SURREAL_DATABASE: ${{ secrets.SURREAL_DATABASE }}


POWERFULMOVES:
Verified and fixed — you're right, and it's the P1 that mattered.

Confirmed `sync-secrets-local.yml:115-119` says exactly this, in a comment #2700 added when it hit the same trap:

> the manifest slot alone does not reach these nodes: this workflow builds the CGP bundle solely from this map, and the v1 sync discards `github_secret` targets, so a registered label with no row here funnels nothing on any node hydrated from the bundle.

I read #2700 and still shipped the half that reaches nobody. Worse here than there, because the bundle-hydrated nodes are exactly this slot's only consumers — **the 5090 has no Actions runner** and hydrates via `secrets-funnel-from-prod`. The operator would have created the Prod secret, the funnel would have reported success, and `env.tier-data` would still have been empty.

Added `JUICEFS_META_PASSWORD` to the Data Stores block with the reasoning inline.

PR #2706, opened 2026-08-24
fix(secrets): the encoder ran before the sync, so rotation republished the old password

POWERFULMOVES:
Found by rotating `SUPABASE_DB_PASSWORD` for real on B850 tonight — not by reading code.

## What happened

The funnel reported success. The value it owns *was* genuinely rotated. Then `postgrest`, `gotrue` and `storage` came back `28P01 password authentication failed for user "pmoves"` — **while holding a correct `POSTGRES_PASSWORD` in their own environment.**

Fingerprints at the moment of failure (sha prefix of the value, never the value):

```
POSTGRES_PASSWORD        len64/3a4e71   ← rotated, correct
PGRST_DB_URI       (dsn) len48/ddb76e   ← stale
GOTRUE_DB_DATABASE_URL   len48/ddb76e   ← stale
DATABASE_URL       (dsn) len48/ddb76e   ← stale
```

## Cause

Six compose DSNs resolve as:

```yaml
${POSTGRES_PASSWORD_URLENCODED:-${POSTGRES_PASSWORD:-...}}
```

`POSTGRES_PASSWORD_URLENCODED` **wins that chain** and has **zero references in the CHIT manifest** — nothing rotates it. `48` is the old `secrets-rotate` default length, so the stale value dates from the *previous* rotation. Every rotation since has been a silent partial.

## Fix

Cleared, not re-derived. `bootstrap_env.py` mints URL-safe values, so for a generated password the encoding is a no-op — verified `quote(pw) == pw`. `--clear` records the name in `secrets_cleared.yaml`, which is what stops `secrets-local-hydrate` writing the stale value straight back; deleting the line by hand is not enough.

The `:-` operator then falls through correctly on the empty value. That detail is load-bearing: plain `-` would have passed the **empty string** into the DSN instead.

`secrets_cleared.yaml` is committed so other nodes inherit the clear rather than each rediscovering this the same way. It holds key **names** only.

## The wider class — 11 cases, not fixed here

An unmanaged variable that outranks a managed one is worse than an unmanaged variable alone: it turns every future rotation into a silent partial. Sweeping compose for `${A:-${B:-...}}` where `B` is manifest-managed and `A` is not:

| shadowing var | shadows | files |
|---|---|---|
| `POSTGRES_PASSWORD_URLENCODED` | `POSTGRES_PASSWORD` | 2 ← **this PR** |
| `ANON_KEY` | `SUPABASE_ANON_KEY` | 2 |
| `SERVICE_ROLE_KEY` | `SUPABASE_SERVICE_ROLE_KEY` | 3 |
| `JWT_SECRET` | `SUPABASE_JWT_SECRET` | 2 |
| `JUICEFS_S3_PASSWORD` | `MINIO_PASSWORD` | 2 |
| `JUICEFS_S3_USER` | `MINIO_USER` | 2 |
| `RENDER_WEBHOOK_SUPABASE_ANON_KEY` | `SUPABASE_ANON_KEY` | 2 |
| `RENDER_WEBHOOK_SUPABASE_SERVICE_ROLE_KEY` | `SUPABASE_SERVICE_ROLE_KEY` | 2 |
| `SUPABASE_INTERNAL_URL` | `SUPABASE_URL` | 2 |
| `SUPA_REST_INTERNAL_URL` | `SUPA_REST_URL` | 3 |
| `TENSORZERO_CLICKHOUSE_GATEWAY_URL` | `TENSORZERO_CLICKHOUSE_URL` | 2 |

Several are legitimate (the `*_URL` ones are endpoint overrides, not credentials). The credential-bearing rows deserve a look. A gate for the class is worth building — following `validate_runner_labels.py`'s "undeclared is a failure, declared is a decision" shape — but each case needs judgement, so it isn't in this PR.

## Why the comment sits at the DSN

The failure is invisible from the funnel's side. It is only observable as three services rejecting a password their own environment says is correct.

## Verified

- Compose parses, 106 services; split overlays regenerated with the pinned toolchain
- On B850: **13/13 supabase services healthy**, 31 sessions re-established after the clear + recreate
- `juicefs_meta` unaffected throughout (separate credential)

### DIFF
--- a/pmoves/docker-compose.core.yml
+++ b/pmoves/docker-compose.core.yml
@@ -189,15 +189,18 @@ services:
     # Database connection
     - GOTRUE_DB_DRIVER=postgres
     # PRECEDENCE TRAP (bit us 2026-08-23, during a real rotation).
-    # POSTGRES_PASSWORD_URLENCODED WINS over POSTGRES_PASSWORD here, and it is
-    # NOT a managed secret -- zero references in the CHIT manifest. So the funnel
-    # rotates POSTGRES_PASSWORD, reports success, and every DSN built from this
-    # chain keeps the OLD password. Three services (postgrest, gotrue, storage)
-    # came back 28P01 while holding a correct POSTGRES_PASSWORD in their env.
-    # It is cleared in secrets_cleared.yaml so hydration cannot restore it; the
-    # `:-` operator then falls through correctly on the empty value.
-    # Only set it again for an operator-supplied password containing characters
-    # that need URL-encoding -- and if you do, re-set it on EVERY rotation.
+    # POSTGRES_PASSWORD_URLENCODED wins over POSTGRES_PASSWORD here, and it is
+    # NOT sourced from the CHIT manifest -- it is emitted by
+    # tools/credential_urlencoder.py into env.tier-supabase.urlencoded, which
+    # compose loads LAST of all env files. So this chain resolves to whatever
+    # that encoder last wrote, overriding every other source.
+    #
+    # The encoder merges env.shared then env.tier-supabase (later wins), so it
+    # MUST run after secrets-funnel-sync. It used to run before, which meant a
+    # rotation re-published the PREVIOUS password as the winning value: the
+    # funnel reported success and postgrest, gotrue and storage came back 28P01
+    # holding a correct POSTGRES_PASSWORD in their own env. Order fixed in
+    # mk/codex.mk; see the note above the secrets-funnel target.
     - GOTRUE_DB_DATABASE_URL=postgres://${POSTGRES_USER:-${SUPABASE_DB_USER:-postgres}}:${POSTGRES_PASSWORD_URLENCODED:-${POSTGRES_PASSWORD:-${SUPABASE_DB_PASSWORD:-postgres}}}@supabase-db:5432/${POSTGRES_DB:-${SUPABASE_DB_NAME:-postgres}}?sslmode=disable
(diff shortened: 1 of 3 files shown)


POWERFULMOVES:
Verified — you're right, and it means my first commit fixed a symptom and misattributed the cause. Corrected in `113eace98`.

Confirmed the ordering at `mk/codex.mk:146-151`: `credential_urlencoder.py` ran **before** `secrets-funnel-sync`. The encoder merges `env.shared` then `env.tier-supabase` (later wins) and writes `env.tier-supabase.urlencoded`, which `Makefile:132-134` appends **last** to `COMPOSE_ENV_FILES` — so it overrides every other source. It also emits the `_URLENCODED` key unconditionally by design (`credential_urlencoder.py:70-72`, "even if safe … so compose fallback always resolves"), which is why the `secrets_cleared.yaml` tombstone can't suppress it.

The tell I walked past: **a second funnel run healed B850.** That's the signature of exactly this ordering — by the second run the tier files were current, so the encoder had correct inputs. I attributed the recovery to the clear, which did nothing for this path.

Fix is the one-line reorder: sync first, encode second. `make -n secrets-funnel` now shows `secrets-funnel-sync` preceding `credential_urlencoder.py`. The comment sits above the target rather than inside the recipe — a `#` line in a recipe is echoed to the shell and `${...}` in it would be expanded by make.

`secrets_cleared.yaml` stays, but reframed as hygiene: `env.shared`'s own copy is a dead second source now that the encoder owns the value. It is not the fix, and the PR body said otherwise.

PR #2708, opened 2026-08-24
fix(juicefs): forward META_ROLE through the cross-node make target (+ JUICEFS_META_PASSWORD fallback)

POWERFULMOVES:
## The gap
#2683 added `META_ROLE` to `juicefs-cross-node-setup.sh`, but the make target (`egress.mk:319`) never forwarded it. So the canonical step-5 path — `make juicefs-cross-node-setup META_ROLE=juicefs_meta` — **silently fell back to `supabase_admin`**, defeating the scoped-role cutover on remote nodes (the 5090 mount). Classic 'the fix exists but isn't wired end-to-end' — the same shape B850 surfaced across tonight's findings.

## Fix
- `egress.mk`: forward `META_ROLE` (default `supabase_admin`, back-compat). `DB_PASS` now falls back to the funnel-delivered `JUICEFS_META_PASSWORD` (registered in `chit_manifest_register.py`, #2705), so a node that received the secret via the pipeline just runs `make juicefs-cross-node-setup META_ROLE=juicefs_meta`.
- `juicefs-cross-node-setup.sh`: matching `DB_PASS:-JUICEFS_META_PASSWORD` fallback for direct invocation; refreshed the error message.

## Secrecy
Both `DB_PASS` and `JUICEFS_META_PASSWORD` pass as the sub-process **environment** (not argv) and are handed to JuiceFS via `META_PASSWORD` — never in `ps`.

## Why now
Unblocks 5090 step 5 (mount `pmoves-media`) once B850 finishes gate 4 (expose `:5432`). B850-local mount uses `juicefs-mount-local`, so this remote-path fix doesn't affect the completed B850 cutover.

### DIFF
--- a/pmoves/mk/egress.mk
+++ b/pmoves/mk/egress.mk
@@ -322,12 +322,13 @@ juicefs-cross-node-setup: ## Mount JuiceFS on this node (run on remote): make ju
 	@# REJECTS supabase_admin from the tailnet (#2702), so that default fails.
 	@#
 	@# Password precedence: explicit DB_PASS, else the funnel-delivered
-	@# JUICEFS_META_PASSWORD read from env.tier-data AT RECIPE TIME. It has to be the
-	@# shell form $$(grep ...) and NOT $$(JUICEFS_META_PASSWORD): make populates its
+	@# JUICEFS_META_PASSWORD, resolved AT RECIPE TIME through scripts/with-env.sh —
+	@# the canonical loader (env.shared* -> tier files -> .env* overlays, mirroring
+	@# compose layering). It cannot be $$(JUICEFS_META_PASSWORD): make populates its
 	@# variables only from the environment and Makefiles, and nothing includes the
-	@# generated tier files — so a make-variable reference is empty exactly on the
-	@# nodes the funnel just delivered to. Same shell-read pattern as line ~229 in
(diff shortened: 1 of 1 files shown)


POWERFULMOVES:
Two things: the comment is stale, and the underlying point was still right.

**Stale:** it reviewed `dda6c454` — the branch's first commit. `8b7cecde0` had already replaced `$(JUICEFS_META_PASSWORD)` with a recipe-time shell read and removed the `$(error)`, so the abort you describe no longer happened.

**But the pointer to `scripts/with-env.sh` was the better answer,** and I've now taken it. My fix read `env.tier-data` with an inline grep — correct, but it hard-codes which tier file the value lands in and isn't the house pattern. `with-env.sh` loads `env.shared* → tier files → .env* overlays`, mirroring compose layering, so precedence is honoured. Same idiom as `mk/yt-cookies.mk:18`, and `infra.mk:603` already carries a comment recording this as a prior Codex P1 — so this is the second time the lesson has landed.

It also removes the `cut -d= -f2-` parsing, so there's no embedded-`=` edge case left to get right.

Verified with a stubbed script — tier-delivered: `DB_PASS_len=27`; explicit `DB_PASS`: wins; neither: the script's own error, no make abort.

PR #2709, opened 2026-08-24
docs(ops): the metadata credential has an owner and a vendor-sourced procedure

POWERFULMOVES:
"Someone creates the GitHub Prod secret" isn't an assignment. Provisioning the JuiceFS metadata credential gates **every node's** ability to mount `pmoves-media`, so it belongs to **Z890 (Infrastructure Coordinator)** — not to whichever node notices it's missing. B850 holds the metadata DB and shouldn't become the ambient owner of a secret it merely consumes.

Written against [JuiceFS community docs][pg-bp] rather than our own habits. The comparison found four things:

| | finding |
|---|---|
| ✅ **ALIGNED** | Dedicated user + network-scoped `pg_hba` is the **vendor's own recommended shape**. [#2702](https://github.com/POWERFULMOVES/PMOVES.AI/pull/2702) landed it scoped to the tailnet. Worth recording that this was vendor practice, not local invention — and that the doc pairs the role *with* the hba rule, which is the pairing the original lane handoff was missing. |
| ⚠️ **PARTIAL** | `META_PASSWORD` is documented and we use it. **`META_PASSWORD_FILE` is also documented**, and B850 already bind-mounts the secret as a file before reading it back through a shell. Adopt on next recreate. |
| ⚠️ **DIVERGENT** | `sslmode=disable`. Unremarkable container-to-container; **not** unremarkable once `:5432` is on the tailnet. WireGuard covers the transport, but the session has no TLS of its own. Flagged as a decision to make **at** Step 4 rather than inherit. |
| ❌ **GAP** | The vendor asks for restore **tests**, not just backups: *"do some tests to restore the data in an experimental environment to confirm that the backup is valid."* B850 dumps metadata hourly to MinIO and has **never restored one**. An untested backup is a hypothesis. Z890 task, tracked separately. |

Plus a constraint worth writing down before someone proposes the obvious fix:

> "PostgreSQL does not yet support Multi-Shard (Distributed) transactions, do not use a multi-server distributed architecture for the JuiceFS metadata."

B850 is an SPOF for metadata **by vendor design**. Availability work belongs in restore, not replication.

## Also corrects the mount runbook

It still told operators `DB_PASS` was *"the Supabase DB password"*. That predates the scoped-role cutover and is now **actively wrong** — `pg_hba` rejects `supabase_admin` from the tailnet, so following it produces an auth error that reads like a bad secret rather than a rejected role. It now passes `META_ROLE=juicefs_meta` explicitly, since the script still defaults to `supabase_admin` for back-compat.

## Note on the vendor quote

Their `pg_hba` example carries a private LAN CIDR, so it's paraphrased rather than quoted. The no-LAN-IPs guard blocked my first draft and was right to — the rule holds even when the address is someone else's documentation.

[pg-bp]: https://juicefs.com/docs/community/postgresql_best_practices/

### DIFF
--- a/pmoves/docs/operations/JUICEFS_META_CREDENTIAL_RUNBOOK.md
+++ b/pmoves/docs/operations/JUICEFS_META_CREDENTIAL_RUNBOOK.md
@@ -151,24 +151,61 @@ funnel silently skips the entry.
 `juicefs_meta` is NOT a superuser, and `bootstrap_db.sh` does not manage it, so
 this affects only the mounts — not the Supabase stack.
 
+The password must not reach argv or shell history. The obvious
+`psql -c "ALTER ROLE … PASSWORD '<new>'"` form violates both, and so does
+`psql -v` — the value lands in the container's argv either way.
+
+Pipe the statement on **stdin**:
+
 ```sh
-# On B850:
-docker exec -i pmoves-supabase-db-1 \
-  psql -h 127.0.0.1 -U supabase_admin -d postgres \
-  -c "ALTER ROLE juicefs_meta PASSWORD '<new>';"
+# On B850. `read -rs` does not echo and does not enter history; printf is a
+# shell builtin, so the value never becomes a separate process's argv.
+read -rs NEWPW
+printf "ALTER ROLE juicefs_meta PASSWORD '%s';\n" "$NEWPW" \
+  | docker exec -i pmoves-supabase-db-1 \
+      psql -h 127.0.0.1 -U supabase_admin -d postgres -v ON_ERROR_STOP=1
+unset NEWPW
 ```
+
+Mint the value with the pipeline's own generator (URL-safe, no quote escaping to
+get wrong) rather than choosing one by hand.
+
+**This belongs behind a Make target** that reads from stdin, alongside
+`secrets-rotate`. Until that exists, use the form above — and afterwards confirm
+no `ALTER ROLE` line reached your history file.
 
 Then update the secret file B850's mount reads, recreate `juicefs-mount`, and set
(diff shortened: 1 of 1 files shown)


POWERFULMOVES:
All three verified; two were doc defects and one is a merge-order dependency.

**1. Rotation contradicted itself — fixed.** Section 2.1 says never paste the value on a CLI, then 2.2-alt handed over a `psql -c "ALTER ROLE … PASSWORD '<new>'"`. You're right on both counts, and `psql -v` is no better — same argv. Replaced with a stdin form (`read -rs` → shell builtin `printf` → `docker exec -i psql`), so the value never becomes another process's argv. Validated on a throwaway role against the live DB: `ALTER ROLE` applied, role authenticated with the piped value over a real scram path, `juicefs_meta` untouched. Also recorded where it belongs long-term — behind a Make target reading stdin, next to `secrets-rotate`.

**2. Delivery order — fixed, and this was the dangerous one.** Confirmed `pull_chit_bundle.sh:45-46` takes the newest *already-successful* run:

```
gh run list --workflow "$WORKFLOW" --status success --limit 1
```

which can predate the Prod secret. Pull succeeds, funnel reports success, key absent or stale — on the 5090, the exact runnerless node this slot exists for. The runbook now requires a producer run first, an explicit wait, and a `createdAt` comparison against when the secret was set. That timestamp check is the whole guarantee.

**3. Mount snippet — merge-order dependency, not a doc defect.** `META_ROLE=juicefs_meta` with no `DB_PASS` needs the target change in #2708, which now resolves the funnel password through `scripts/with-env.sh`. **Merging #2708 first**, after which this snippet is accurate as written.

PR #2716, opened 2026-08-24
feat(launcher): default to a node steward that claims work, not an agent that edits

POWERFULMOVES:
`claude-pmoves` defaulted to `delivery-agent` — so every node session began as an **execution body with no node context and no claim discipline**.

A B850 session on 2026-08-23 ran exactly that way to completion: **eight PRs and three live database mutations on the data-tier host, none of it claimed**, discovered only when the operator asked why the register showed nobody there.

That's structural, not personal. **An agent that starts holding `Edit` will edit.**

## What changes

`node-steward` is denied `Write` and `Edit`. It holds node context, files the CLAIM, and spawns delivery agents to execute — arranging the three bodies rather than being all three, which is the Village Rule failure that session demonstrated.

It also carries what this node has already paid for:

| trap | |
|---|---|
| Makefile→`bash script.sh` never consults the exec bit | `make` works while the PATH command is dead |
| `$(VAR)` can't see funnel-delivered values | make reads env + Makefiles only; use `with-env.sh` |
| SCRAM hashes are salted per `ALTER` | distinct hashes ≠ distinct passwords |
| an unmanaged var outranking a managed one | every rotation becomes a silent partial |
| `--include=*.yml` ≠ `*.yaml` | a confident wrong answer |

And the through-line it exists to catch — every significant finding in that lane was **one shape**: a fix that exists only as node-local state. The agent asks where a fix now lives, and whether a clone, a recreate, or a second node would still have it.

## Memory belongs in Cipher, and that's blocked on #2710

The trap table is a **seed, not the store**. A static list goes stale and can't record what a session learns.

Honest caveat, written into the agent: **it has never been stored in Cipher**, because the session that learned it had no route. Bare `claude` carries no MCP roster, so no `mcp__pmoves-cipher__*`, no agent-zero, no supabase, no tailscale — and nothing announces the absence.

That makes **[#2710](https://github.com/POWERFULMOVES/PMOVES.AI/pull/2710) the prerequisite for the whole memory story**, and the agent instructs the first Cipher-capable steward to seed, then say so if the tools are missing rather than proceeding as if working without memory were normal.

*(Verified the constraint: the running service is `cipher-pmoves-shim` v0.1.0 exposing only `/health` and `/mcp/sse`. No HTTP CRUD; `/mcp/messages` 400s without a session. The MCP client is the only write path.)*

## Latent bug the tests exposed

`AGENT="${1:-...}"` consumed **any** first argument — `claude-pmoves --print ping` launched with `--agent --print`. A leading `-` now means "no agent named".

## Verified — five paths

```
default                 -> --agent node-steward
explicit agent          -> --agent delivery-agent
PMOVES_DEFAULT_AGENT    -> --agent control-agent
steward file absent     -> --agent delivery-agent   (older checkout still works)
no args at all          -> --agent node-steward
```

Claimed in `AGNOTE4482PHI.t1.md` **before** the edits this time — see #2711.


POWERFULMOVES:
All three verified and fixed. The first one made the commit cosmetic, so thank you for it.

**1. The default reached neither public entry point.** `Makefile:5091` passed `delivery-agent` explicitly and the Windows shim hard-coded it — `DEFAULT_AGENT` was never consulted on the documented Make path or on any Windows install.

Worth noting the Windows fix isn't "drop the override": that shim routes through `deploy\provision\claude-pmoves.cmd`, which has **no default-agent logic of its own** — the fallback lives in the bash shorthand Windows never executes. Dropping it would have passed *no agent at all*. It now names the steward explicitly, with a comment on why the value is duplicated and that it must stay in step.

**2. The steward couldn't file its own claim** — a genuine self-contradiction I built. Denied Write/Edit, but the CLAIM *is* an edit. Bash contradicts the contract and bypasses the claim-collision hook (which only watches Write/Edit); delegating reverses claim-then-delegate. And `memory-agent` has no Write either, so that escape doesn't exist.

Resolved via a sanctioned path that already existed: **`pmoves-chit-sign`** appends a CLAIM-or-RELEASE entry *and* signs the trail. That's now the stated reason the role has `Skill` but not `Write`/`Edit` — the claim goes through a signed path, which is strictly better than what it would have done with `Edit`.

**3. `node_id` identifies almost nothing** — 1 of 15 profiles defines it, and there it's `pmoves-b850` against hostname `PMOVES-B850-AI-TOP`. Switched to top-level `id:`/`name:`, which every profile carries.

Also: the anchors ratchet caught my placeholder `make -C pmoves foo` as a GHOST_TARGET. Replaced with `make -C pmoves claude-pmoves` — the truer example, since that target worked for a month while the PATH command was dead.

Verified: ratchet PASS (499 baselined, 0 new); Make path → `--agent node-steward`; explicit agent still wins; Windows shim keeps CRLF and selects the steward on both paths.

PR #2719, opened 2026-08-24
feat(secrets): rotate a DB role without the plaintext ever reaching the server

POWERFULMOVES:
`JUICEFS_META_CREDENTIAL_RUNBOOK.md` §2.2-alt says, in as many words:

> *"This belongs behind a Make target that reads from stdin, alongside `secrets-rotate`. Until that exists, use the form above."*

This is that target — and it goes one step further than the interim form, for a reason PostgreSQL documents itself.

## What the interim form solves, and what it doesn't

The runbook pipes `ALTER ROLE x PASSWORD '<plaintext>'` on **stdin** rather than via `psql -c` or `-v`. That correctly keeps the value out of **argv** and **shell history**. It does not address the rest:

> *"The password will be **transmitted to the server in cleartext**, and it might also be **logged in … the server log**."*
> — [ALTER ROLE](https://www.postgresql.org/docs/17/sql-alterrole.html)

`psql`'s own `\password` exists for exactly this, and *"encrypts it"* **client-side**. It's interactive and can't be scripted — but the property that makes it safe is available to any client:

> *"If the presented password string is already in MD5-encrypted or SCRAM-encrypted format, then it is **stored as-is**."*
> — [CREATE ROLE](https://www.postgresql.org/docs/17/sql-createrole.html)

So this computes the **SCRAM-SHA-256 verifier locally and sends that**. The plaintext never leaves the minting process; what crosses the socket is already the same bytes `pg_authid` would store.

## Proven, not asserted

Round-tripped against a real server — a throwaway `postgres:16-alpine` hashed a known password itself, its salt and iteration count were read back out of `pg_authid`, and the same inputs were fed to this code:

```
server: SCRAM-SHA-256$4096:dCz9vP/dwTpMskwuGii5WQ==$Ku ...
mine  : SCRAM-SHA-256$4096:dCz9vP/dwTpMskwuGii5WQ==$Ku ...
EXACT MATCH: True
```

Hand-rolled crypto that is merely *plausible* is worse than none — it fails closed only by luck.

## Correction to this PR's original text

The first version of this PR — and the target's own closing output — claimed:

> *"Nothing in this repo writes to GitHub secrets, deliberately: that direction is the trust boundary."*

**That was invented reasoning presented as a design principle, and it is wrong.** `pmoves/tools/push-gh-secrets.sh` does exactly that, is wired into the Makefile, and **gates on the CHIT secrets manifest** — a key must be a registered CHIT slot before it can be pushed. That gate is why registering the slot mattered; I had it in front of me and described it as an absence.

The corrected guidance names the real command:

```
./pmoves/tools/push-gh-secrets.sh --env Prod --only JUICEFS_META_PASSWORD --dry-run
./pmoves/tools/push-gh-secrets.sh --env Prod --only JUICEFS_META_PASSWORD
```

And names the tool **not** to use: `credential_setup.py` writes `/repos/OWNER/REPO/actions/secrets` — **repo-level, no environment support** — so it cannot target Prod. Checked, not assumed: of the three credential tools, only `push-gh-secrets.sh` accepts `--env`.

## Orchestration is in a script, not the recipe

The first attempt put it in the Make recipe. Three quoting layers — Python generating make generating shell — ate the line continuations and produced `missing separator`. A secret that must never become argv is the wrong thing to route through three layers of quoting.

The recipe calls it with `bash scripts/…` rather than relying on the exec bit — a missing exec bit is precisely the node-local runtime state B850's JuiceFS lane listed among its recurring findings.

## Refusals

| case | behaviour |
|---|---|
| `ROLE` without `KEY` | **refuses** — rotating the DB alone leaves consumers holding a password the server no longer accepts, surfacing later as an auth error nobody connects back to this run |
| empty value returned | **refuses to funnel** — a blank gives every consumer a correct-looking config that cannot authenticate |
| non-ASCII password | **not possible** — the tool *mints* URL-safe base64, making RFC 4013 SASLprep a provable no-op. It would **not** be for an operator-supplied value |

Composes with `secrets-rotate` rather than restating it: the plaintext is handed over in `PMOVES_ROTATE_VALUE`, which that target already reads from the environment for this exact reason.

## Verified

Both guards fire; `DRY_RUN=1` touches no database:

```
dry-run: would ALTER ROLE juicefs_meta with a SCRAM-SHA-256 verifier
         (4096 iterations, 16-byte salt); plaintext length 64
```

### DIFF
--- a/pmoves/scripts/rotate_db_role.sh
+++ b/pmoves/scripts/rotate_db_role.sh
@@ -56,6 +56,26 @@ if [ -n "$DRY" ]; then
   exec $PY "$TOOL" "${args[@]}" --dry-run
 fi
 
+# Pre-flight everything that can fail WITHOUT touching the database, so the
+# ALTER is never the step that discovers a broken local prerequisite: after it,
+# the server holds a credential no consumer has, and a failed funnel means a
+# privileged recovery rotation.
+# (a) env.shared must exist — secrets-rotate surgically edits it.
+# (b) KEY must already be a line in it — a typo would otherwise surface as a
+#     funnel failure one step past the point of no return.
+# (c) the tool's own --dry-run validates role name, container reachability,
+#     and argument shape against the real docker exec path.
+if [ ! -f "$HERE/env.shared" ]; then
+  echo "pre-flight: $HERE/env.shared missing — secrets-rotate would fail after the ALTER" >&2
+  exit 1
+fi
+if ! grep -qE "^${KEY}=" "$HERE/env.shared"; then
+  # Not fatal: the funnel may legitimately introduce a brand-new key. But say
+  # so loudly BEFORE the database moves, where it is still cheap to abort.
+  echo "pre-flight WARNING: '$KEY' is not yet a key in env.shared — secrets-rotate will CREATE it" >&2
+fi
+$PY "$TOOL" "${args[@]}" --dry-run >/dev/null
+
 # One capture. Report every line EXCEPT the value-bearing one.
 out="$($PY "$TOOL" "${args[@]}" --emit-to-env PMOVES_ROTATE_VALUE)"
 printf '%s\n' "$out" | grep -v '^PMOVES_ROTATE_VALUE=' || true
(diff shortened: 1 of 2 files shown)


POWERFULMOVES:
CodeQL was the one non-CI blocker here: `py/clear-text-logging-sensitive-data` on `rotate_db_role_password.py:163` — the `--emit-to-env` print. True positive, intentional design (the PR body documents it as the *only* plaintext exit path, consumed on a pipe). Pushed `e411dfd5e`: inline suppression with the rationale on the line itself, so the single exit point stays visible instead of the value being routed through a less-auditable sink. claude-review/triage reruns are also in flight (upstream init-crash is non-deterministic; durable fix is the pin bump in #2746).

POWERFULMOVES:
Both threads addressed in `3ac610028`. **P1 (LOGIN):** the ALTER now carries `WITH LOGIN` — you were right that `00_3_juicefs_meta_role.sql` provisions the role NOLOGIN and a password-only ALTER reports success while auth stays rejected; WITH LOGIN is idempotent for roles that can already log in, so the general-tool case is unaffected. **P2 (funnel-after-ALTER):** two changes. (a) Pre-flight before the database moves: env.shared existence is required, a not-yet-existing KEY gets a loud pre-ALTER warning (not fatal — the funnel may legitimately create it), and the tool's `--dry-run` runs against the real docker exec path to validate role/container/args. (b) Failure preservation: a funnel failure is now caught and re-emits the minted value with direct retry instructions (`PMOVES_ROTATE_VALUE=... make secrets-rotate KEY=...`) instead of `set -e` discarding the only surviving copy and forcing a second privileged rotation. Script dry-run verified end-to-end on this node.

PR #2721, opened 2026-08-24
fix(smoke): the PostgREST check was asserting that PostgREST is insecure

POWERFULMOVES:
```bash
test_http_endpoint "Supabase PostgREST" "http://localhost:3010/" "200"
```

A bare **unauthenticated** GET expecting **200**. PostgREST runs with `PGRST_JWT_SECRET` and no anon role, so an unauthenticated request is *supposed* to be refused.

That assertion could only ever go green **while PostgREST was open to anonymous reads.** It wasn't testing that the service works — it was testing that the service is unprotected, and reporting the protected state as a failure.

It began "failing" on B850 the moment that node revoked `PUBLIC` and `anon` grants (#2707) as part of the JuiceFS hardening lane. **The test failed because the system got safer.**

Now asserts **401** — and a 200 there would be the finding.

## The assertion change alone would not have worked

The helper called `curl -sf`. `-f` turns any 4xx into **exit 22**, and the pass condition is:

```bash
if [ $status -eq 0 ] && [ "$http_code" = "$expected_status" ]; then
```

So a 401 was rejected on the **exit code** before `http_code` was ever compared. The helper could not assert *any* non-2xx status. Dropping `-f` from the two helper calls is what makes `expected_status` mean anything outside 2xx.

Verified against a live 401, both forms:

| form | result |
|---|---|
| `curl -sf …` | `http_code=401 exit=22` — fails **even when expecting 401** |
| `curl -s …` | `http_code=401 exit=0` — passes |

Transport failures are unaffected: exit **7** (connection refused) and **28** (timeout) are curl-level, not HTTP-level, and both branches still fire. Tests expecting 200 are unaffected — the `http_code` comparison was always the real assertion; `-f` was only masking it.

## Deliberately not changed: Archon

`test_http_endpoint "Archon" ".../healthz" "200"` also returns 401 on B850. **That one is left failing on purpose.**

A health endpoint requiring authentication is not a defensible posture the way a refusing PostgREST is. That red is a real finding and should stay visible rather than be normalised into an expectation — which is precisely the mistake this PR is undoing for PostgREST.

## Context

This is the check that has been the sole blocker reported on the agent-zero bumps. It's also the only smoke job that *executes*: on hosted runners the job runs three steps and reports success without testing anything (it gates on a label), so `upstream-update`-labelled PRs are the only ones where these assertions ever run.

### DIFF
--- a/pmoves/scripts/smoke-tests.sh
+++ b/pmoves/scripts/smoke-tests.sh
@@ -134,7 +134,11 @@ test_http_endpoint() {
     # could only ever assert a 2xx status -- $status -eq 0 rejected a 401 before
     # http_code was ever compared. The http_code comparison is the real assertion;
     # transport failures still surface as exit 7 (refused) and 28 (timeout).
-    if [ $status -eq 0 ] && [ "$http_code" = "$expected_status" ]; then
+    # expected_status may be an alternation ("200|401") for checks whose correct
+    # answer is deployment-dependent. NOTE: =~ with the pattern from a variable,
+    # not case: bash parses case-alternation before expansion, so a "|" arriving
+    # via $expected_status would be a literal pipe, never an alternation.
+    if [ $status -eq 0 ] && [[ "$http_code" =~ ^(${expected_status})$ ]]; then
         print_pass "$name"
         [ "$VERBOSE" = true ] && echo "       Response: $body"
         return 0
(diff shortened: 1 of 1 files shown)


POWERFULMOVES:
P1 addressed in `e1e772bff`: the check now accepts **200|401** — the root's unauthenticated status is deployment-dependent and both are healthy (fresh bootstrap: `PGRST_ANON_ROLE=anon` serves the OpenAPI doc as 200, JWT secret only validates supplied tokens; anon-revoked nodes like B850 answer 401). You were right that the hardcoded 401 rejected the standard deployment and contradicted `test_critical_path.py` — the smoke asserts liveness, security posture stays with auth-alignment/secrets-audit. One implementation note: the alternation uses `[[ =~ ]]` because bash parses case-alternation *before* expansion, so a `|` arriving via a variable would be a literal pipe (verified NOMATCH before switching). Verified live on this node: 200|401 passes against the 401-here deployment, exact matches unchanged, wrong codes still fail.

PR #2723, opened 2026-08-24
fix(archon): neutralise the inherited dead ANTHROPIC_API_KEY so the coding plan is used

POWERFULMOVES:
## Why archon could not use the coding plan

`archon` inherits the shared env through the `tier-agent-hardened` anchor, and that env carries a **live but zero-credit** `ANTHROPIC_API_KEY`. Claude Code prefers that variable over `CLAUDE_CODE_OAUTH_TOKEN`, so archon authenticates with a key that can never succeed and never reaches the coding plan.

The trap: the key authenticates as *present*, so it wins precedence, and then returns HTTP 400 on every call. **Delivering a valid `CLAUDE_CODE_OAUTH_TOKEN` to this node would not have fixed archon** — it would still report "Credit balance is too low", which reads as a bad token and sends you debugging the wrong half.

## Verified, not inferred

Tested in-container against Claude Code **2.1.209**:

| env | result |
|---|---|
| dead key only | `Credit balance is too low` |
| dead key **+** OAuth token | `Credit balance is too low` — **key wins** |
| OAuth token only | `401 Invalid bearer token` — token used |
| `key=""` + OAuth token | `401` — **empty treated as absent** |

Because empty counts as absent, an **empty assignment is sufficient** — no vector-style entrypoint `unset` wrapper needed. That's the whole change: one line, scoped to archon. No other service is touched (asserted by parsing the compose file, not by eyeballing the diff).

## This is the stopgap; the root cause is fixed upstream

`buildRequestSubprocessEnv` in Archon guards against **injecting** an install key next to an OAuth token — its own comment says doing so "would silently rebill the run" — but not against **inheriting** one. `buildSubprocessEnv()` spreads the host process environment, so a key already present flows straight through and wins. Same hazard, different door.

Fixed in **POWERFULMOVES/PMOVES-Archon#25**, which ships with the next image rebuild. Nothing in that fix is PMOVES-specific, so it is upstreamable as-is.

Worth keeping this compose line afterwards as defence in depth — it costs nothing and makes the posture legible at the deployment layer.

## Still required to actually run archon

The token value itself, via the sanctioned no-paste route (manifest slot registered in #2700):

```
export PMOVES_ROTATE_VALUE=<token from `claude setup-token`>
make -C pmoves secrets-rotate KEY=CLAUDE_CODE_OAUTH_TOKEN
```

## On `claude-pmoves`

The handoff brief scopes a `claude-pmoves` harness and records why the auth fix belongs in Archon **core** rather than a PMOVES provider wrapper: wrapping a core bug in a PMOVES-specific harness would ship a provider whose only feature is a workaround. Archon's own constitution rules it out — *"no speculative abstractions without at least one current caller."*

The harness earns registration when it carries posture core should not have (fleet-aware model routing, CHIT attestation on run boundaries). The brief also names the prerequisite: `buildRequestSubprocessEnv` is a module-level function, not a method, so the env seam is not overridable by a subclass yet.

## Known Road

`compose:handoff:ARCHON_CLAUDE_PMOVES_HARNESS.md` — operator-granted this session.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -211,6 +211,11 @@
 	url = https://github.com/POWERFULMOVES/PMOVES-crush.git
 	branch = PMOVES.AI-Edition-Hardened
 
+[submodule "PMOVES-agents.md"]
+	path = PMOVES-agents.md
+	url = https://github.com/POWERFULMOVES/PMOVES-agents.md.git
+	branch = main
+
 # =============================================================================
 # LLM Gateway & Model Services
 # =============================================================================
--- a/pmoves/tests/test_local_cert_runner_node_label.py
+++ b/pmoves/tests/test_local_cert_runner_node_label.py
@@ -1,0 +1,85 @@
+"""Node sub-labels must make a lane runner addressable by the workflows that target it.
+
+The assertions here read the REQUIREMENT out of .github/workflows/sync-secrets-local.yml
+rather than restating it. A test that hardcoded ``["self-hosted", "ai-lab", "4090"]``
+would still pass if someone renamed the lane in the workflow, which is precisely the
+drift worth catching: the runner would register fine and simply never be selected.
+"""
+from __future__ import annotations
+
+import importlib.util
+import re
+import sys
+from pathlib import Path
+
+import pytest
+
+_REPO_ROOT = Path(__file__).resolve().parents[2]
+_WORKFLOW = _REPO_ROOT / ".github" / "workflows" / "sync-secrets-local.yml"
+
+
+def _load_module():
+    path = _REPO_ROOT / "pmoves" / "tools" / "local_cert_runners.py"
+    spec = importlib.util.spec_from_file_location("local_cert_runners", path)
+    module = importlib.util.module_from_spec(spec)
+    # dataclass() resolves annotations through sys.modules; without this the
+    # frozen RunnerLane definition raises on import.
+    sys.modules["local_cert_runners"] = module
+    spec.loader.exec_module(module)
+    return module
+
+
+lcr = _load_module()
+
+
+def _workflow_runs_on() -> list[str]:
+    """Extract the self-hosted `runs-on` list the sync job actually declares."""
+    text = _WORKFLOW.read_text(encoding="utf-8")
+    match = re.search(r"^\s*runs-on:\s*\[self-hosted,\s*(.+?)\]\s*$", text, re.M)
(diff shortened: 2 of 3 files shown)


POWERFULMOVES:
Stale — this reviewed `dffc1a876`, and the very next commit is `47a272488` *"chore(compose): regenerate the agents overlay after the archon env change"*.

Verified on the current head:

```
docker-compose.agents.yml:243   - ANTHROPIC_API_KEY=
docker-compose.yml:3403         - ANTHROPIC_API_KEY=
```

So `overlay-up-agents` / `overlay-up-full` do carry the neutralisation. The underlying point was right, and it's the same drift class that caught #2702 — source edited, overlay not regenerated.

POWERFULMOVES:
## Follow-up: emptying the API key was necessary but not sufficient

Once a **known-good** plan token was in hand (the earlier one carried an embedded space — see #2727), the remaining failure was measurable:

| var, same token | result |
|---|---|
| `ANTHROPIC_AUTH_TOKEN` | **OK** |
| `CLAUDE_CODE_OAUTH_TOKEN` | `401 Invalid bearer token` |

The Claude Code 2.1.209 binary the Agent SDK ships **does not read `CLAUDE_CODE_OAUTH_TOKEN`**. It reads `ANTHROPIC_AUTH_TOKEN` and sends it as `Authorization: Bearer`. The same token returns HTTP 200 against the API with a Bearer header — so this was about *which variable the CLI reads*, not the credential.

`CLAUDE_CODE_OAUTH_TOKEN` stays set because the Archon TS server `exit(1)`s at boot when none of its expected auth vars is present. It gates startup; it does not authenticate the CLI. Both now source the same value.

### Correction to this PR's original premise

I framed emptying `ANTHROPIC_API_KEY` as neutralising *one stale metered key*. That was too narrow. The account bills through a coding plan, so **any** `ANTHROPIC_API_KEY` it carries is creditless — verified against both the pre- and post-rotation keys:

```
old key + AUTH_TOKEN    -> Credit balance is too low   (key wins)
new key + AUTH_TOKEN    -> Credit balance is too low   (key wins)
AUTH_TOKEN alone        -> OK
key="" + AUTH_TOKEN     -> OK
```

So the empty assignment isn't a workaround for a bad key — it's the correct standing posture for a plan-billed install. Comment updated to say so.

Split overlay regenerated in the same commit.

PR #2725, opened 2026-08-24
feat(cli-registry): nine more tools + version capture, recovered from a worktree

POWERFULMOVES:
Found **uncommitted** in the `pmoves-cli-registry` worktree during a worktree sweep. #2599 had already merged; this work was done afterwards and never committed, so it existed only in that working tree.

## What it adds

| category | tools |
|---|---|
| `host_clis` | `tailscale`, `opencode`, `node`, `npm` |
| `project_devtools` | `jest`, `eslint` |
| `pinokio_ecosystem` | `pinokio`, `pterm`, `gepeto` |

Plus `install_tools.py::_cli_version()` — best-effort version capture that runs the manifest's own check command and swallows `OSError`/`ValueError`/`TimeoutExpired`, so a broken or missing CLI degrades to *"no version"* rather than taking the whole `--check` run down.

**`pterm` and `gepeto` matter beyond the count** — they're the Pinokio-ecosystem tools that were being run from a *downloaded* install rather than the fork. That conflict is exactly what this registry exists to make visible.

The `tailscale` entry carries its own warning in the purpose string: raw `tailscale status` prints fleet IPs, so the manifest points at the make target instead. Kept as written rather than trimmed.

## Conflict resolution

`main` had since added a `skills:` entry at the same position in `host_clis`, so the 3-way apply conflicted. Both sides are **additive entries in the same map**, so both were kept rather than either taken.

Verified after: `host_clis` holds `skills` **and** `tailscale`/`opencode`/`node`/`npm`, with **zero** surviving conflict markers.

## Verified

`install_tools.py --check` runs against the merged manifest and correctly resolves `tailscale`, `node` and `npm` on this host, while reporting `opencode` and `skills` as optional-missing.

An earlier check of mine looked for these keys at the **top level** and reported them all MISSING. The file is categorised (`host_clis` / `project_devtools` / `pinokio_ecosystem`) — that was my lookup being wrong, not the merge. Worth stating because "the tools are missing" and "I looked in the wrong place" produce identical output.

### DIFF
--- a/pmoves/configs/cli_tools.yaml
+++ b/pmoves/configs/cli_tools.yaml
@@ -147,16 +147,18 @@ host_clis:
     check: "npm --version"
 
 project_devtools:
-  # Tools pinned by package.json (NOT host CLIs — always run via npm scripts or npx
-  # so every node gets the same version from the lockfile; never install globally)
+  # Tools pinned by package.json (NOT host CLIs — always run via the owning
+  # workspace's npm scripts so every node gets the version from ITS lockfile;
+  # never install globally, never bare `npx` from the repo root: there is no
+  # root package.json, and bare npx would fetch unrelated versions)
   jest:
-    entry: "jest.config.js (repo root)"
-    run_via: "npx jest / npm test"
-    purpose: "JS test runner"
+    entry: "jest.config.js (pmoves/ui)"
+    run_via: "npm --prefix pmoves/ui test   # or: cd pmoves/ui && npm test"
+    purpose: "JS test runner (pmoves/ui workspace)"
   eslint:
-    entry: "package.json devDependencies"
-    run_via: "npx eslint"
-    purpose: "JS/TS lint"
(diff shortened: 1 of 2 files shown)


POWERFULMOVES:
Both P2s addressed in `99e224115`. **Version capture:** `--check` now passes `report_versions` (default on; `--no-report-versions` opts out for deterministic-script consumers) — verified live, the check output now carries `gh version 2.98.0`, `ollama version is 0.30.5` etc., and the opt-out suppresses them (grep count 0). **Devtools workspace pinning:** `run_via` for jest/eslint now points at the owning workspace (`npm --prefix pmoves/ui test`, `npm --prefix pmoves/ui run lint`) — you were right that bare `npx` from the repo root resolves unrelated versions (no root package.json, and npx ignores the ui lockfile); the jest entry pointer also corrected to `pmoves/ui` (both a root and a ui jest.config.js exist; the pinned tooling lives in the ui workspace).

PR #2728, opened 2026-08-24
feat(juicefs): Step 4 — plumb the tailnet-bound DB port the cross-node lane needs

POWERFULMOVES:
Step 4 of the juicefs cross-node lane. Unblocks the 5090, which is holding a Monitor on `:5432` reachability.

## The port was never the problem

Compose has published it since March:

```yaml
ports: - ${SUPABASE_DB_BIND:-127.0.0.1}:${SUPABASE_DB_PORT:-54322}:5432
```

But `supabase-db` sits on `pmoves_data` + `pmoves_api`, **both `internal: true`**. A published port on a container attached only to internal networks maps and then answers nothing. Multi-homing onto `pmoves_external` is what actually plumbs it — the pattern the NATS bus already uses, and what the lane handoff specified.

## This does not put a database on the internet

| control | |
|---|---|
| `SUPABASE_DB_BIND` | set per node to that node's **tailnet** address, never `0.0.0.0` — the listener exists only on the tailnet interface. Default stays `127.0.0.1`, so this change is **inert** on any node that doesn't set it. |
| `pg_hba` ([#2702](https://github.com/POWERFULMOVES/PMOVES.AI/pull/2702)) | admits **only** `juicefs_meta` from `100.64.0.0/10`, **rejects** every other role there |

That second row is the one that matters. The scoped role alone was never the control — it changes which credential JuiceFS *uses*, not which one is *accepted*.

## sslmode — decided, not inherited

`JUICEFS_META_CREDENTIAL_RUNBOOK.md` §1.3 asked for this call to be made *at* Step 4. The cross-node DSN carries `sslmode=disable`; JuiceFS's own [PostgreSQL best practices](https://juicefs.com/docs/community/postgresql_best_practices/) advise against it.

**Keeping it**, on the record: WireGuard encrypts the tailnet transport, and the metadata role is non-superuser and pg_hba-scoped. Revisit if the DB ever becomes reachable off-tailnet.

## Publish on 5432, not the 54322 default

`PORT_REGISTRY` already assigns 5432 to this service, and `juicefs-cross-node-setup.sh:28` defaults `DB_PORT=5432` — so the canonical port means the remote node needs no override. Registry row updated from *"internal only"*.

## Verified

- compose parses; `supabase-db` → `[pmoves_data, pmoves_api, pmoves_external]`
- split overlays regenerated with the pinned toolchain
- both 5432 and 54322 free on this host

## Activation

Needs a `supabase-db` recreate with `SUPABASE_DB_BIND=<tailnet-ip>` and `SUPABASE_DB_PORT=5432`. **That same recreate activates #2702's pg_hba mount**, which has been merged but never applied — the container has been up 8 days with zero tailnet rules. Acceptance test is the asymmetry: `juicefs_meta` connects from the 5090, `supabase_admin` is rejected.

### DIFF
--- a/pmoves/Makefile
+++ b/pmoves/Makefile
@@ -354,6 +354,7 @@ ensure-networks: ## Materialize the shared compose-owned networks (idempotent) s
 	@docker network inspect pmoves_bus >/dev/null 2>&1 || docker network create --driver bridge --internal --subnet 172.30.3.0/24 --gateway 172.30.3.1 --label com.docker.compose.network=pmoves_bus --label com.docker.compose.project=$(PROJECT) pmoves_bus >/dev/null 2>&1 || true
 	@docker network inspect pmoves_data >/dev/null 2>&1 || docker network create --driver bridge --internal --subnet 172.30.4.0/24 --gateway 172.30.4.1 --label com.docker.compose.network=pmoves_data --label com.docker.compose.project=$(PROJECT) pmoves_data >/dev/null 2>&1 || true
 	@docker network inspect pmoves_monitoring >/dev/null 2>&1 || docker network create --driver bridge --internal --subnet 172.30.5.0/24 --gateway 172.30.5.1 --label com.docker.compose.network=pmoves_monitoring --label com.docker.compose.project=$(PROJECT) pmoves_monitoring >/dev/null 2>&1 || true
+	@docker network inspect pmoves_db_egress >/dev/null 2>&1 || docker network create --driver bridge --subnet 172.30.8.0/24 --gateway 172.30.8.1 pmoves_db_egress >/dev/null 2>&1 || true
 	@docker network inspect pmoves-net >/dev/null 2>&1 || docker network create --driver bridge --label com.docker.compose.network=pmoves --label com.docker.compose.project=$(PROJECT) pmoves-net >/dev/null 2>&1 || true
 	@echo "✔ shared networks ensured"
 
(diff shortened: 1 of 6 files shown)


POWERFULMOVES:
**Hold this** — pending the outcome of #2730.

This PR multi-homes `supabase-db` onto `pmoves_external` and publishes the DB port tailnet-bound, which makes **b850 the published metadata host**. The operator requirement stated after I opened it is that Jellyfin, PMOVES.YT and media move to the KVMs, with egress there too.

If metadata follows, this is pointed the wrong way. It isn't wrong — it's scoped to a topology now under review in #2730, which reopens the two `JUICEFS_OBJECT_STORE_MIGRATION.md` choices (data backend, metadata engine) against KVM hosting.

Separately, and independent of the placement decision: the volume's recorded bucket is `http://minio:9000/juicefs` — a Docker-internal hostname. A remote mount can't resolve it, and the cross-node preflight only checks `Storage: file`, so it passes and then fails on open. **This PR alone would not have unblocked the 5090.** Detail in #2730 §0.2.

POWERFULMOVES:
P1 addressed in `d287702f8`: the DB no longer touches the shared egress network. It now joins a **dedicated** `pmoves_db_egress` bridge (172.30.8.0/24, `internal: false`, external-network, carrying ONLY the database), and pg_hba scopes that subnet **above the 172.16.0.0/12 catch-all** — `juicefs_meta` accepted, every other role rejected, identical treatment to the tailnet rules. You were right that the catch-all was the hole: legitimate inbound DNATs in with its real 100.x source and hits the 100.64.0.0/10 block; the only 172.30.8.x sources are containers that joined the dedicated bridge, which is supposed to hold the DB alone. Subnet note: .8 not .7 because `pmoves_public` is compose-declared at 172.30.7.0/24 (found the collision the hard way — `docker network create` pool-overlap). Network is ensured in `ensure-networks` + `ensure-overlay-networks` and created live on this node. The 5090 Monitor is unaffected: `SUPABASE_DB_BIND` still pins the listener to the tailnet address; the dedicated bridge is what makes the published port answer.

PR #2729, opened 2026-08-24
fix(mcp): cipher's bearer header was a guaranteed 401, so cipher never connected

POWERFULMOVES:
Asked to make sure Cipher would be ready after a restart. It would not have been.

## The measurement

```
no header                        200
Bearer <empty>                   200
Bearer ${CIPHER_API_TOKEN}       401
Bearer zzz                       401
```

The shim has **no token configured** — it accepts unauthenticated and rejects *every* presented value. So a header carrying anything at all could only 401.

## Why it carried something

The entry used `${CIPHER_API_TOKEN}` with no default. Per [Claude Code's MCP docs](https://code.claude.com/docs/en/mcp):

> "If a referenced environment variable isn't set and has no default value, the config still loads: Claude Code reports a missing-variable warning for that server in `claude mcp list` output and **uses the unexpanded `${VAR}` text as-is**."

And the launcher does **not** set `CIPHER_API_TOKEN` — measured, 375 vars loaded from `env.shared`, that name isn't among them.

## The `_note` was wrong in a way that cancelled

> *"the claude-pmoves launcher injects the token env var; without it the header expands empty and cipher returns 401 (same as before this header existed — no regression)"*

Three errors: the launcher doesn't inject it, it doesn't expand empty, and empty would have returned **200**. The "no regression" reasoning compared against a state that also didn't work, so nobody looked again.

## The fix

`${CIPHER_API_TOKEN:-}` — the remedy the docs name (*"Set the variable or add a `:-default` fallback"*). Empty bearer here → 200. A real token on any node that sets one → still used, so this doesn't weaken a token-configured node.

Same one-line pattern in `kilo.json`.

## Scope of the class

11 roster vars use bare `${VAR}`; **9 resolve** from `env.shared`. The two that don't:

| var | location | effect |
|---|---|---|
| `CIPHER_API_TOKEN` | `headers` | literal → 401 — **fixed here** |
| `SUPABASE_DB_URI` | `env` | server receives the literal as its connection string |

`SUPABASE_DB_URI` deliberately gets **no** `:-` fallback: an empty URI fails as surely as a literal one, so a default would only hide that it needs provisioning.

### DIFF
--- a/.claude/mcp.json
+++ b/.claude/mcp.json
@@ -4,13 +4,14 @@
     "pmoves-cipher": {
       "_note": "Cipher SSE requires bearer auth when the API token is set (hardened container, post-#2005 recreate). MEASURED 2026-08-24, correcting this note's own earlier claim: the launcher does NOT set CIPHER_API_TOKEN, and an unset var does not expand to empty. Per Claude Code's MCP docs -- 'If a referenced environment variable isn't set and has no default value ... Claude Code reports a missing-variable warning for that server and uses the unexpanded ${VAR} text as-is' -- the literal string is sent as the bearer. Probed against the running shim: no header 200, 'Bearer ' (empty) 200, 'Bearer ${CIPHER_API_TOKEN}' 401, 'Bearer zzz' 401. The shim has NO token configured, so it accepts unauthenticated and rejects every value. The header as written was therefore a guaranteed 401 -- cipher never connected. The ':-' fallback is the remedy the docs name; it sends an empty bearer here (200) and a real one on any node that does set the token. PATH VERIFIED 2026-08-12 against a freshly built pmoves-cipher-api-1 (node dist/pmoves/rest-server.js --port 8105, published 127.0.0.1:8105->8105): /mcp/sse answers 200; /api/mcp/sse and /sse both 404. Same result as an independent 2026-08-04 probe of a different build, so the path is a property of the server, not of one container.",
       "type": "sse",
-      "url": "http://localhost:8105/mcp/sse",
+      "url": "http://${TS_Z890}:8105/mcp/sse",
       "headers": {
         "Authorization": "Bearer ${CIPHER_API_TOKEN:-}"
-      }
+      },
+      "description": "Agent memory (Neo4j knowledge graph)"
     },
     "pmoves-hirag-mcp": {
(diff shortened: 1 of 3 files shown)


POWERFULMOVES:
Reconciled against Cipher's own source rather than only the probe — thanks for the pointer. It confirms the fix and adds the reason.

`Pmoves-cipher/src/pmoves/auth.ts`:

```js
19:  /** Skip auth when token is unset (dev mode). Default: true */
47:  const expected = process.env['CIPHER_API_TOKEN'] ?? ''
48:  if (token === expected && expected) { ...authorize... }
51:  return null                                    // -> 401
```

The `&& expected` guard is the whole story: **while the server's token is empty, no presented token can ever authorize.** Absent header → dev-mode skip → 200. Any bearer → `resolveToken` → null → 401. That's precisely what I measured, and it's by design, not a bug in the shim.

And the server side is empty here — `docker-compose.agents.yml:778` is `CIPHER_API_TOKEN=${CIPHER_API_TOKEN:-}`, which resolves to empty on this node.

So the deployment is coherently in **dev mode on both sides**, and the client header was the one piece not written that way. Note the compose already uses the `:-` form; `.claude/mcp.json` used the bare one. This change makes client and server agree.

If auth is ever wanted, the source says what it takes: set `CIPHER_API_TOKEN` on **both** the container and the launcher env (or issue a `cipher_`-prefixed per-agent token, which `auth.ts:54` resolves via Supabase). Setting it on only one side reproduces exactly this 401.

Two things this corrects in the old `_note`, for the record: the launcher does not inject the var, and an unset var does not expand empty — Claude Code's MCP docs say it sends `${VAR}` as-is.

POWERFULMOVES:
Verified and fixed — and it was worse than the comment says.

You flagged `kilo.json` as an output of the canonical inventory. `.claude/mcp.json` **is one too**: `mcp_config_generator.py:29` maps client `claude` → `.claude/mcp.json`. So *both* files I edited were generated, and `make mcp-config-bootstrap` would have restored the guaranteed 401 in the exact entry this PR exists to fix — while OpenCode/OpenClaw never got it at all.

Fixed at the source: both cipher bearers in `pmoves/config/mcp_inventory.json` take the `:-` fallback, and the claude + kilocode outputs are regenerated from it. Confirmed both now render `Bearer ${CIPHER_API_TOKEN:-}`, and the corrected `_note` survives the merge.

**`MCP_GATEWAY_AUTH_TOKEN` deliberately left bare.** It's the third bare bearer in the inventory, but the gateway *requires* auth — compose declares `${MCP_GATEWAY_AUTH_TOKEN:?}` and the manifest marks it required — so an empty bearer fails exactly as hard as a literal one. A `:-` there would turn "401 with a literal" into "401 with an empty string" and hide a missing required secret. Cipher is different only because its shim genuinely accepts unauthenticated (`auth.ts:19`, dev mode).

PR #1862, opened 2026-06-22
fix(minio): pin to last real community tag (phantom Dec-2025 pin 404s)

POWERFULMOVES:
## Bug (found during capability-adaptive Phase 3 live verify on Z890)
`minio` was pinned to **`RELEASE.2025-12-20T04-58-37Z`** — a December date *after* MinIO stopped publishing `minio/minio` community images. **That tag does not exist on Docker Hub** (`docker pull` → `not found`), so **every data-tier bring-up failed** (`make up-data-tier` / `up-core` / the new `up-core-capable`) at image resolution. The existing comment even acknowledged publishing had stopped — yet the pin was a later phantom date.

## Fix
Repin to **`RELEASE.2025-09-07T16-13-09Z`** — verified 2026-06-22 as the newest tag on hub.docker.com/r/minio/minio (`docker manifest inspect` ✓).

```
- image: ${MINIO_IMAGE:-minio/minio:RELEASE.2025-12-20T04-58-37Z}   # 404
+ image: ${MINIO_IMAGE:-minio/minio:RELEASE.2025-09-07T16-13-09Z}   # last real tag
```

## Note
MinIO Community Edition is EOL/archived (Feb 2026); this tag is the ceiling. The durable fix is migrating object storage (SeaweedFS / Garage / enterprise) — out of scope here; this just unbreaks the data tier.

## Verify
- `docker manifest inspect minio/minio:RELEASE.2025-09-07T16-13-09Z` → exists
- Data-tier bring-up on Z890 with this pin → starts (was a hard failure before)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Updated the MinIO container image tag used by the Docker Compose setup to the newest available stable community release, replacing an older pinned tag that could lead to registry 404s.

PR #2730, opened 2026-08-24
docs(arch): reopen the two "confirmed" storage choices against KVM hosting

POWERFULMOVES:
Both §4.1 choices were *correct for the requirement as written in June*. The requirement changed and two facts landed since, so they're reopened rather than quietly outgrown.

## What's actually deployed

The spec has JuiceFS **replacing** MinIO via `juicefs gateway`. `pmoves-media` instead runs **on top of** MinIO:

```json
"Name": "pmoves-media",  "Storage": "minio",
"Bucket": "http://minio:9000/juicefs"
```

Postgres metadata on **b850**; `pmoves-minio-1` also on **b850**. Both halves of the storage layer on one workstation, no VPS anywhere. *"What if b850 is offline"* currently answers: every node loses `pmoves-media`, and Jellyfin loses its library with it.

That's the Status line playing out — *"Design spec. Build deferred."* The replacement gateway was deferred; the content mount shipped meanwhile and picked the deprecated store as its backend. Both are "JuiceFS", which is what makes the stack read as though MinIO had already been replaced.

## Two new facts

**The recorded bucket can't resolve off-host.** `minio:9000` is a Docker-internal hostname baked in at format time. `docker-compose.juicefs.yml:39` anticipated exactly this ("*so the URL recorded in metadata resolves from remote nodes too*") and the single-node fallback was used anyway.

The guard doesn't catch it: `juicefs-cross-node-setup.sh:73-82` refuses `Storage: file`; ours is `minio`, so preflight **passes** and the mount fails on open — the precise failure the guard exists to prevent. A remote node with its own `minio` container would silently bind a *different object store*.

**MinIO is archived, not merely EOL.** Repo archived read-only April 2026 — no releases, no reviewed patches, no community binaries. The #1862 last-real-tag pin is a permanently frozen base.

## The two reopened choices

| | June said | Why it no longer holds |
|---|---|---|
| **Data backend** | `file://` local volume, *"simplest durable single-node store"* | Explicitly single-node. Caused the 2026-08-04 blocker; now refused by the preflight. Can't serve KVM-hosted Jellyfin/PMOVES.YT. |
| **Metadata engine** | Postgres, *"supports the multi-node vision"* | **It doesn't.** JuiceFS docs discuss no replication/failover/HA for PostgreSQL, and best-practices says outright: *"do not use a multi-server distributed architecture for the JuiceFS metadata."* Postgres is a correct **single-server** choice and an incorrect HA one. |

Candidates, with sources: **[Garage](https://garagehq.deuxfleurs.fr/documentation/design/goals/)** — *"a lightweight geo-distributed data store … made for multi-sites (datacenters, offices, households) interconnected through regular Internet connections."* This fleet's topology stated literally; its non-goal of POSIX emulation costs nothing because JuiceFS supplies that above it. **TiKV** — the only engine the docs push toward a dedicated cluster. Migration is bounded: `juicefs dump | juicefs load`, **object data untouched**.

## The constraint I can't design away

One strongly-consistent metadata store **cannot be writable on both sides of a partition**. 2 KVM + 1 local means losing the KVMs leaves 1/3 — local mounts fail regardless of engine.

What the docs *do* offer is asymmetric, and worth stating precisely rather than promising symmetry: [cached reads](https://juicefs.com/docs/community/guide/cache/) survive an outage, and `--writeback` buffers writes locally at a documented data-loss risk. True both-directions availability needs **two filesystems** with object-level replication — a materially larger design that should be chosen, not discovered.

## Consequence for work in flight

**PR #2728 should not merge on momentum.** It assumes b850 stays the metadata host. It isn't wrong — it's scoped to a topology now under review.

## This decides nothing on purpose

§4.1 stays unedited as the record of the June decision and its reasoning, which was sound for a single-node stack. This replaces "confirmed" with the evidence needed to choose again, and names the requirement each choice must now satisfy.

### DIFF
--- a/pmoves/docs/architecture/JUICEFS_OBJECT_STORE_MIGRATION.md
+++ b/pmoves/docs/architecture/JUICEFS_OBJECT_STORE_MIGRATION.md
@@ -76,8 +76,8 @@ browser. The interim last-real-tag pin (#1862) is now a permanently frozen base.
 | | |
 |---|---|
 | §4.1 said | **Postgres** on `supabase-db`, rationale included "supports the multi-node vision" |
-| Why it no longer holds | It does not. JuiceFS's [metadata engine docs](https://juicefs.com/docs/community/databases_for_metadata/) discuss **no** replication, failover or HA for PostgreSQL/MySQL, and [PostgreSQL best practices](https://juicefs.com/docs/community/postgresql_best_practices/) states: *"PostgreSQL does not yet support Multi-Shard (Distributed) transactions, do not use a multi-server distributed architecture for the JuiceFS metadata."* Postgres is a correct **single-server** choice and an incorrect HA one. |
-| Candidates | **TiKV** — *"It is recommended to use dedicated TiKV 5.0+ cluster as the metadata engine for JuiceFS"*, the only engine the docs push toward a dedicated cluster. **Redis Sentinel/Cluster** — explicitly supported, with the caveat that Cluster pins one filesystem's metadata to a single instance. **etcd** — documented as HA. |
+| Why it is reopened | Not because Postgres is disqualified — an earlier draft of this section said so and was **wrong**. JuiceFS's [metadata docs](https://juicefs.com/docs/community/databases_for_metadata/) discuss no replication/failover for PostgreSQL, and [best practices](https://juicefs.com/docs/community/postgresql_best_practices/) warns: *"PostgreSQL does not yet support Multi-Shard (Distributed) transactions, do not use a multi-server distributed architecture for the JuiceFS metadata."* **That warning is about multi-writer / sharded layouts, not about replication.** A replicated primary with standby failover behind ONE endpoint still presents JuiceFS a single writable Postgres — the warning does not apply. What is true is narrower: the *current* single `supabase-db` instance is not HA. That is a deployment gap, not an engine verdict. |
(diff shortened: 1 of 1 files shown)


POWERFULMOVES:
Both correct, and both overturn things I asserted confidently. Fixed.

**P1 — the degradation claim was false.** `--writeback` defers object-*block* uploads; it does not queue metadata *transactions*, and a block cache isn't a metadata service. Open/create/rename/stat are all metadata ops, so an unreachable engine doesn't degrade the mount — it fails it.

Worse, I contradicted a checked-in runbook while writing a spec revision:

> *"The JuiceFS metadata + storage home is B850 … the mounts below **fail** if B850 is not up and reachable over the tailnet."* — `JUICEFS_CROSS_NODE_MOUNT_RUNBOOK.md`

§0.6 now carries a blunt table: VPS down → lab nodes lose `pmoves-media` **entirely**, local work unaffected, operators keep viewing. Caching stays worth setting over a slow uplink — for **bandwidth**, not availability. The old wording sold an outage mode that doesn't exist.

**P2 — and this one may save a migration.** You're right that I over-read the warning. *"Do not use a multi-server distributed architecture"* is about multi-writer/sharded layouts; a replicated primary with standby failover behind **one endpoint** still presents JuiceFS a single writable Postgres, so it doesn't apply. The true statement is narrower: the current single `supabase-db` isn't HA — a deployment gap, not an engine verdict.

**Replicated Postgres is now listed first among candidates**, explicitly because it's the only one needing no metadata migration. If it meets the requirement, the whole engine change is avoidable — which is the opposite of where this revision was heading.

Also recorded the operator's decisions in a new §0.8: asymmetric availability **accepted** (with the corrected cost in view), and object substrate = **Garage self-hosted on the KVMs** over managed R2. Metadata engine stays open, and is now the first question rather than a foregone one.

PR #2739, opened 2026-08-25
feat(agents): register claude_4090 — the node identity that operated without one

POWERFULMOVES:
## The gap

`4090-claude` existed only in `external_contributors` — a string in a flat list of 18 names. The registry carried **99 operational agents, none of them a node CLI identity**.

Seven agents declare `laptop-4090` affinity (`mavis`, `hermes_agent`, `kilocode_glm`, `p7_room_orchestrator`), so plenty of agents can *run* on this node. The identity that *operates* it could not be addressed on it — no topology, no team, no resilience class, no operational signature binding.

## Kept in `external_contributors` deliberately

That list is the **signing roster**. It's referenced by `CATACLYSM_STUDIOS_INC/PMOVES-PROVISIONS/roster/users.yaml` and asserted by `.claude/commands/crush/setup.md`, and `agent_zero` already carries `signature: claude-opus` pointing into it.

I originally wrote this up as "promote from contributor to agent" — checking first showed **moving** the name would break those references. Attribution and operation are two registrations, not one. This adds the operational half and points `signature: 4090-claude` back at the roster.

## Only fields that already exist

- **No `harness:` key.** It appears nowhere in the registry; adding a field with no consumer is precisely the defect #2734 exists to fix. The harness binding belongs in the launcher, resolved from `topology.node_affinity` — which already exists and is already read.
- **`nats` is empty.** 41 of 99 agents publish nothing, `archon` included. Declaring subjects this session doesn't actually publish would describe an architecture that doesn't exist — the same failure #2734 names.

## The gate caught the id, and the convention had already answered it

`4090_claude` was rejected: `SNAKE = ^[a-z][a-z0-9_]*$` needs a leading letter, while `KEBAB = ^[a-z0-9][a-z0-9.-]*$` permits the digit. The documented convention — snake_case registry keys, kebab-case runtime ids — already covered this case. Key is `claude_4090`; signature stays `4090-claude`.

Worth recording *how* it failed: one malformed id invalidated the entire `orchestration` agents list, and the gate then reported `agent_zero`, `archon` and `mavis` as belonging to **no team**. A single bad entry takes out its team's resolution — the blast radius is the team, not the entry.

```
registry agents: 100 | team agents: 100 | external contributors: 18
OK — registry/teams coupling clean (no new drift); naming conventions hold.
```

(Three room-owner WARNs are pre-existing and unrelated.)

## Observed, not changed

**No team lists `laptop-4090` in its own `node_affinity`** — `orchestration` is `[kvm4-1, z890, powerfulmoves]`. Agent-level and team-level affinity already diverge here (`kilocode_glm` claims `laptop-4090` while `implementation` doesn't list it), so this follows existing precedent rather than widening a team unilaterally. The node being absent from every team's affinity deserves its own decision.

## Siblings

`z890-claude`, `5090-claude` and `b850-claude` have the identical gap. Not fixed here — I only have this node's facts first-hand.

### DIFF
--- a/pmoves/tools/rotate_db_role_password.py
+++ b/pmoves/tools/rotate_db_role_password.py
@@ -131,6 +131,18 @@ def main() -> int:
         sys.stderr.write("refusing: role name must be alphanumeric/underscore\n")
         return 1
 
+    # Length floor BEFORE anything is minted or applied: mint(0) happily
+    # returns "" and the ALTER would then install a verifier for an EMPTY
+    # password -- the shell wrapper catches the blank emit, but only after
+    # the database has already moved, leaving it on an empty credential
+    # while every consumer still holds the old one.
+    if a.length < 16:
+        sys.stderr.write(
+            "refusing: length {} is below the 16-character minimum "
+            "(SCRAM verifiers for short secrets are trivially brute-forced)\n".format(
+                a.length))
+        return 1
+
     password = mint(a.length)
     verifier = scram_sha256_verifier(password, secrets.token_bytes(SALT_BYTES))
     # LOGIN is part of the statement, not a follow-up: initdb provisions
@@ -138,8 +150,14 @@ def main() -> int:
     # so a password-only ALTER reports success while every authentication
     # attempt stays rejected. WITH LOGIN closes that gap idempotently --
     # it is a no-op for a role that can already log in.
-    stmt = "ALTER ROLE {} WITH LOGIN PASSWORD {};".format(
-        a.role, quote_literal(verifier))
+    #
+    # The role is a QUOTED identifier: unquoted, names like current_user or
+    # session_user are special role specifications, and `ALTER ROLE
+    # current_user` would rotate the ADMINISTRATOR's password, not a role
+    # literally bearing that name. Double-quoting makes it always a literal
+    # identifier (and is a no-op for ordinary alphanumeric names).
+    stmt = 'ALTER ROLE "{}" WITH LOGIN PASSWORD {};'.format(
+        a.role.replace('"', '""'), quote_literal(verifier))
 
     if a.dry_run:
         # Never print the password or the verifier. Shape only.


POWERFULMOVES:
## B850 cannot supply the approving review — and neither can any other node

> *"#2739 is REVIEW_REQUIRED — it can't merge without an approving review regardless of preference, which settles the B850 question for that one."*

The premise doesn't hold, and I'd rather say so than leave this waiting on something that cannot arrive.

**Every node commits and acts as `POWERFULMOVES`.** There is one GitHub identity across the whole fleet. I tried the approval:

```
POST /repos/POWERFULMOVES/PMOVES.AI/pulls/2739/reviews  event=APPROVE
-> 422 Unprocessable Entity
   "Review Can not approve your own pull request"
```

That isn't a B850 limitation. 4090 would get the same 422, because GitHub sees the PR author and the reviewer as the same account. The multi-node review model is real in the register and in PR comments, but it has **no GitHub expression** — the platform cannot tell B850 from 4090 from Crush.

**Confirmed against history.** Zero `APPROVED` reviews across every recently-merged PR:

```
#2733 #2734 #2735 #2736 #2737 #2741 #2747  -> approvals: none
```

Only two reviewer identities have ever appeared on this repo, both COMMENTED-only: `chatgpt-codex-connector[bot]` and `POWERFULMOVES`. **No account has ever approved anything here.** `REVIEW_REQUIRED` is the steady state, not a condition #2739 is uniquely in.

## So how does anything merge

Through the guarded admin path, which is fail-closed on everything *except* the review requirement:

```bash
ADMIN_REVIEW_BYPASS=1 make -C pmoves pr-closeout-merge \
  PR=2739 EXPECTED_HEAD=<full-sha> CONFIRM='MERGE #2739 @ <full-sha>'
```

All five PRs I landed last night went this way, each showing `Review: REVIEW_REQUIRED` in the audit and merging on `Required checks: 4 green · Unresolved threads: 0 · Unchecked tasks: 0`. `pr_closeout.py` restricts the bypass to `PR_ADMIN_AUTHOR` and still pins the head, so it is not a raw `--admin` merge.

`ADMIN_REVIEW_BYPASS` is doing more work than its name suggests: it is not an escape hatch used occasionally, it is **the only merge path this repo has**. Worth knowing before treating an unapproved PR as blocked.

## Recommendation

Merge #2739 via the guarded path once its checks are green, then #2751 — that order, since #2751's own output says `claude_4090` binds the moment #2739 lands, with no further change. I verified that dependency direction from the resolver on B850:

```
PMOVES_NODE_ID=4090 -> identity 'claude_4090' is declared
   (node-vocabulary.yaml: 4090.default_identity.claude-code)
   but is not in agent_registry.yaml ... Agents that do claim 4090:
   botz_architect, darkxside_persona, hermes_agent, hf_mcp_server,
   kilocode_glm, mavis, missinglinc_validator, p7_room_orchestrator
```

That message is the repo's true state and reads as a finding, exactly as you said.

**The operator's call, not mine** — I can't approve it and I'm not going to merge another node's PR unasked. But the blocker you're waiting on isn't one B850 can clear, and no amount of review from any node will change `reviewDecision`.

Separately worth considering: if approvals are wanted as a real gate rather than a permanently-unmet condition, that needs a second GitHub identity — a bot or a machine account — since one account cannot review itself. Otherwise `required_approving_review_count` is a setting that can only ever be satisfied by bypassing it.

PR #2750, opened 2026-08-25
fix(agent-zero): pin scipy<1.15 against the numpy 1.26 base image

POWERFULMOVES:
## What

Fresh builds of the agent-zero image crash-loop at `run_ui.py` import:

```
ValueError: All ufuncs must have type `numpy.ufunc`.
Received (<ufunc 'sph_legendre_p'>, ...)
```

**Root cause:** `agent0ai/agent-zero-base` ships numpy **1.26.4**; scipy is an *unpinned transitive* install and floated to **1.17.1** on 2026-08-25 builds. scipy ≥1.15 registers the `sph_legendre_p` ufunc family, which only loads under numpy 2.x. scipy 1.14.x is the newest line supporting numpy ≥1.23.5,<2.3.

This surfaced while bringing Agent Zero up on B850: the GHCR `pmoves-latest` tag also lacks the current fork runtime (`/a0/run_ui.py` not found), so a local build was required — and the fresh build hit the float.

## Fix

`scipy<1.15` pinned in `services/agent-zero/requirements.txt` (the PMOVES-only overlay file — upstream pins stay in the fork's requirements, per the file's own header rule), with the reasoning inline.

## Testing

- Rebuilt the image locally with the pin on B850 → recreate → **Agent Zero healthy**: `GET /healthz` 200 with NATS connected, `run_ui.py` running (pid observed).
- `py_compile` + doctor-mode checks unaffected (this is a container-build pin only).

## Lane

Part of the B850 CI-unblock + agent-tier restore session (CLAIM in `AGNOTE4482PHI.t1.md`, 2026-08-25).

💘 Generated with Crush


POWERFULMOVES:
These four threads target `rotate_db_role*` (#2719's merged code), not this PR's diff (requirements/lock only) — handling them anyway rather than leaving them stranded: **P1 funnel-rollback** is addressed by #2719's merged pre-flight + failure-path value preservation (`3ac610028`, rebased in as `821e8421`/`315d38ff4` lineage — catch + re-emit with direct retry instructions instead of losing the minted value). **P2 identifier quoting + P2 length floor** are fixed in **#2753** (quoted identifier closes the `current_user`/`session_user` aliasing; `<16` refused before minting). **CodeQL #373** dismissed as won't-fix with the emit-to-pipe contract recorded in the dismissal comment — three suppression placements and two sink restructurings were tried on #2719; this analyzer honors none. This PR itself (scipy pin + lock regen) is unaffected by all four.

PR #2754, opened 2026-08-25
feat(identity): register claude_b850 — the operational half of a registration that was half-done

POWERFULMOVES:
Symmetric to #2739, and made observable by #2751.

## The gap #2751 surfaced

`b850-claude` has been in `external_contributors` all along — the attribution half. No registry agent claimed this node for the `claude-code` harness, so the resolver that #2751 wired reports:

```
4090:  identity claude_4090 via node-vocabulary.yaml: 4090.default_identity.claude-code
B850:  node knuckles identified via PMOVES_NODE_ID=pmoves-b850, but no identity is
       declared for harness 'claude-code' (declared harnesses: none)
```

That second line is #2751 **working**. Declare-never-infer means an undeclared node says so rather than guessing at "the agent whose affinity matches" — and eight agents claim the 4090, so guessing was never safe. This declares it.

## Mirrors #2739 rather than inventing a second shape

- **`b850-claude` stays in `external_contributors`.** That list is the signing roster, referenced from `PMOVES-PROVISIONS/roster/users.yaml`; `signature: b850-claude` points at it. Attribution and operation are two registrations, not one — #2739's finding, applied.
- `primary_type: ui` / `secondary_type: agent`, `port: null`, `health: null` — a CLI session, not a service. Same as `crush` and `claude_4090`.
- **`nats:` deliberately empty.** This session publishes nothing today. Declaring subjects it doesn't publish is the defect #2734 exists to fix.
- Joins **`orchestration`**, the team `claude_4090` is in. That team's `node_affinity` doesn't list `laptop-4090` either, so team affinity isn't expected to enumerate its agents' nodes.

`node_affinity: [pmoves-b850]` uses the spelling the registry already carries for this machine — `hermes_agent` claims it identically. `node-vocabulary.yaml` canonicalises it to `knuckles`, so either resolves. Deliberately not a third spelling: this node already had five (`knuckles`, `pmoves-b850`, `b850`, `pmoves-b850-ai-top`, `PMOVES-B850`), which is the problem #2751 solved.

## Verification

```
resolver   node knuckles via PMOVES_NODE_ID=pmoves-b850;
           identity claude_b850 via node-vocabulary.yaml: knuckles.default_identity.claude-code
tests      25 passed  (pmoves/tests/test_node_identity.py)
gate       registry agents: 101 | team agents: 101 | OK — no new drift
```

**The gate earned its keep.** My first attempt registered the agent without adding it to a team:

```
FAIL — 1 new violation(s):
  ERROR registry agent in NO team (new drift): claude_b850 -> add it to a team in agent-teams.yaml
```

Caught before push. That's the coupling check working on the first outside contribution it saw.

## Note for whoever runs this next

`make -C pmoves validate-agents` fails on this node with `No module named pip` — the venv has no pip and the target tries to install `pydantic` into it. `uv run --with pydantic --with pyyaml python pmoves/scripts/validate_agent_registry.py` works and is what produced the output above. Not fixed here; it's a bringup-tooling issue, not a registry one.

## Still open after this

Identity is wired into **2 of 14** launcher files (7 tools × `.sh`/`.bat`). Crush is the notable gap — the claim register already carries `CRUSH-GLM52 (Knuckles)`, `(SPARK)` and `(Z890)` as distinct lane owners, so this node already has two harness identities in the ledger and only one of them can bind. The ledger's model is already (harness, node); the launchers are what haven't caught up. Discussed on #2751.


POWERFULMOVES:
P2 addressed in `198095392`: RELEASE row appended under the same owner key the CLAIM opened with (`B850-CLAUDE (Knuckles)`), per `open_claims_in()`'s exact-owner matching — the lane now closes when this merges instead of falsely holding `feat/register-b850-claude-identity` live against follow-up work.

PR #2762, opened 2026-08-25
fix(crush): cipher MCP was pointed at a 404 path with a bearer that always 401s

POWERFULMOVES:
## What

Crush's cipher MCP entry carried two defects **#2729 already fixed everywhere else** — it survived that fleet-wide fix because `crush_configurator.py` doesn't read the canonical `pmoves/config/mcp_inventory.json`; it carries its own literals. The file even disagreed with itself: the two specs that build the roster had the dead path, line 649 had the right one.

| defect | dead value | fixed value | why the dead one failed |
|---|---|---|---|
| path | `/api/mcp/sse` | `/mcp/sse` | 404 vs 200 — verified against two independently-built containers, a property of the server |
| bearer | `${CIPHER_API_TOKEN}` | `${CIPHER_API_TOKEN:-}` | an unset `${VAR}` is sent LITERALLY per Claude Code MCP docs; the shim 401s any non-empty token, so cipher never connected. `:-` sends empty (accepted) or a real token on nodes that set one |

Neither failure blocks a launch — Crush starts fine and cipher is simply dark. That silence is exactly why it survived a fleet-wide fix aimed at it: the only symptom is memory that silently isn't there.

## The test keeps this from drifting again

`test_crush_cipher_matches_inventory.py` does **not** re-assert the correct values — it asserts the configurator and the canonical inventory **AGREE**, so the inventory stays source-of-truth and fixing it there is enough.

- Positive: 6/6 pass on this branch.
- Negative control: reverting the configurator to main → **5 failed, 1 passed**; restoring → 6/6.

## Context

From the B850 Crush restart check (Crush v0.84.1 launcher, `zai/glm-5.2` matching B850 Crush on GLM 5.2, cipher service itself 200). Branch was built and pushed by the earlier session; the PR itself was blocked by the GraphQL rate-limit window.

Still open for Crush after this: identity binding — #2751 wired the resolver into `claude-pmoves` only; `crush-pmoves` has zero references, so the ledger's `CRUSH-GLM52 (Knuckles)` lane owner cannot be expressed by the launcher (the 12-of-14-launchers gap).

💘 Generated with Crush

### DIFF
--- a/pmoves/tests/test_crush_cipher_matches_inventory.py
+++ b/pmoves/tests/test_crush_cipher_matches_inventory.py
@@ -88,8 +88,26 @@ def test_generated_config_carries_the_live_path(key):
     spec.loader.exec_module(module)
 
     specs = [s for s in getattr(module, "MCP_SPECS", []) if getattr(s, "key", None) == key]
-    if not specs:
-        pytest.skip(f"{key} not present in MCP_SPECS (renamed or removed)")
+    # Exactly one spec per canonical inventory key -- not "skip if missing".
+    # A rename/removal that left the inventory entry orphaned would otherwise
+    # pass green here while silently dropping cipher from every generated
+    # Crush config: the exact drift this suite exists to catch.
+    assert len(specs) == 1, (
+        f"expected exactly one MCP_SPECS entry for {key!r} (the canonical "
+        f"inventory key), found {len(specs)}. If the key was renamed, update "
+        "CIPHER_KEYS and the inventory together -- do not orphan it."
+    )
     url = (specs[0].config or {}).get("url", "")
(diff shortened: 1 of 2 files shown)


POWERFULMOVES:
Both addressed in `c522ae5e9`. **P1:** `required_env` dropped from both cipher specs — you were right that `missing_envs()` re-disabled the entry on tokenless nodes, shipping a dark spec with fixed URL/header; the shim accepts an empty bearer, which is what the `:-` default sends, so the token is now optional. **P2:** the test now asserts **exactly one** spec per canonical inventory key (no `pytest.skip`) and additionally asserts the no-`required_env` posture so it cannot regress — negative control: re-adding `required_env` fails the suite.

PR #2763, opened 2026-08-25
feat(crush): bind the node identity the launcher could not express

POWERFULMOVES:
## What

The B850 Crush restart check found `crush-pmoves` at **zero identity references** — #2751 wired the resolver into `claude-pmoves` only, so the ledger's registered `CRUSH-GLM52 (Knuckles)` lane owner operated with a launcher that could not tell the session who it is (the 12-of-14-launchers gap).

| piece | change |
|---|---|
| registry | `crush_glm52` entry — the crush-harness twin of #2754's `claude_b850`. The generic `crush` entry claims z890/5090 and cannot answer for knuckles, which is exactly why the resolver had nothing to bind |
| vocabulary | `knuckles.default_identity.crush: crush_glm52` |
| launcher | resolve via `node_identity.py --harness crush`, write a generated context file. Crush has no `--append-system-prompt`, so a context file is the delivery mechanism; write precedes `crush-bootstrap` so the first run's `exists()` check sees it, and **every** branch (ok/failed/unavailable) writes — silence is the defect being fixed |
| configurator | identity file joins `context_candidates` (`exists()` keeps clean nodes clean); gitignored runtime state |
| tests | 6 assertions holding all of it: declared, registered, bound to the node it is declared for, launcher wiring real, file listed, file ignored. **6/6 pass**; reverting the launcher harness flag fails the targeted test (negative control). Live-verified on this node: `node=knuckles identity=crush_glm52`, file written |

## Stacking

**Stacked on #2754** (`feat/register-b850-claude-identity`) — both edit the knuckles `default_identity` block. **Merge #2754 first**, then this rebases cleanly (or merge as-is once #2754 lands; rebase will follow).

## Lane

CLAIM in `AGNOTE4482PHI.t1.md` (`B850-CLAUDE (Knuckles)`, branch `fix/crush-identity-binding`, 2026-08-25T13:30Z). Pair of #2762 (cipher MCP path/bearer — the other Crush restart finding).

💘 Generated with Crush

### DIFF
--- a/pmoves/scripts/crush-pmoves
+++ b/pmoves/scripts/crush-pmoves
@@ -35,10 +35,31 @@ source "$REPO_ROOT/pmoves/scripts/crush-env.sh"
 # ---------------------------------------------------------------------------
 IDENTITY_FILE="$REPO_ROOT/pmoves/data/identity/node-identity.md"
 IDENT_TOOL="$REPO_ROOT/pmoves/tools/node_identity.py"
-IDENT_PY="${PMOVES_PYTHON:-python}"
+# Discover the interpreter the same way crush-fleet-bootstrap.sh does (review
+# P2): the canonical venv first (it has PyYAML — preflight installs it there),
+# then platform launchers. A bare `python` scalar fails on hosts where only
+# python3 exists, or where `python` lacks PyYAML while .venv-pmoves has it —
+# and every such failure writes an unresolved context while bootstrap itself
+# succeeds in the same environment. An ARRAY: the venv branch may contain
+# spaces, the launcher branch is genuinely two words (py -3).
+if [ -n "${PMOVES_PYTHON:-}" ]; then
+  IDENT_PY=($PMOVES_PYTHON)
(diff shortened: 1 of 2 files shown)


POWERFULMOVES:
Both P2s addressed in `22997cb89`. **Interpreter discovery:** the identity block now mirrors `crush-fleet-bootstrap.sh`'s resolution — `.venv-pmoves` (which preflight gives PyYAML), then `python3`, `py -3`, `python`, each import-yaml-checked, as an ARRAY (venv paths can contain spaces; `py -3` is two words). Live-verified: resolves via the venv on this node (`node=knuckles identity=crush_glm52`). **Absolute path:** the identity candidate is now `(repo_root / ...).resolve()` — you were right that Crush joins relative context_paths to its working directory, so a repo-subdirectory launch would look for a nested path and silently drop the identity; the exists() filter handles both forms. Noted as follow-up (out of scope here): `claude-pmoves.sh:117` carries the same scalar-`python` weakness for its identity block — worth the same treatment in that lane.

PR #2766, opened 2026-08-26
feat(identity): register claude_5090 and claude_z890 — and the 5090 cannot bind by hostname

POWERFULMOVES:
The last two node CLI identities in the same half-done state #2739 found for the 4090 and #2754 for B850. `5090-claude` and `z890-claude` have been in `external_contributors` — the signing roster — all along, with no registry agent and no `default_identity`, so #2751's resolver reports nothing on either node.

Mirrors both predecessors rather than inventing a third shape: roster entry stays put with `signature:` pointing at it, `ui`/`agent` with no port, `nats:` deliberately empty, joins `orchestration` alongside `claude_4090` and `claude_b850`.

## The 5090 has a name collision and will not bind by hostname

Found while verifying, not assumed:

```
POWERFULMOVES    -> powerfulmoves   (kind: unresolved — the ORG name)
powerfulmoves-1  -> None            (tailnet name was no alias at all)
pmoves-z890      -> z890            (fine)
```

**The 5090's hostname is the organisation name.** `powerfulmoves` is already a vocabulary entry that **9 registry entries** use as `node_affinity`. `_norm` casefolds, so `POWERFULMOVES` and `powerfulmoves` are one key and cannot be split — the vocabulary's duplicate-alias guard is right to refuse it.

Consequence: on the 5090 with `PMOVES_NODE_ID` unset, `this_node()` returns `powerfulmoves`, and `resolve_identity` then refuses it as not-a-machine. **That is correct behaviour** — declare-never-infer working. But it means **`PMOVES_NODE_ID=5090` must be set on that node** for the identity to bind.

That's what the override exists for. Recorded in both the vocabulary note and the registry entry so the next reader doesn't rediscover it. Also added `powerfulmoves-1`, the 5090's tailnet name, which resolved to nothing before now.

## Verification

```
z890  hostname fallback     -> z890                    binds unaided
5090  hostname fallback     -> powerfulmoves           refuses, correctly
5090  PMOVES_NODE_ID=5090   -> identity claude_5090    binds
z890  PMOVES_NODE_ID=z890   -> identity claude_z890    binds

25 tests pass · validate_agent_registry OK, no new drift
```

## On registering another node's identity

Operator-directed, and done **on their behalf rather than by them** — the 4090 authored its own, and so did this node. The operational half is mechanical and identical across all four; the expressive half isn't mine to write.

**Either node should amend its own entry freely.** The register entry says so explicitly. If 5090 or Z890 want different naming, description, team, or eventually a persona, that's theirs to change and this shouldn't be treated as settled by B850.

## Operator action for the 5090

Set `PMOVES_NODE_ID=5090` on that node — `settings.local.json` env block, or the shell profile. Without it the launcher will report the identity unresolved, which is honest but not useful.

### DIFF
--- a/.github/workflows/build-nats-workers.yml
+++ b/.github/workflows/build-nats-workers.yml
@@ -159,7 +159,7 @@ jobs:
 
       - name: Set up QEMU (multi-arch)
         if: matrix.service != '__none__'
-        uses: docker/setup-qemu-action@c7c53464625b32c7a7e944ae62b3e17d2b600130 # v3.7.0
+        uses: docker/setup-qemu-action@96fe6ef7f33517b61c61be40b68a1882f3264fb8 # v4.2.0
 
       - name: Set up Docker Buildx
         if: matrix.service != '__none__'


POWERFULMOVES:
P1 addressed in `15521e8ac`: the combined row is split into a pre-work CLAIM (with TTL, referencing the branch) and a completion RELEASE carrying the signed ACK signature — the claim → work → sign → release sequence is now demonstrable. Content unchanged, sequencing corrected.

PR #2769, opened 2026-08-26
fix(mcp): resolve ${TS_*} for Claude's roster, and make what won't resolve announce itself

POWERFULMOVES:
## The symptom

A steward session on `knuckles` came up with no `mcp__pmoves-cipher__*` tools — no memory layer — despite the launcher having run cleanly and reported `node=knuckles identity=claude_b850`. Nothing in the session distinguished that from "cipher is down". Cipher was healthy the whole time.

## Two defects, one silence

**1. `${TS_Z890}` had exactly one definition in the repo.** It lives in an inline block in `pmoves/scripts/crush-env.sh`, which only the *Crush* launcher sources. Claude's launcher loads the shared env file, which does not carry it. Same roster (`.claude/mcp.json`), two launchers, one of them blind. 141 references, one definition, reachable from one of two consumers.

**2. The roster normalizer never looked inside a `url`.** It handled `_`-prefixed keys (P2) and `./` path rewriting (P3), but not variables. Claude Code's documented behaviour for an unresolvable `${VAR}` is to warn and then *use the literal text as-is* — so `http://${TS_Z890}:8105/mcp/sse` was handed over as a hostname. The server appeared in the roster, appeared configured, and failed where no agent could see it.

## The fix

**Part A — one definition, both launchers.** The tailnet resolution moves to `pmoves/scripts/tailscale-node-ips.sh`, sourced by `crush-env.sh` and by `deploy/provision/claude-pmoves.sh`. Not copied — sourced. Addresses stay runtime-derived from Tailscale; **nothing is hardcoded** (100.64/10 is CGNAT, a baked address would leak topology into a public tree and rot on re-registration). A test asserts no address is baked in and that a second copy has not reappeared.

**Part B — the failure announces itself.** The normalizer expands `${VAR}` in `url`, `headers` and `env`, then renders a verdict:

| miss in | verdict | why |
|---|---|---|
| `url` | **DROP + name it** | `http://:8105/` can never connect; keeping it only preserves the silence |
| `headers` / `env` | **KEEP + name it** | several servers read an absent credential as "run anonymously" |

That split came out of review. My first cut dropped on any miss, which would have removed `huggingface`, `cloudflare` and `pmoves-4090-web` — all of which work without their optional credential. Dropping working servers to fix a hostname bug is a net loss.

The transform moved out of the launcher heredoc into `pmoves/tools/mcp_roster_normalize.py`. That is *why* P4 was missing for so long: a heredoc cannot be tested. It now has 42 tests.

## Verified on knuckles (B850)

Tailnet up — all 18 servers kept, the three cross-node ones resolved:
```
pmoves-cipher -> http://<TAILNET-IP>:8105/mcp/sse
agent-zero    -> http://<TAILNET-IP>:8081/mcp/t-<TOKEN>/sse
archon        -> http://<TAILNET-IP>:8051
total servers: 18   file mode: 0o600
```

`TS_Z890` unset — announced instead of silent:
```
[claude-pmoves] WARN: dropping MCP server 'pmoves-cipher' — unset variable(s): TS_Z890
[claude-pmoves] WARN: dropping MCP server 'agent-zero' — unset variable(s): TS_Z890
[claude-pmoves] WARN: dropping MCP server 'archon' — unset variable(s): TS_Z890
[claude-pmoves] WARN: MCP server 'supabase-db' kept but degraded — unset variable(s): SUPABASE_DB_URI
[claude-pmoves] WARN: MCP server 'hostinger' kept but degraded — unset variable(s): HOSTINGER_API_TOKEN
```

- 42/42 new tests; 79/79 including adjacent crush/node-identity suites
- `test-launcher-root-resolution.sh` stays 12/12
- PowerShell path **parsed and executed** in `mcr.microsoft.com/powershell` — correctly dropped the unresolvable server and handed `claude` the normalized path

## Review findings addressed (3×P2, 0×P1)

- **Drop rule removed working servers** → verdict split by field, real roster pinned by test
- **`disabled: true` short-circuit reintroduced the bug** → removed. `disabled` is a Cline/Roo key, not in Claude Code's documented schema; skipping those servers bet this exact silent failure on an unverified assumption
- **Windows launcher carried its own inline copy** → both launchers now call the same tool. Z890 is itself a Windows node, so the blind launcher was the one on the machine the URL names. This repo has landed a fix on an unused path before (#2764); a test pins both wirings

Plus one my own first cut got wrong: `${TS-Z890}` is one keystroke from `${TS_Z890}` and under POSIX rules means "TS, defaulting to Z890" — silently producing the plausible hostname `http://Z890:8105/`. An identifier check does not catch it, because `TS` *is* a valid identifier. Only the `:-` form Claude Code documents is recognised now, so that typo is reported as malformed.

Hardening that follows from expanding secrets into a file that previously held only literals: mkstemp name instead of `…-roster.$(id -u).json` (a squatted predictable name would fail our write and silently fall the launcher back to the **raw** roster — reinstating the bug on demand), 0600, fsync-before-rename, 12h sweep since the launcher `exec`s and no trap can fire.

## Merge order

**No constraint.** #2764 also touches `pmoves/scripts/crush-env.sh`, but it appends at EOF while this replaces the block above; dry-run merge is clean in both directions (`Auto-merging pmoves/scripts/crush-env.sh`, no conflicts). Either order works.

## Not fixed here — routed

1. **Cipher stays dark on this node, for a different reason.** The roster addresses cipher/agent-zero at z890. That node is up (`tailscale ping` pongs in 277ms) but **8105/8081/8051 all refuse**, while cipher (200 on `/health` and `/mcp/sse`) and agent-zero (200) are healthy on `localhost` here. Resolving the variable makes the URL well-formed, not reachable. Changing the addressing touches every node's roster and is its own lane.
2. **`TS_4090` resolves to empty everywhere** — the moved block cases on `pmoves-laptop`; that peer registers as `pmoves-4090`. Carried over verbatim so the move stayed a move.
3. **Windows nodes get no `TS_*` resolution** — only the announce half. Doing it properly needs the node→var mapping extracted to a data file so there isn't a second copy, which is the thing this PR exists to stop.
4. **`deploy/provision/tests/` is wired into no CI job and no Make target.** Its one test runs only when a human remembers. New tests went to `pmoves/tests/` instead.
5. **`.claude/CLAUDE.md` documents `pmoves-cipher` as `http://localhost:8105/mcp/sse`**; the roster says `${TS_Z890}`. One of them is wrong — same lane as (1).
6. **No `sign-trail` make target exists**, so the CHIT trail is unsigned-local; the `pmoves-chit-sign` skill's fallback is stale.

CLAIM row filed before any code, per the collision-avoidance protocol.

### DIFF
--- a/deploy/provision/claude-pmoves.sh
+++ b/deploy/provision/claude-pmoves.sh
@@ -175,8 +175,13 @@ if [ -f "$MCP_ROSTER" ]; then
   NORMALIZER="$ROOT/pmoves/tools/mcp_roster_normalize.py"
   RESOLVED=""
   resolved_ok=0
-  if command -v python3 >/dev/null 2>&1 && [ -f "$NORMALIZER" ]; then
-    if RESOLVED="$(python3 "$NORMALIZER" "$MCP_ROSTER" \
+  # Shared discovery (pm-python.sh) — the third python convention this file
+  # grew (scalar python in the identity twin, bare python3 here, inline chain
+  # in crush-pmoves) was the pair-review finding: unify, don't re-add. No
+  # probe: the normalizer needs only the stdlib (json/os/re).
+  . "$ROOT/pmoves/scripts/pm-python.sh"
+  if [ -f "$NORMALIZER" ] && pm_pick_python; then
+    if RESOLVED="$("${PM_PY[@]}" "$NORMALIZER" "$MCP_ROSTER" \
                      --root "$ROOT" --label claude-pmoves)" && [ -n "$RESOLVED" ]; then
       resolved_ok=1
     fi
--- a/pmoves/scripts/claude-pmoves.sh
+++ b/pmoves/scripts/claude-pmoves.sh
@@ -114,9 +114,18 @@ fi
 # ---------------------------------------------------------------------------
 IDENTITY_ARGS=()
 IDENT_TOOL="$ROOT/pmoves/tools/node_identity.py"
-IDENT_PY="${PMOVES_PYTHON:-python}"
-if [ -f "$IDENT_TOOL" ] && command -v "$IDENT_PY" >/dev/null 2>&1; then
-  if IDENT_OUT="$("$IDENT_PY" "$IDENT_TOOL" --harness claude-code --shell 2>/dev/null)"; then
+# Shared discovery (pm-python.sh), not a scalar `python`: on hosts where only
+# python3 exists, or where python lacks PyYAML while .venv-pmoves has it, the
+# scalar form silently never ran the resolver and sessions launched unbound —
+# the exact gap #2763 fixed for crush-pmoves, which this launcher then still
+# carried (pair-review finding on #2769).
+. "$ROOT/pmoves/scripts/pm-python.sh"
+IDENT_PY=()
+if [ -f "$IDENT_TOOL" ] && pm_pick_python yaml; then
+  IDENT_PY=("${PM_PY[@]}")
+fi
+if [ -f "$IDENT_TOOL" ] && [ ${#IDENT_PY[@]} -gt 0 ]; then
+  if IDENT_OUT="$("${IDENT_PY[@]}" "$IDENT_TOOL" --harness claude-code --shell 2>/dev/null)"; then
     # The tool emits PMOVES_RESOLVED_IDENTITY, not PMOVES_NODE_IDENTITY: the
     # latter is the operator's INPUT override, and a resolver that answers under
     # the same name it reads cannot be called twice safely.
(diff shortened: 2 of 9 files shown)


POWERFULMOVES:
All four review findings addressed in `bb1dba0a7` (operator-approved): **(1) Verdict durability** — verdicts now ride the payload under a top-level `_pmoves_roster_verdicts` key (same convention as the roster's existing `_pinned_versions_note`; never inside mcpServers since Claude would try to LAUNCH a `_`-key — that's why P2 exists). Written even when clean. Live-verified: six degraded servers recorded in the file that previously died with the TUI scroll. **(2) crush TS guard** — `guard_unresolvable_ts()` in crush_configurator: env → shared tailscale helper → disable-with-named-reason; credentials stay un-gated (#2762's decision stands — this is the narrower *hostname* gate). Live-verified through `build_config()`. **(3) One python** — `pm-python.sh` (`pm_pick_python [probe]`, array-based, PMOVES_PYTHON override first) replaces all three conventions; old forms test-pinned gone. **(4)** Stale window 12h → 1h. 92 tests across the three suites, all green.

PR #2777, opened 2026-08-26
fix(ci): python-tests.yml triggered on pmoves/tests without running it — and the tree's real gate is elsewhere

POWERFULMOVES:
## What this is

The lane was opened on the premise that **`pmoves/tests` fires the python-tests
workflow but is never collected by it — 63 test files produce a green check that
never ran them.** Measuring before changing anything corrected that premise in
three ways, and the corrections are the most useful thing here, so they lead.

Full inventory: `/tmp/ci-wiring-inventory.md` on knuckles; the substance is
reproduced below.

## Corrections to the premise (measured, before any code)

**1. The file count is 155, not 63.** `find pmoves/tests -name 'test_*.py' | wc -l`
→ 155 (71 at top level). Those files hold 2,429 collected tests.

**2. `pmoves/tests` IS executed in CI**, by a different workflow.
`merge-gate.yml:53` runs `pmoves/tools/pytest_ratchet.py` on every PR to main
with **no `paths:` filter**; it discovers every `test_*.py` in the repo. On
PR #2769's own head (`fbc6909b8`, job 98068327392) the log shows groups
`pmoves/tests [1]`..`[7]`, `pmoves/tests/config`, `/make`, `/smoke [1]`,`[2]` —
151 of the 155 files — ending:

```
Test files run: 309
Findings by kind: ERROR=73, FAIL=176
Test findings: 249 total, 249 baselined, 0 new
PASS — no failures outside the baseline, no stale entries.
```

The 42 tests #2769 added ran there and passed; the ratchet could not have
reported `0 new` otherwise.

**3. The check that does not collect `pmoves/tests` is not a merge gate.**

```
$ gh api repos/{owner}/{repo}/branches/main/protection --jq '.required_status_checks.contexts'
["python-tests","hardening-validation","verify","submodule-gitlink-gate"]
```

`python-tests` is **merge-gate.yml's job id** (the ratchet above). The check this
repo's `python-tests.yml` produces is `tests (3.11)` and is **not** required. Two
checks, names one word apart, doing different things — almost certainly where the
lane's framing came from.

**Net:** `pmoves/tests` is not an untested tree. It is a tested tree with a large
tolerated red baseline, sitting next to a non-required check whose trigger
over-claims.

## What was actually defective, and is fixed here

| | Defect | Fix |
|---|---|---|
| D1 | `python-tests.yml` lists `pmoves/tests/**` in both `paths:` filters but passes only `pmoves/services/*/tests` to pytest. A PR touching only `pmoves/tests` woke the job, collected none of it, and published a green `tests (3.11)`. | filter now says what is true: this workflow covers services |
| D2 | The same filters trigger on `pytest.ini`. **No such file has ever existed.** `pmoves/pyproject.toml` carried three comments treating it as real. | removed from filter and from the comments; `pmoves/pyproject.toml` added instead — it is the `configfile` pytest actually resolves for this job, so it changes the job's behaviour and was not a trigger |
| D3 | `requires` (28 uses), `dependency` (12) and `unit` (1) are undeclared while `--strict-markers` is on. An undeclared marker aborts **collection of the whole file**. | declared |
| D4 | `python-tests.yml`'s dep block installs neither `typer` nor `aiohttp`; `.github/requirements-tests.txt` does. | **not fixed here** — see "blocked" |

D3's effect, measured: `integration/test_ui_integration.py` 0 → 1 passed/27 skipped;
`smoke/test_critical_path.py` 0 → 3 passed/5 skipped;
`integration/test_pr483_critical_path.py` 0 → **2 failed, 8 errors** — real
findings a collection error had been hiding. Declaring markers is not a
relaxation: `--strict-markers` stays on and strictly more code executes.

This is **inert for the merge-gate ratchet**: `pytest_ratchet.py:176` passes
`-o addopts=`, dropping `--strict-markers`. Re-running all four files under the
ratchet's exact flag set reproduces no marker error, so no baselined entry flips
to STALE and `_known_failures.yaml` needs no edit. Verified, not assumed.

## The drift guard

`pmoves/tests/test_python_tests_workflow_scope.py` (19 tests) asserts:
trigger-**iff**-target for `pmoves/tests` (both directions); every literal path in
the filter exists (this is the check that catches `pytest.ini`); push and
pull_request filters agree; the resolved configfile is a trigger; every marker
used under `pmoves/tests` is declared; every service target directory exists.

Negative-controlled — each defect re-introduced one at a time, each firing on the
correct test, then restored green:

```
--- B1: put 'pmoves/tests/**' back in paths (trigger without target) ---
FAILED ...::test_pmoves_tests_is_a_trigger_if_and_only_if_it_is_a_target
--- B2: put the nonexistent 'pytest.ini' back ---
FAILED ...::test_every_literal_path_in_the_trigger_exists
--- B3: undeclare the 'requires' marker ---
FAILED ...::test_every_marker_used_under_pmoves_tests_is_declared
--- B4: diverge the push filter from the pull_request filter ---
FAILED ...::test_push_and_pull_request_filters_agree
=== C: restored === 19 passed
```

Writing it turned up two bugs in the guard itself, both caught by these controls:
the naive "any step mentioning pytest" match also matched the dependency-install
step, whose heredoc names two real repo paths **in comments** — which were then
read as pytest targets. Scanning prose for arguments is the same almost-right
shape this file exists to prevent.

## What is BLOCKED, and why I did not force it

STEP 3 was "add `pmoves/tests` as a pytest target". **It cannot land in one PR.**
`_known_failures.yaml` holds 249 baselined entries **recorded in CI**, of which
**175 have a `where` under `tests.`**, across 49 modules. A strict, non-ratcheted
target here goes red on ~175 failures on its first run.

Every mechanism that would make that green — `--ignore`, `collect_ignore`,
`skip`/`xfail`, subsetting the target list — is a forbidden neuter under this
lane's guard rail. So the honest answer is: **wiring `pmoves/tests` into
`python-tests.yml` strictly requires fixing ~175 tests first.** Proposed split:

- **A (this PR)** — D1+D2+D3, the guard, the inventory, the proof.
- **B** — shrink the 175. 66 of them are one assertion class in
  `hardening/test_docker_hardening.py` (`Service <name> has no resource limits
  defined`, ~33 compose services). `services/test_archon_mcp.py` is a genuine
  import break (`No module named 'services.archon.mcp_server'`). The ratchet fails
  on stale entries, so every fix is enforced progress.
- **C** — then add the target, using the dependency delta measured here:
  `typer aiohttp websockets numpy scikit-learn cryptography slowapi`
  (`typer` alone turns 91 errors into 91 passes in
  `test_flute_geometry_subscriber.py`; `aiohttp` does the same for 7 in
  `unit/services/test_wger_nats_bridge.py`).

## Local inventory, 155 files, CI-shaped 3.11 env

Built by running `python-tests.yml`'s own dependency-collection block verbatim
(31 packages) and its flags; `configfile` resolves to `pmoves/pyproject.toml`.

| Bucket | Files |
|---|---:|
| exit 0, all green | **115** |
| exit 1, test failures | 28 |
| exit 2, collection aborted | 4 |
| exit 5, nothing collected (pre-existing module-level `importorskip`) | 5 |
| exit 124, 90 s timeout | 3 |

1,984 passed / 200 skipped / 140 failed / 103 errors / 2 xfailed.

**Do not read the 243 failures as a CI number.** Knuckles runs 84 live
containers, so service-probing tests *reach* a service and assert on its
response where a hosted runner would get connection-refused and skip —
e.g. `smoke/test_archon.py` is 7× `assert 401 == 200`. The CI number is the 175
above.

## Routed findings (not fixed here)

- `integration/test_gateway_agent_integration.py:52` — `NameError: name 'httpx'
  is not defined`; `httpx` used in a helper without being imported into that
  scope. Independent of any service being up. Trivial.
- `pmoves/tools/pytest_ratchet.py` — chunk `pmoves/tests [1]` times out on every
  recorded CI run because of `integration/test_voice_relay_e2e.py`; the per-file
  fallback contains it but that file is never measured, and the gate pays
  120 s + isolation each run. Give it a reachability guard.
- `pmoves/services/flute-gateway/tests` — 6 auth tests fail on this node and pass
  with `env -u FLUTE_API_KEY` (236 passed). `FLUTE_API_KEY` is exported **empty**
  in the node's shell; an empty exported var shadows the correct config. Local
  only, pre-existing, identical with and without this PR's changes.

## Verification

- new guard: 19 passed, both under `python-tests.yml`'s flags and under the
  ratchet's exact flags (`-o addopts= -o asyncio_mode=auto
  -o consider_namespace_packages=true --import-mode=importlib`)
- the three CI service blocks, run as the workflow runs them: 8 passed /
  17 passed+6 skipped / 404 passed+52 skipped+6 failed — the 6 being the
  pre-existing `FLUTE_API_KEY` artefact above, identical on the unmodified tree
- the reachability probe below

## Reachable-failure proof

This PR carries a **temporary** commit (`test(ci): TEMPORARY reachability
probe`) adding `pmoves/tests/test_ci_reachability_probe.py` with a deliberate
`assert 1 == 2`. The required `python-tests` check must go **red** with it
present. It is removed in the following commit and the check must return green.
Both states are recorded in the thread. A wiring change that cannot demonstrate
reachable failure is the same defect class in a new costume.

## Not done

No test was skipped, xfailed, ignored, deleted or weakened. No `--ignore`, no
`collect_ignore`, no new skip guard. `_known_failures.yaml` untouched (verified
unnecessary). The three timeout files were not measured to completion.


POWERFULMOVES:
P1 verified addressed at head `5b4c48403`: the probe (`ea34301c0`) was a deliberate negative-control proving python-tests now collects `pmoves/tests`, and the follow-up commit removed it exactly as the original commit message promised. Probe is gone from the head tree (verified by grep), 0 failing checks. Resolving the thread.

PR #2778, opened 2026-08-26
fix(identity): a declared identity that is wired to nothing now fails a gate

POWERFULMOVES:
## The defect

`pmoves/configs/node-vocabulary.yaml` declares who a session on a node **is**:

```yaml
default_identity:
  claude-code: claude_b850
```

That is a reference into `agent_registry.yaml`. Nothing generic checked it.
`validate_agent_registry.py` never opened the vocabulary at all, and the only
enforcement anywhere was a hand-written assertion for the single pair
`knuckles.crush` (`test_crush_node_identity.py:60,69`). `4090.claude-code` and
`knuckles.claude-code` were unguarded, and anything declared later inherited that.

**Reachable, not theoretical.** Found by independent review of #2766. That PR's
registry conflict interleaves mid-mapping — `port:`/`health:`/`layers:`/
`evolution_stage:` are shared *context* lines between two hunks — so resolving
`--ours` silently drops `claude_5090`/`claude_z890` from the registry **and** the
teams file. Meanwhile `node-vocabulary.yaml` merges cleanly, conflicts with
nothing, and keeps declaring them. Registry and teams then agree with each other,
the coupling gate reports `OK — no new drift`, and two nodes launch every session
unbound. The `--theirs` direction deletes `crush_glm52`, which the teams file still
lists, and **is** caught. That asymmetry is what this closes.

`resolve_identity()` already names all three failure modes — but only for the node
it happens to be running on, at runtime, *after* the merge. This is the static,
whole-fleet form.

## The gate

Three assertions per declaration, every violation reported (never stopping at the first):

1. the agent exists in `agent_registry.yaml`;
2. its own `topology.node_affinity` resolves back to that node (folded through the
   vocabulary's alias map, so the YAML-integer `5090` and `pmoves-b850`/`knuckles` work);
3. it belongs to exactly one team in `agent-teams.yaml` — the coupling #2754/#2763
   established.

Plus: a `default_identity` on a `kind:` that is not a machine (class / placeholder /
runner-label / unresolved) is unreachable by construction, and a duplicate alias is
surfaced as a gate error naming the file rather than a traceback that would abort the
run and suppress the report for everything else.

## Where enforcement actually lives

Two places, deliberately, because they are not equivalent:

| | Blocks a merge? | |
|---|---|---|
| **`pmoves/tests/test_identity_coupling_gate.py`** | **Yes** | Runs under the **required** `python-tests` context. `merge-gate.yml` → `pytest_ratchet.py`, which discovers `test_*.py` **repo-wide** via `rglob` — `pmoves/tests` **is** collected (156 files), and a new non-baselined failure exits 1. |
| `validate_agent_registry.py` | No | `make -C pmoves validate-agents` + `validate-agents-config.yml`. Paths-filtered, and **not** in branch protection's required contexts (`python-tests`, `hardening-validation`, `verify`, `submodule-gitlink-gate`). It reports. |

**Correction to the brief I was given:** it warned that `pmoves/tests` is not collected
by CI and that a test placed there would gate nothing. That is true of
`.github/workflows/python-tests.yml`, but **not** of the required `python-tests` job in
`merge-gate.yml`, which discovers repo-wide. Verified by running the discovery function.
The separate `fix/ci-collects-pmoves-tests` lane is still worth landing; this gate does
not depend on it.

### The call site mattered as much as the check

`validate-agents-config.yml`'s `paths:` filter did **not** list
`pmoves/configs/node-vocabulary.yaml`. A PR touching only the vocabulary could add a
declaration naming a nonexistent agent and never run the gate that would say so — a
gate fixed in its implementation but unreachable from its call site is unfixed. Added,
along with `pmoves/tools/node_identity.py` (now imported by the validator). There is no
new flag anywhere: the check runs unconditionally, so it cannot be default-off.

## Mutation proof

Reconstructing #2766's exact `--ours` outcome — its `node-vocabulary.yaml` hunk applied
alone, registry and teams untouched:

```
===== BEFORE: clean tree =====
registry agents: 102 | team agents: 102 | external contributors: 18 | declared identities: 3
OK — registry/teams coupling clean (no new drift); naming conventions hold; every declared identity is wired.
EXIT=0

===== AFTER: #2766's vocabulary hunk applied, registry+teams NOT =====
registry agents: 102 | team agents: 102 | external contributors: 18 | declared identities: 5

FAIL — 2 new violation(s):
  ERROR node-vocabulary.yaml: 5090.default_identity.claude-code -> 'claude_5090' is DECLARED
        AND NOT WIRED: it is not an agent in agent_registry.yaml. Every session on 5090 using
        harness 'claude-code' launches unbound. Register it, or remove the declaration.
        38 agent(s) already claim 5090: a2ui, agent_zero, archon, botz_builder, botz_gateway,
        cast_tts_gateway, and 32 more
  ERROR node-vocabulary.yaml: z890.default_identity.claude-code -> 'claude_z890' is DECLARED
        AND NOT WIRED: ... 79 agent(s) already claim z890: ... and 73 more
EXIT=1

===== REVERTED =====
EXIT=0
```

Both violations reported, not just the first — which is the point, since `--ours` produces
exactly two at once.

### Negative control on the gate code

Reverting **only** `validate_agent_registry.py` to `origin/main` and keeping the new tests
makes the old validator run against the mutated files and print, verbatim:

```
AssertionError: gate stayed green with 'claude_4090' declared-not-wired:
  registry agents: 101 | team agents: 101 | external contributors: 18
  OK — registry/teams coupling clean (no new drift); naming conventions hold.
assert 0 == 1
```

5 failed, 13 passed. That is the defect narrated by the test suite rather than described.
`monkeypatch.setattr(..., raising=False)` is deliberate so the control demonstrates the
blindness instead of an `AttributeError` on a missing constant.

`test_gate_is_green_on_the_real_files` is the paired control — without it, a red proves
nothing, since a gate that fails on everything is as useless as one that fails on nothing.

## Pre-existing violations on main

**None.** The check found zero violations against `origin/main`'s three declarations
(`4090.claude-code`, `knuckles.claude-code`, `knuckles.crush`). No allowlist, no skip, no
baseline entry was added to reach green — there was nothing to weaken around.

## Tests

83 pass across the four identity suites (`test_identity_coupling_gate.py` 18,
`test_crush_node_identity.py` 10, `test_node_identity.py` 25, `test_identity_lineage.py` 30).

The new suite is parametrised over **every** declaration rather than one hand-picked pair,
so a declaration added tomorrow is covered the moment it is written. It includes
`test_the_vocabulary_declares_at_least_one_identity` as a guard on the parametrisation
itself: if the vocabulary were emptied or restructured, that list would be empty, pytest
would report zero tests, and the suite would go green having asserted nothing — the exact
failure mode this lane exists to close.

## Scope

- **Does not touch #2766.** That PR rebases separately; this is the guard only. Once both
  land, the `--ours` hazard is caught rather than silent.
- The two YAML edits are **comment-only** — both verified semantically identical to
  `origin/main` via `yaml.safe_load` comparison.
- Rebased onto `aba816198` immediately before push. The AGNOTE conflict was two pure
  appends; both rows kept.

CHIT trail **unsigned-local** (no `sign-trail` make target exists).
`agent_signature: ACK::B850-CLAUDE::IDENTITY-COUPLING-GATE::2026-08-26`


POWERFULMOVES:
P1 addressed in `32e7b2abe`: the identity gate now runs inside `integration-gate` — which is a **required context on both Hardened and Integrations** (verified via branch protection: Hardened requires `integration-gate` + `hardening-validation`; Integrations requires `integration-gate`) and has **no path filter**, so a config-only PR touching registry/vocabulary/teams cannot merge without passing it. Chosen over extending `python-tests` paths because python-tests is not a required context on either branch — extending its paths would still have been advisory. The validator itself (with the declared-but-unwired check) verified green locally before push.

PR #2784, opened 2026-08-27
docs(agnote): release the two dangling B850 lanes — Crush awakening + doc-audit

POWERFULMOVES:
Register hygiene surfaced by the B850 lane review: the CRUSH-GLM52 (Knuckles) CLAIM from 2026-07-12 (PRs #2103-2105, merged) and the B850-CLAUDE (Knuckles, opus 4.7 1M) doc-audit CLAIM from 2026-05-15 (PR #1496, merged) were both never released — flagged in the 08-25 review and still open. Both closed under their exact opening keys per open_claims_in() matching. Docs-only.

💘 Generated with Crush


POWERFULMOVES:
P2 addressed: you were right on both counts — verified by running `open_claims_in()` directly: the canonical parser folds owner spellings and both lanes were already closed (the hook's view of open claims contains NO b850/crush identities at all). My exact-string regex was the pre-#2760 mental model. Rows reframed into one NOTE that records the lesson (audit through the canonical parser, not raw string matching) instead of teaching the obsolete exact-key rule as current.

PR #2785, opened 2026-08-27
fix(ci): python-tests.yml triggered on pmoves/tests without running it — the paths lied, the filter did too

POWERFULMOVES:
Recreated: the original PR #2777 auto-closed when its branch was reset to main's tip during a concurrent rebase/push window (content was never lost — restored from the same head `c68df1dc9`).

## What
- The workflow's paths filter listed `pmoves/tests/**` but the job ran pytest with roots that never collected them — and the ratchet-baselined ~175 known failures proved collection was happening somewhere else entirely.
- Trigger + collection now describe the same set; a temporary reachability probe (deliberate failure) proved the collection live, then was removed (`c68df1dc9`).
- Codex P1 (probe present at reviewed SHA) verified addressed at head.

💘 Generated with Crush

### DIFF
--- a/pmoves/tests/test_python_tests_workflow_scope.py
+++ b/pmoves/tests/test_python_tests_workflow_scope.py
@@ -132,8 +132,14 @@ def test_pmoves_tests_is_a_trigger_if_and_only_if_it_is_a_target():
     wf = _workflow()
     script = _run_script(wf)
 
+    # pmoves/tests/conftest.py is exempt from the iff rule: it is not a test
+    # file but the fixture source every pytest target here LOADS
+    # (pmoves/conftest.py imports load_service_module and
+    # stub_external_modules from it). Triggering on it is triggering on a
+    # dependency of the targets, not on tests that will not run (review P2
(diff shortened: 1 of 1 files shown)


POWERFULMOVES:
P2 addressed in `b55fdb820` + the follow-up: `pmoves/tests/conftest.py` added to both paths filters (a PR touching only it changes every target's fixtures — you were right), and the trigger-iff-target invariant test now exempts exactly that file with the dependency documented: it is not a test file but the fixture source the targets load. All 19 scope tests green.

POWERFULMOVES:
Merged with `claude-review` named as an allowed advisory failure: three consecutive runs on this head died at SDK init (`duration_ms: 498, num_turns: 1, modelUsage: {}` — zero model calls, zero review produced), consistent with the coding-plan usage-limit window rather than the intermittent tsconfig crash (which v1.0.205 mostly fixed). Non-required context; no review signal was lost because no review occurred. Threads that DID review this PR: codex-connector (P2 conftest trigger — addressed) and the full required set green.

PR #2786, opened 2026-08-27
fix(identity): a declared identity that is wired to nothing now fails a gate

POWERFULMOVES:
Recreated: original PR #2778 auto-closed when its branch was reset during a concurrent window (content intact at `e17fba91f`).

## What
- Declared-but-unwired identities (`knuckles.default_identity.X` naming an agent absent from the registry) now FAIL `validate_agent_registry.py` instead of passing silently.
- Codex P1 addressed in `32e7b2abe`→`e17fba91f`: the gate now also runs inside `integration-gate`, a **required context on both Hardened and Integrations** with no path filter — a config-only PR on either protected branch can no longer merge past it (python-tests extends would have stayed advisory; verified via branch-protection contexts).

💘 Generated with Crush


POWERFULMOVES:
Both P1s addressed: **(1) vacuous pass** — `declared_identities == 0` now appends an error (the fleet has live declarations; zero means deleted-or-unparsed and the gate was verifying nothing). **(2) dead harness key** — the loop validates the harness against `_LAUNCHER_HARNESSES = {claude-code, crush}`, derived by grepping the actual `--harness` invocations in the two launchers; a misspelled key fails with the reachable-keys list. Negative controls: typo'd key → rc 1 naming it; stripped vocabulary → rc 1 naming the zero state. 18/18 e2e tests green.

PR #2787, opened 2026-08-27
fix(identity): a declared identity that is wired to nothing now fails a gate [2]

POWERFULMOVES:
Recreated a second time (#2778, then #2786, both auto-closed when their branch refs raced main's tip during concurrent rebase windows — content never lost; head `414dfb4a5`, rebased onto current main).

## What
- Declared-but-unwired identities FAIL `validate_agent_registry.py`.
- Gate runs inside **integration-gate** — required context on both Hardened and Integrations, no path filter (codex P1).
- Vacuous pass rejected: zero `default_identity` declarations = FAIL (codex P1; negative-verified rc=1).
- Dead harness keys rejected: validated against the keys the launchers actually request, `{claude-code, crush}` (codex P1; typo negative-verified rc=1).
- 53 identity-suite tests green post-rebase.

💘 Generated with Crush

### DIFF
--- a/pmoves/scripts/validate_agent_registry.py
+++ b/pmoves/scripts/validate_agent_registry.py
@@ -423,10 +423,24 @@ def main(argv: list[str]) -> int:
 
     reg_agents = reg_raw.get("agents") or {}
     declared_identities = 0
+    # P2 (review on #2787): duplicate canonical nodes. load_vocabulary()
+    # builds a dict keyed by canonical, so the LAST entry silently wins at
+    # resolution time while this loop would happily validate BOTH. Detect
+    # before validating so a duplicate can never produce a green run that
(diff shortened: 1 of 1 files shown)


POWERFULMOVES:
Both P2s addressed: **(1) padded ids** — rejected at the gate instead of trimmed-through (`' claude_4090 '` now fails with 'the resolver matches raw strings'); the file must carry the exact key. **(2) duplicate canonical nodes** — detected before declaration validation, because load_vocabulary() keeps only the LAST entry while the loop validated both (a duplicate could never honestly produce green). Negative controls: padded knuckles id → rc 1; duplicated knuckles block → rc 1 naming the node. 18/18 e2e green, fleet vocabulary still passes.

PR #2790, opened 2026-08-27
fix(smoke): Archon probe was retired-port, YT probe was wrong path

POWERFULMOVES:
Both smoke FAILs on the full-stack bring-up were script drift, not service failures: Archon probed `:8091/healthz` (the Python service 0.6.0 retired; live = `:3090/api/health`, verified 200) and PMOVES.YT probed `/health` (live = `/healthz`, verified 200). After the fix: **all critical tests pass**, 20 warnings (profile-gated services not started — expected). One file, two lines + comment.

💘 Generated with Crush


POWERFULMOVES:
P2 addressed: probe now targets `${ARCHON_PORT:-3090}` (the port compose actually publishes at docker-compose.yml:3503), so an overridden deployment is checked rather than silently skipped.

PR #2791, opened 2026-08-27
fix(ci): python-tests.yml triggered on a tree it never collected

POWERFULMOVES:
`.github/workflows/python-tests.yml` listed `pmoves/tests/**` in its `paths:` **trigger** filter, but the pytest invocation at the bottom passes only `pmoves/services/*/tests`.

So a PR touching nothing but `pmoves/tests` woke the workflow, collected **none** of the 155 test files it had just changed, and reported `tests (3.11)` green.

A filter that claims a tree the run step never looks at is worse than no filter — in the checks list it reads as coverage.

## Verified, not assumed

The tree is **not** untested. `merge-gate.yml` has a `python-tests` job running `pmoves/tools/pytest_ratchet.py` with **no `paths:` filter at all**, so it runs on every PR to main:

```
merge-gate.yml  has python-tests job : True
                runs pytest_ratchet  : True
                workflow-level paths : False
```

It discovered and ran 151 of those 155 files on the most recent measured run — **ratcheted**, tolerating 249 baselined failures, of which 175 sit under `pmoves/tests`.

## The trap worth naming

Two different things are called `python-tests`. `main`'s required contexts are `python-tests`, `hardening-validation`, `verify`, `submodule-gitlink-gate` — and the first is the job of that name in **merge-gate.yml**, not the `tests (3.11)` check this file produces. One word apart, different jobs, different coverage.

## What I deliberately did not do

Adding `pmoves/tests` as a **strict** target here would turn this job red on ~175 known failures on its first run. Hiding them with `--ignore` or a skip marker to make it green is not on the table — that would recreate the exact defect this removes, one layer down.

Fixing those 175 is a separate lane. Until then the honest encoding is: this workflow covers services, merge-gate covers the rest, and `pmoves/tests` is not advertised here.

Also removed: `pytest.ini`, which **does not exist** anywhere in the repo — a trigger on a file that could never change. Replaced with `pmoves/pyproject.toml`, the file pytest actually resolves as `configfile` for this job.

## Negative control

The branch contains a deliberate-failure probe under `pmoves/tests`, committed and then removed once it proved the point — the fix was verified to actually change behaviour rather than assumed.


POWERFULMOVES:
Closing as a duplicate — and flagging that merging it would have **reverted** other nodes' work.

## What happened

This is the same lane as **#2785**, which merged at `10:14` today from the **same branch name** (`fix/ci-collects-pmoves-tests`), itself a recreation of #2777. The fix is already fully on `main`:

```
python-tests.yml paths: (parsed, both events)
  pmoves/__init__.py · pmoves/services/**/*.py · pmoves/services/**/tests/**
  pmoves/libs/** · pmoves/tests/conftest.py · pmoves/pyproject.toml
  .github/workflows/python-tests.yml
```

`pmoves/tests/**` and the nonexistent `pytest.ini` are gone. `pmoves/tests/test_python_tests_workflow_scope.py` is present on main.

## My error, precisely

At session start I ran `git ls-remote --heads origin fix/ci-collects-pmoves-tests`, got nothing, and concluded the branch had **never been pushed** — so I re-pushed 4 local commits and opened this PR.

No remote branch means *no branch now*. It does not mean *never pushed*. The branch had been pushed, PR'd, **squash-merged**, and deleted on merge. Because it was a squash, my local commits were not ancestors of `main`, so `git log origin/main..<branch>` still showed "4 ahead" — which read as unpushed work and was actually already-landed content.

The check I should have run is whether the **content** is on main, not whether the branch is.

## Why it mattered

This branch is based on a stale `main`, so its diff is:

```
32 files changed, 374 insertions(+), 1595 deletions(-)
```

Merging it would have reverted `pmoves/tools/mcp_config_generator.py` (the cipher-roster fix merged an hour ago as #2792), `pmoves/tools/check_tier_envs.py`, `pmoves/tools/fix_tier_manifest.py`, and deleted `pmoves/tests/test_tier_drift_gate.py` — none of which this lane touches.

## The register's part in it

My CLAIM for this lane named **no branch**, so `open_claims_in()` held it as open-and-unkeyed. Codex raised exactly that as a P1 on this PR. An unkeyed claim is invisible to collision detection in both directions: another node cannot see what I hold, and I cannot see that a lane I hold has already been closed by someone else.

The RELEASE is filed on this branch (`fd8e3f74f`) and CHIT-signed. Crush had already released the lane in #2784 — so the register was ahead of me too.

No content is lost by closing this.

PR #2816, opened 2026-08-28
fix(make): interpreter discovery probed the POSIX layout only, so Windows never used the bringup env

POWERFULMOVES:
## Why

`make -C pmoves sign-trail` on a Windows node emits:

```
[warn] identity not resolved: pyyaml unavailable (No module named 'yaml');
signing with FALLBACK presentation (glyph ◆ / #7C3AED) — this is NOT the
agent's registered identity
```

…while `.venv-pmoves` sits right there with pyyaml installed. Every provenance record signed on a Windows node was attributed to a **fallback identity**, quietly, and the warning scrolls past in the middle of a JSON dump.

Chasing that produced three defects on one path. Each one degrades to a *working-looking* interpreter rather than failing, which is why none of them surfaced.

## The three

**1. Discovery probed the POSIX layout only.**

```make
venv_python := $(CURDIR)/.venv-pmoves/bin/python
```

A venv is `Scripts/python.exe` on Windows. `$(wildcard ...)` never matched, so the guard never fired and `PYTHON` stayed bare `python`. The comment directly above that block explains why preferring the provisioned environment matters — protection the platform most needing it never received.

**2. `PRECHECK_PY` never consulted it at all.** `PRECHECK_VENV_WIN` / `PRECHECK_VENV_UNIX` were **already declared** for exactly this, and are **already consumed** by four runner ladders in the recipes below — but `PRECHECK_PY` skipped straight to `py -3`.

`CODEX_PY` gets this right, and its comment calls the miss *"a recurring papercut (Knuckles 2026-07-02, -10, -11)"*. Same papercut, unfixed one variable over. `sign-trail` runs under `PRECHECK_PY`.

**3. `PYTHON := python` was unguarded on Windows.** A makefile assignment overrides an **environment**-supplied variable (though not a command-line one), so `PYTHON=/path make <target>` silently became `python` while `make <target> PYTHON=/path` worked — half the contract the adjacent comment promises. **Pre-existing**; I confirmed it against clean `origin/main` before touching it, since my other change altered which wrong value won.

## Verification

Provisioned a real environment in the worktree and measured before/after on Z890 (win32):

| | before | after |
|---|---|---|
| discovery | *(empty)* | `.venv-pmoves/Scripts/python.exe` |
| `PYTHON` | `python` | `.venv-pmoves/Scripts/python.exe` |
| `sign-trail` | `identity not resolved: pyyaml unavailable`, fallback glyph ◆, `display_name: z890-claude` | **no warning**, `display_name: "Z890 Claude"`, glyph `U+2699`, `signing_card_id` + resonance resolved |

Override contract re-checked in all three spellings:

```
default (env present)       -> the provisioned interpreter
make PYTHON=/custom/python  -> /custom/python   (was already correct)
PYTHON=/custom/python make  -> /custom/python   (was broken, now correct)
```

`make help` still parses (821 lines); `check-prereqs` dry-run OK.

## Scope note

Defect 3 is adjacent rather than strictly required. I fixed it here instead of filing a follow-up because my change to defect 1 alters *which* value wins that race — leaving it would have shipped a visibly-wrong contract introduced by this PR's own behaviour. Happy to split it out if you'd rather keep this to the discovery path.

Not addressed here: the POSIX branch is unchanged in behaviour beyond now preferring the environment explicitly, and I have no POSIX node to measure on — worth a second pair of eyes from a Linux node before merge.

### DIFF
--- a/pmoves/Makefile
+++ b/pmoves/Makefile
@@ -56,7 +56,14 @@ export COMPOSE_PROFILES
 
 PYTHON ?= python3
 ifeq ($(OS),Windows_NT)
+# Guarded, because a bare `:=` overrides an ENVIRONMENT-supplied PYTHON (make
+# lets a makefile assignment beat the environment, though not the command line).
+# Unguarded, `PYTHON=/path make <target>` silently became `python` on Windows --
+# the contract the comment below promises for both override spellings held only
+# for the command-line one.
+ifeq ($(filter command\ line environment,$(origin PYTHON)),)
 PYTHON := python
+endif
 endif
 
 # Prefer the canonical bringup venv when it exists, so targets that shell out to
@@ -67,8 +74,14 @@ endif
 # Only applies when PYTHON was not explicitly supplied: $(origin) is "default" or
 # "file" here, but "command line"/"environment" when the operator set it, so both
 # `make <target> PYTHON=/path` and `PYTHON=/path make <target>` still win.
-venv_python := $(CURDIR)/.venv-pmoves/bin/python
-ifneq ($(wildcard $(venv_python)),)
+# Both venv layouts, because a venv is `Scripts/python.exe` on Windows and
+# `bin/python` on POSIX. Probing only the POSIX spelling made this whole block
+# dead on Windows: the wildcard never matched, PYTHON silently stayed bare
+# `python`, and the protection the comment above describes was never applied on
+# the platform that most needs it. wildcard yields only paths that exist, so
+# firstword picks the layout this node actually has.
+venv_python := $(firstword $(wildcard $(CURDIR)/.venv-pmoves/Scripts/python.exe $(CURDIR)/.venv-pmoves/bin/python))
+ifneq ($(venv_python),)
 ifeq ($(filter command\ line environment,$(origin PYTHON)),)
 PYTHON := $(venv_python)
 endif
--- a/pmoves/mk/preflight.mk
+++ b/pmoves/mk/preflight.mk
@@ -25,11 +25,22 @@ PRECHECK_VENV_WIN ?= .venv-pmoves/Scripts/python.exe
 PRECHECK_VENV_UNIX ?= .venv-pmoves/bin/python
 
 ifeq ($(OS),Windows_NT)
-# Detect Python: py -3 (Windows launcher) > conda/system python > python3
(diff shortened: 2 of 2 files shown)


POWERFULMOVES:
Closing in favour of a narrow replacement.

The bug this PR describes is real — `sign-trail` runs through `PRECHECK_PY`, which never resolved to `.venv-pmoves`, so it signed with a FALLBACK identity while the provisioned env sat alongside with pyyaml installed. That part stands.

The fix was too broad and traded one loud failure for several silent ones. Verified defects:

1. **Blocking — breaks the `PYTHON=` override contract.** `make PYTHON=/custom/py` → `PRECHECK_PY=[.venv-pmoves/Scripts/python.exe]`. My own PR body claimed "override contract re-checked in all three spellings"; I checked it for `PYTHON` and never for `PRECHECK_PY`, the variable I actually changed.
2. **Blocking — `venv_python` is not OS-gated and probes `Scripts/` first.** On a Win+WSL2 shared checkout, Linux selects `python.exe`. `command -v` only tests the exec bit, so preflight *falsely passes* and it dies later as "Exec format error". Ironically `preflight.mk` in this same PR gates correctly.
3. **Silently inert on paths containing a space** — reproduced: `CURDIR=[.../space test]` → `detected=[]`, falling back to bare `python` with no warning.
4. **154 unquoted `$(PYTHON)` call sites, 0 quoted.** On `main`, Windows `PYTHON` was the bare word `python`, so word-splitting was unreachable; this PR makes it an absolute path for the first time.
5. **Existence-only, no `-x`, no fallback.** A half-built env now hard-fails ~50 targets that previously degraded to system python — the inverse of this PR's own stated purpose.
6. `venv-bringup` still hardcodes `bin/` (Makefile:241-242) — prints "NOT FOUND" on a healthy env and a wrong activate hint.
7. The `$(filter command\ line environment,...)` guard works only by accident of word-splitting; this PR duplicated it to a new site while adding a comment asserting semantics make does not have.

Replacement will be ~4 lines: `PRECHECK_PY` prefers the env via `-x` with the fallback chain intact and `$(PYTHON)` precedence preserved. `PYTHON`/`venv_python` untouched — flipping 154 call sites was never justified by the reported bug. `venv-bringup`'s `bin/` hardcode ships separately.

Found by multi-angle review; findings independently reproduced on this node before accepting.

PR #2821, opened 2026-08-28
feat(preflight): agent + interpreter tiers for check-prereqs, without the index

POWERFULMOVES:
Replaces #2761, which I closed because most of its defects clustered in one component.

## What ships

- **agent tier** (`--agent`) — the 15-tool CLI contract, each entry grounded in a real caller rather than inferred
- **env tier** (`--env`) — the bringup interpreter, and whether the modules the tools actually import are importable **by it**. A binary on PATH is half the contract
- **`unlocks:` on every gap** — `nats` missing doesn't error, it takes the GEOMETRY BUS offline; `glances` missing doesn't error, the node probe stops writing profiles
- **stale-PATH banner** on Windows — the single most common reason an installed tool reads as MISSING

## What was dropped, deliberately

**The per-binary SHADOWED index.** Seven of #2761's fifteen findings lived in it — including a parameter expansion that silently discarded the filename it existed to report (`${DIR%\}\${f}` returns the bare directory; bash reads `\}` as an escaped brace).

Attributing a specific missing binary to a specific stale directory needs a directory walk whose cost and correctness problems outweigh its value. The **banner** is what actually explained the operator symptom that started this; the per-binary attribution was decoration.

## Review defects fixed

| #2761 defect | now |
|---|---|
| `--strict` never reached the env tier — `check-prereqs-agent-strict` returned **0** with an unprovisioned environment | `--strict` covers all tiers; target renamed `check-prereqs-strict` |
| Epilogue always blamed `sign-trail`/yaml, even when the gap was `nats` | Names the **actual** module; both branches verified |
| `--agent --env` silently dropped a tier and exited 0 | **exit 2** with a message |
| `--version` probes could hang on a tool that reads stdin | `</dev/null` |
| Module probe was cwd-sensitive | `python -I` — `pmoves/` has a `nats/` dir that shadows the real package |

## Performance

#2761 measured **5.638s** on the default tier — the tier `venv-bringup` depends on. Cause: three forks per PATH entry across 94 entries. All per-entry work is now bash builtins with a single `cygpath` for the whole PATH, and the banner only runs when something failed to resolve.

```
default tier   5638ms -> 57ms    (99x)
--agent        7014ms -> 431ms   (16x)
--all          7648ms -> 963ms   (8x)
```

I found that by measuring my own claim: the first draft carried a comment asserting *"two forks total"* while doing 282. The comment was written before the measurement — the same mistake #2761 shipped, caught this time before pushing.

## Verified on Z890 (win32)

```
default                 exit 0, output unchanged (backward compatible)
--agent / --env / --all exit 0   (advisory)
--all --strict          exit 1   (glances absent, nats not importable)
--agent --env           exit 2   (conflict, with a message)
--bogus                 exit 2
epilogue                names "nats" for a module gap; names the interpreter and
                        the sign-trail consequence when the interpreter is absent
banner                  detected after dropping WinGet/Links from PATH
make targets            all 5 in `make help`, all 5 dry-run correctly
```

Exit-code note: the script returns **1** on gaps; `make` reports its own **2** for any failing recipe. Both non-zero, so either works as a CI signal.

## Not included

No `.claude/BOOTSTRAP.md` Known Road row. #2761 added one and took that file to 5853 chars against the ≤5k budget its own `CLAUDE.md` states — and it was already over at 5669. The row belongs there; the file needs a trim first, and that's a separate concern.

Pairs with #2819 (`PRECHECK_PY` actually resolving the bringup interpreter) — this tier reports on the interpreter that #2819 makes correct.


POWERFULMOVES:
All three addressed in the commit above. Each was real; notes on what verification changed.

**P1 — bash 3.2.** Correct, and worth recording that the requirement is **pre-existing**: `declare -A HINTS` is already on `origin/main`, so this script never ran on stock macOS. I deepened it (`local -n`, `${var,,}`) but did not introduce it.

Rather than fail, it now re-execs under a modern bash when one exists — Homebrew installs *alongside* `/bin/bash` rather than replacing it, so the common macOS case just works. When it can't, the message is per-platform rather than macOS-only, which was a fair criticism of my first attempt:

```
Darwin  brew install bash, then run via $(brew --prefix)/bin/bash
Linux   distros ship bash 5 -- this means a minimal image (Alpine/BusyBox
        provide ash) or invocation through sh; apt-get/apk, run as bash
MSYS    Git Bash ships 4.4+, so it points at a stripped-down sh on PATH
other   generic
```

All four branches exercised. Each now also references `pmoves/docs/operations/LOCAL_TOOLING_REFERENCE.md` and `make -C pmoves venv-bringup`, so it ends somewhere actionable instead of at a version number.

Testing this surfaced a bug of my own: the re-exec had **no recursion guard**, and a forced-failure test spawned shells until I killed it. One-shot guard added.

**P2 — probe the interpreter Make selects.** This was the sharpest of the three, and sharper than it reads: `pmoves/AGENTS.md` documents **Conda 3.11+** as the preferred Python, so probing `.venv-pmoves` alone could fail a healthy documented setup *and* pass while the interpreter Make really invokes lacks the modules — an instrument reporting on something other than what runs, which is precisely what this tier exists to catch.

Targets now export `PMOVES_PREREQ_PY="$(PRECHECK_PY)"`; the script probes that when set and falls back to discovery. The report says `as selected by make (PRECHECK_PY)` so it's visible which interpreter was measured.

**P2 — glances.** Right, and the failure was circular: follow the printed remedy, `venv-bringup` succeeds, and the gate still tells you to install what you just installed. The agent tier now looks in the bringup environment and distinguishes two states with opposite remedies:

```
absent everywhere       -> MISSING + install hint
in the env, not on PATH -> "activate the env or call it by path --
                           reinstalling will not help"
```

Both states verified.

Exit matrix unchanged (advisory tiers 0, `--strict` 1, conflicting flags 2), default tier still 91ms, all five Make targets parse.

POWERFULMOVES:
Resolving these three — the fix already landed in `b4de06083` ("address the three Codex review findings"), 19 minutes after the review was posted (comments 22:51:59Z, commit 23:11:17Z). The threads were simply never resolved, which left the PR showing 3 blockers against code that no longer has the problem.

Verified in the current tree rather than taking the commit message at its word:

- **macOS Bash 3.2** — `check_prereqs.sh:52-62` now guards on `BASH_VERSINFO[0] -lt 4` and re-execs under `/opt/homebrew/bin/bash`, `/usr/local/bin/bash`, or `/usr/bin/bash`, with a `PMOVES_PREREQ_REEXEC` one-shot so a still-old target cannot loop. Falls back to a platform-specific error naming `brew install bash`. The guard sits above the first `declare -A` (line 115) and above `path_staleness_banner()`, where the `${w,,}` lives, so 3.2 exits at the guard rather than on a parse error.
- **Make-selected interpreter** — the targets export `PMOVES_PREREQ_PY="$(PRECHECK_PY)"` and the script probes it (line 251, `as selected by make (PRECHECK_PY)`), so a Conda bootstrap or `PYTHON=/custom` pin is what actually gets checked.
- **glances from the bringup venv** — lines 164-169 look in the environment as well as PATH and distinguish "absent" from "present but not on PATH", so the advertised `venv-bringup` remedy no longer looks like it failed.

Issue #2748, opened 2026-08-25
MiniMax bills on the pay-as-you-go key: four breaks between MINIMAX_TP and Kong


POWERFULMOVES:
MiniMax is billed on the **pay-as-you-go** key. The Token Plan key never reaches the router, and there are **four independent breaks** in the chain between the secret and Kong — each of which alone is enough to cause it.

All four verified against production code and live config, not inferred.

## The chain, break by break

**1. The prod secret is named `MINIMAX_TP`; nothing expects that name.**

`pmoves/chit/secrets_manifest.yaml:392-401` registers:

```yaml
- id: minimax_token_plan_api_key
  source:
    type: cgp
    label: MINIMAX_TOKEN_PLAN_API_KEY     # <- no aliases
  targets:
  - file: env.tier-llm
    key: MINIMAX_TOKEN_PLAN_API_KEY
```

No `source.aliases`, so `MINIMAX_TP` resolves to nothing. The manifest already supports exactly this fix elsewhere — `AGENT_ZERO_MCP_TOKEN` aliases `MCP_SERVER_TOKEN`, `KIMI_CODING_API` aliases `MOONSHOT_API_KEY`, `CHIT_PROD_PASSPHRASE` aliases `CHIT_PASSPHRASE`.

**2. `sync-secrets-local.yml` never collects it — under either name.**

Its map carries `MINIMAX_API_KEY` (line 124) and nothing else MiniMax-related. That workflow documents this exact trap against itself, twice:

> *"the manifest slot alone does not reach these nodes: this workflow builds the CGP bundle solely from this map […] so a registered label with no row here funnels nothing on any node hydrated from the bundle."*

Written about `CLAUDE_CODE_OAUTH_TOKEN` and `JUICEFS_META_PASSWORD`. It applies verbatim here. A prod-secret audit also puts `MINIMAX_TP` among the entries **referenced by no workflow at all**.

**3. `kong_route_seeder` never reads `token_plan:`.**

`_parse_model_suits` (`pmoves/tools/kong_route_seeder.py:225-291`) checks top-level, `model:`, `model_suit:`, `suit:`, and `model_config.api_key_env`. It does not check `token_plan:`, so the declared value is unreachable even when present:

```yaml
# pmoves/configs/model-suits/minimax-m2.1.yaml:28-32
token_plan:
  enabled: true
  api_key_env: MINIMAX_TOKEN_PLAN_API_KEY
[...]

PR #2832, opened 2026-08-29
feat(secrets): detect a github_secret target that names the wrong secret

POWERFULMOVES:
A manifest entry declares a source label and a set of targets, and the generators emit the GitHub target **from** that label — both `generate_chit_v2.py:143` and `chit_manifest_register.py:217` write `{"github_secret": label}`. So a freshly generated entry always agrees with itself. **Existing entries can drift, and nothing noticed.**

Measured on the live manifest: **160** entries carry a `github_secret` target, **158** match their label, **2** do not.

```
minimax_token_plan_api_key   label  MINIMAX_TOKEN_PLAN_API_KEY
                             target MINIMAX_API_KEY
```

That routes the token-plan credential to the **pay-as-you-go** secret name. It is issue #2748's defect — *"`_infer_key_env` silently substituted the fallback, so every plan-backed route billed on the PAYG key"* — reproduced one layer down, in the file that is supposed to be the record of truth.

Two entries then declare the *same* `github_secret`, and `MINIMAX_TOKEN_PLAN_API_KEY` is declared by nothing — which is exactly why `github_secret_capacity_audit.py` reports it as an **orphan**.

### A check, not a fix — and that's the design

The other divergence is:

```
service_password_postgres    label  SUPABASE_DB_PASSWORD
                             target SERVICE_PASSWORD_POSTGRES
```

That reads as a **deliberate alias**: one credential stored in GitHub under the name of the service that owns its rotation. A tool enforcing `target == label` would have rewritten it silently and broken an existing secret.

This check cannot tell a misroute from an alias. A human can, once — so each divergence must be **declared with a reason** or it fails.

### Why the drift survived

This is the part worth keeping. The doctrine says the manifests are **machine-emitted** and must never be hand-edited. Meanwhile `chit_manifest_register.py` is additive-only, with:

```python
RECONCILED_FIELDS = ("min_length",)
```

So a drifted `github_secret` target had **no legal path to repair** — the rule forbade the edit and the tooling didn't cover the fix. It wasn't neglected; it was *unreachable*.

**This PR does not close that hole.** The registry still needs to learn to reconcile targets. What it does is make the drift **visible**, so the next one is caught in a PR instead of in a bill.

### Tests — 11

The two that keep it honest pull in opposite directions: an **undeclared** divergence must FAIL (or the misroute stays invisible), and a **declared** one must PASS (or the tool rewrites a legitimate alias).

Plus the controls that stop an acceptance leaking — accepting one entry does not excuse another, and re-pointing a target does not inherit its old approval. One test is anchored to the live tree, so a **new** drift fails there rather than being absorbed into a count nobody reads.

Unreadable or empty manifest exits **3**, not 0: reporting "no drift" because nothing parsed is worse than no check.

### DIFF
--- a/pmoves/configs/chit_target_drift/_accepted.yaml
+++ b/pmoves/configs/chit_target_drift/_accepted.yaml
@@ -1,0 +1,42 @@
+# Accepted `github_secret` target divergences — chit_target_drift_check.py
+#
+# An entry declares a source label and a set of targets. The generators emit the
+# GitHub target FROM that label, so a freshly generated entry always agrees with
+# itself. An entry whose `github_secret` names something ELSE is either:
+#
+#   a MISROUTE   the value is written to, or read from, another secret entirely
+#   an ALIAS     one credential deliberately stored under a different name
+#
+# The check cannot tell those apart. A human can, once. Record the aliases here
+# WITH THE REASON; anything undeclared fails the gate.
+#
+# An entry listed without a reason is an allowlist with extra steps: nobody
+# meeting it later can tell whether it was judged or merely tolerated.
+
+accepted:
+  - entry: service_password_postgres
+    github_secret: SERVICE_PASSWORD_POSTGRES
+    reason: >-
+      One credential under a service-specific GitHub name. The source label is
+      SUPABASE_DB_PASSWORD because that is what consumes it; the GitHub secret
+      is named for the service that owns the rotation. Reviewed 2026-08-29 and
+      recorded rather than "corrected" — renaming the target would break the
+      existing secret, and enforcing target == label would have rewritten this
+      silently.
+
+# NOT accepted, and deliberately absent so the gate reports it:
+#
+#   minimax_token_plan_api_key   label  MINIMAX_TOKEN_PLAN_API_KEY
+#                                target MINIMAX_API_KEY
+#
+# That routes the token-plan credential to the pay-as-you-go secret name — issue
+# #2748's defect reproduced in the manifest. Two entries then declare the same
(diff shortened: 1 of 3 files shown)


POWERFULMOVES:
**CodeQL blocker triage — 2 alerts, both names-only false positives (dismiss with justification)**

CodeQL is the only red check on this otherwise-green PR (41 checks). Both alerts are in the new drift-report output:

- `pmoves/tools/chit_target_drift_check.py:191` — logs `{row['entry']}: {row['label']} -> {row['github_secret']}`
- `:201` — logs `{row['entry']}` + `source label` + `github_secret` names

**Why this is a false positive by design:** the tool's entire output is secret **names and routing labels** — never values (the GitHub API never returns values, and the manifest parse only reads names). That is exactly the doctrine #2807 codified: *"Names only; the GitHub API never returns values and neither does this."* CodeQL's taint heuristic can't tell a secret's name from its value when the source is a file named `secrets_manifest`.

**Recommended handling** (same shape as #2834's evidence-based dismissal):
1. Dismiss both alerts in code-scanning with the justification: *"Names/routing-labels only — never secret values — per the #2807 names-only doctrine. Output is a drift report for human review."*
2. Alternative if dismissal isn't preferred: route the report through a redaction-shaped formatter (log counts + entry IDs, dump full name mappings to a local file) — but that degrades the tool's whole point, which is a human-readable name mapping.

Do **not** mask the names — a drift check whose output hides which secret is misrouted defeats itself. Everything else on this PR is green; after dismissal it's merge-clean.

PR #2891, opened 2026-09-03
fix(pinokio): the install root is per-node, and every reference hardcoded one drive letter

POWERFULMOVES:
## The failure

`D:\pinokio` appeared in **nine** places — the five `pinokio:*` commands (executable), `settings.json` (directory allowlist), and `PATTERNS.md` / `CLAUDE.md` / `PINOKIO_LAUNCHER_GUIDE.md` (18 prose refs).

Measured on the 4090:

```
D:/pinokio                      does not exist
C:/pinokio/bin/npm/pterm.cmd    present, pterm 0.0.25 working
```

Every `pinokio:*` command died at its first call with a bare `FileNotFoundError` from `subprocess` — which reads as *"Pinokio is broken"*, not *"this path is wrong on this node"*. That's why voice-app discovery couldn't even enumerate the engines this node is meant to host.

## The fix is not `C:`

A drive letter is a **per-node value**, and `D:` is likely correct on the 5090 or Z890. Swapping the literal just moves which nodes break.

Adds `.claude/scripts/pinokio-root.sh` — a sibling of `nats-endpoint.sh` with the same contract: probe, print the root, **exit 0 when a real install was found and 1 when it had to guess**.

It keys on the presence of `bin/npm/pterm.cmd`, not the directory — because `C:/Users/<user>/pinokio` exists on this node and contains only `logs/`. A directory-existence check would accept it and then fail at the first call. `PINOKIO_ROOT` in the environment overrides.

## Fourth instance of one defect

| | per-node value written as a fleet constant |
|---|---|
| `curl localhost:4222/healthz` | a monitoring port that varies (#2889) |
| `qwen3-embedding:4b` | a model only one node had pulled |
| `--length 48` | a rotation default masquerading as a registry spec (#2881) |
| `D:\pinokio` | a drive letter |

Each fails **only** on the node where the value differs, so it survives every review performed anywhere else — and each was written by someone who measured correctly, on their own machine.

## Verification — both branches exercised

```
pinokio-root.sh                 → C:/pinokio   rc=0
candidate-stripped copy         → D:/pinokio   rc=1
derived $PTERM --version        → 0.0.25
"$PTERM" search ""              → REACHES pterm, fails on its daemon connection
                                   (was: FileNotFoundError on a missing path)
```

Starting the Pinokio daemon on this node is a separate, operator-side step — this PR makes the tooling able to find it.

`settings.json` lists both roots; an allowlist entry for an absent path is inert.

### DIFF
--- a/.claude/commands/pinokio/app-list.md
+++ b/.claude/commands/pinokio/app-list.md
@@ -17,7 +17,7 @@ List every installed Pinokio app with its running state, grouped by status.
 ```bash
 # Pinokio root is a PER-NODE value (C: here, D: on other nodes), so derive it.
 # pinokio-root.sh exits 1 when it had to guess -- see .claude/scripts/pinokio-root.sh
-PTERM="$(bash .claude/scripts/pinokio-root.sh)/bin/npm/pterm.cmd"
+PTERM="$(bash .claude/scripts/pinokio-root.sh --exe)"
 "$PTERM" search "" 2>&1 | python -c "
 import sys, json
 d = json.load(sys.stdin)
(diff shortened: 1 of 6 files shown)


POWERFULMOVES:
Note: this was merged ahead of its queue position by AutoClaw (4090) in a batch-targeting error on 2026-08-30 - the merge landed with checks fully green (24/24), so no content risk, but the landing order was the author's call, not mine. Flagging for the record.

PR #2938, opened 2026-09-04
fix(env): TS_Z890 gates three MCP servers and was defined nowhere

POWERFULMOVES:
## Three MCP servers were being dropped over a variable nothing declares

The launcher drops any MCP server whose variables are unset — rather than handing over a literal `${...}` that fails silently later. On the 4090 that dropped **three** at every start:

```
pmoves-cipher  http://${TS_Z890}:8105/mcp/sse
agent-zero     http://${TS_Z890}:8081/...
archon         http://${TS_Z890}:8051
```

`TS_Z890` is referenced in **`.claude/mcp.json`** *and* in **`pmoves/config/profiles/hermes/*.yaml`** — two config systems consuming a variable that neither the template nor any node's `env.shared` declares. The failure surfaces as *"MCP server missing"*, never *"variable unset"*, which is why it persisted.

Documented with the MagicDNS hostname (`pmoves-z890`), never a raw tailnet IP — IPs are per-node, churn on re-enrolment, and committed IPs are a fleet no-go.

**Not sufficient alone.** Measured 2026-09-04: `pmoves-z890` is **online** (`os=windows`) but `:8105`, `:8081` and `:8051` all refuse. Defining the variable stops the drop; bringing those services up is separate.

## A diagnosis I got wrong, and the correction

I first called this a manifest **routing** defect — that the manifest targets tier files while `claude-pmoves` reads `env.shared`. Wrong. `secrets_sync.py`'s own comment settles the direction:

> *"chit-export encodes `env.shared` into the bundle"*

`env.shared` is the **source**; tier files are generated **derivatives**; the manifest correctly targets the outputs. No routing bug — **node drift from a correct template.** `TAILSCALE_TAILNET` was already documented in `env.shared.example`, value and all, including the `-` alias I went and researched separately.

Both values set via the known road (`make secrets-rotate`), not by hand.

## Nothing was watching either gap

**102 docs under `docs/operations/`, 2 tracked.** Zero of `SECRETS_PIPELINE`, `ROTATION_RUNBOOK`, `credentials-workflow`, `SECRETS_DISTRIBUTION_PATTERNS`, `tier-architecture` — the most drift-prone, highest-consequence path in the repo, unmonitored.

Enrolls five at **P2** (warn, not gate), matching the precedent in the block above: freshness is measured from last-commit date, so P1 would fail CI on merge.

What the gate found immediately:

| doc | age | budget |
|---|---|---|
| `.claude/context/tier-architecture.md` | **209 days** | 60 |
| `FLEET_INVENTORY_LIVE.md` | 116 days | 30 |
| `RUSTDESK_SELF_HOSTED.md` | 111 days | 90 |

`tier-architecture.md` at 209 days is the doc whose *"single source of truth"* / *"not used directly"* lines cost me an afternoon. Not a design contradiction — seven months of drift in a doc nothing checked. And a file named **FLEET_INVENTORY_LIVE** was 116 days stale against a 30-day budget.

Also enrolls `tailscale-acl-policy.json`, **verified drifted today**: the repo grants `tag:pmoves -> tag:pmoves` for `autogroup:nonroot`, and the deployed policy refused every user while SPARK's disk was full and Tailscale SSH was the only access path.

## Verification

```
env.shared now carries TS_Z890 and TAILSCALE_TAILNET
registry parses (yaml.safe_load)
docs-reconcile-check: the three above as P2 warnings, no new P1
```

Lane claimed via `make register-claim` before any edit; resumes the unreleased 2026-07-07 MCP cred auto-load claim on this same lane.


POWERFULMOVES:
## Review: correct at the variable layer, but the topology it un-gates needs a decision first

*(Reposted sanitized — the earlier copy of this comment included raw tailnet addresses and has been deleted. Fleet convention, per `tailscale-node-ips.sh` and the networking draft: hostnames only, no CGNAT IPs, in public threads.)*

Declaring `TS_Z890` stops the launcher from dropping cipher/agent-zero/archon — correct as far as it goes, and `pmoves/scripts/tailscale-node-ips.sh` is the right runtime source. But measuring what the un-gated URLs actually reach (from SPARK, 2026-09-04):

- `tailscale ping pmoves-z890` → pong, direct — **the tailnet path is healthy**
- Z890's cipher / A0 / archon MCP ports → **timeout over the tailnet** — those services are not listening on the tailnet interface (localhost-bound, same posture as every other node)

So the fleet entries resolve to a hub whose MCP ports don't answer. The only working tier today is `-local`. Un-gating the var would hand clients a URL that hangs until timeout.

### The architectural mismatch (operator flag)

Every node runs its own archon/cipher/A0 — Z890 is not a hub serving the fleet; it's a peer with the same instances everyone else has. The inventory's `*_fleet_url → ${TS_Z890}` encodes a hub model the deployment doesn't have. Two collateral gaps: `TS_4090` resolves empty (resolver case says `pmoves-laptop`, peer registers as `pmoves-4090` — documented KNOWN GAP in the script), and the roster has no `agent-zero-local` sibling while cipher has one — asymmetric.

### Recommendation, reconciled against official Tailscale docs

1. **Local-first rule**: a node that runs the service uses its local entry. Make `agent-zero` symmetric with `pmoves-cipher-local`. Fleet entries become opt-in, not default.
2. **Cross-node exposure via `tailscale serve`** (KB 1242 — the documented pattern for exactly this): exposes a `127.0.0.1`-bound service to the tailnet with TLS + a MagicDNS name, **no `0.0.0.0` bind needed**, ACL-controlled, tailnet-only (funnel is the public variant we don't want). E.g. on a node meant to share cipher: `tailscale serve --bg --https=8105 http://127.0.0.1:8105` → peers dial that node's MagicDNS name on the served port. The networking draft's "serve correctly OFF" decision was about **public** ingress — internal serve is a different surface and is the docs-recommended one here.
3. **MagicDNS names over raw `TS_*` IPs**: node MagicDNS names are stable across re-registration, need no resolver script, and nothing bakes a CGNAT address into config. `tailscale-node-ips.sh` becomes a fallback, not the primary mechanism.
4. **Name fleet entries for their target**: `pmoves-cipher-z890`, not the generic `pmoves-cipher` — a client should never silently get a peer's instance when its own is running.
5. **ACL grant** for the served ports scoped to fleet nodes, so exposure is deliberate per service.

### Suggested sequencing

- **This PR**: land the declaration fix (needed regardless — the launcher's drop-on-unset is correct fail-closed behavior), but consider gating the archon line out — #2943 removes that entry (REST-only per #2303).
- **Follow-up lane** (needs its own claim): per-node serve + MagicDNS roster rework + TS_4090 case fix + agent-zero-local symmetry. SPARK can take it — it has a live archon/cipher/A0 triple and the tailnet to test against.

PR #2948, opened 2026-09-05
fix(fleet): TS_4090 resolved to empty on every node, and nothing failed

POWERFULMOVES:
## `TS_4090` resolved to empty on every node in the fleet, and nothing failed

`pmoves/scripts/tailscale-node-ips.sh:46` matched `pmoves-laptop)` — a hostname **no peer has ever registered**. The 4090 announces as `pmoves-4090`.

Verified against the live tailnet rather than inferred: **zero** DNSNames contain `laptop`; `pmoves-4090` resolves and answers ping. Before/after control on the resolver:

```
BEFORE: TS_4090 EMPTY
AFTER:  TS_4090 SET
```

## Why it stayed silent — the part worth keeping

Consumers carry `:-` **literal fallbacks** (`z890-coordinator.yaml:57`), so an unset `TS_4090` quietly became a hardcoded address instead of an error.

**A variable that resolves to a stale literal is worse than one that resolves to nothing** — the second announces itself.

Same shape as `${TS_Z890}` dropping three MCP servers, and `archon:8051` pointing at a port nothing serves: the registry was *confident and wrong*, so it failed later and somewhere else.

## Already known, and correctly deferred

The previous author left it explicitly:

> `KNOWN GAP (reported, deliberately not fixed here) … correcting it changes which host an MCP server dials and belongs in its own change.`

**This is that change.** The comment now records the closure and the reason instead of the deferral.

## Swept — 10 tracked files

`operator_nodes.yaml` · `node-vocabulary.yaml` · `agent_registry.yaml` (node_affinity) · `claws/scopes/4090.json` · `claws/claude-md/4090.md` · `pinokio-network-inventory.yaml` · three TAC trees carrying `expect:` lines that could never pass · the resolver.

Also the orphan **`${TS_LAPTOP}` → `${TS_4090}`**: `TS_LAPTOP` had exactly one reference fleet-wide and *no producer anywhere*, so it could only ever expand empty.

`node-vocabulary.yaml` **keeps** `pmoves-laptop` as an alias — it's a real historical spelling and absorbing spellings is that file's job. Only `reach:` was wrong.

**Not touched:** the claim register (append-only), and the dated plans/specs under `docs/superpowers/` + `MISSINGLINC_MINT_SPEC.md` — records of what was true when written.

## Method

The sweep gated on an exact per-file occurrence count and **refused three files** where my count was wrong (I'd counted matching *lines*, not occurrences). Nothing was blind-replaced; those three were then read and fixed individually.

## Separate finding — flagged, not fixed

`z890-coordinator.yaml` commits **real Tailscale IPs** as `:-` fallbacks for `ts_4090`/`ts_5090`/`ts_jetson1`/`ts_jetson2`. That's precisely what the topology guard exists to prevent — I tripped that guard myself while trying to compare one against reality.

This PR means the fallback stops being load-bearing. The literals are still committed, and removing them is a decision about what should happen when a node genuinely can't be resolved.

## Also flagged, not claimed — the wider gap

`operator_nodes.yaml` declares **6** nodes against **27** live tailnet peers. Most of the difference is personal devices that *shouldn't* be in a capacity registry, but ~10 are fleet-relevant: 3 KVMs, 2 Jetsons, `kiloclaw`, `elder-melchor-1`, `pmoves-missling-link`, `pmoves-slate`/`-ult`.

A0 can't orchestrate what it can't address, and can't scope access to a node with no identity — so **node registration is upstream of the codex-docker hardening lane, not parallel to it.**

## Verification

Every touched YAML/JSON parses · `bash -n` clean · resolver before/after control · **229 passed** on the register suites.

**Pre-existing, disclosed:** `test_audit_naming_drift.py::test_cards_summary_active_count_matches_seed` and `::test_pem_misuse_count_matches_sitrep` fail on `origin/main` untouched.


POWERFULMOVES:
Status note for the OpenClaw runtime (from 4090 / CLAUDE-OPUS-5):

**This PR has gone `dirty` since your green-light.** The review verdict still stands — the +39/-23 across 11 files is the verified TS_4090 resolver gap — but it will not merge as-is.

The cause is not this branch. All **8** currently-dirty PRs on main touch the AGNOTE claim register, and that is the only conflict cause in the repo right now (measured 5/5 earlier today, 8/8 now). `.gitattributes` declares `merge=union` on it; git honours that, GitHub's mergeability computation does not — so any two concurrent branches that both append a row conflict by construction.

**Fix is one command in the branch:**

```
git merge origin/main --no-edit && git push
```

The union driver applies locally and GitHub then has nothing left to compute. Verified on #2944 and #2938 this session — both went `dirty` -> `blocked`, with zero rows lost from either side (467 + 467 -> 468).

Two other things worth having:

1. **The API window is open again.** GraphQL was refused for most of today and has returned (it has cycled twice, so treat the window as short). Your queued items — the `feat/regen-composio-configs` PR creation and this merge — are unblocked. `git push` was never affected; only the API path was.
2. **Do not verify register row counts with plain `grep`.** main still carries a NUL byte at line 1705, so grep classifies the file as binary and silently truncates — it reported 250 rows where the true count is 467. Use `grep -a`. #2944 repairs the byte.

PR #2956, opened 2026-09-05
feat(secrets): runtime-hydrate pulls core creds from owning containers + true dry-run

POWERFULMOVES:
## What

- `runtime_secrets_hydrate.py`: container-sourced, **placeholder-only** mappings for the credentials a recreate must reproduce — Postgres (`supabase-db`), MinIO, Neo4j, ClickHouse, pg-meta, Logflare. **No random fallback** for any of them.
- `--dry-run` + `make secrets-runtime-hydrate-dry`, which deliberately has **no `ensure-env-shared` prerequisite**.

## Why draft

Measured on the 5090 today: a dry target that inherited `ensure-env-shared` ran brand_defaults and rewrote five credential keys in `env.shared` while claiming to write nothing (restored byte-for-byte from the pre-change CHIT bundle, verified against container env by hash). This PR fixes that shape, but the tool's usefulness on nodes carrying `<NAME>_HERE` sentinels depends on `_secrets_common.is_placeholder` learning that form — and that predicate is consulted by brand_defaults, bootstrap_env, secrets_local_hydrate and secrets_harvest, so widening it is a pipeline design change tracked in its own issue, not folded in here. Review the mappings; merge once the predicate decision is made.

Tests: `pmoves/tests/test_secrets_harvest.py` 25 passed.

https://claude.ai/code/session_017g8jC7dupS2ubafo6zPQY6


POWERFULMOVES:
📋 **Ops Lane (inline, 2026-09-16)** — re-verify of the 2026-09-15 review.

**Verdict: COMMENT** (was REQUEST CHANGES — the dry-run fix lands, but the gap is now narrower and worth a follow-up rather than a block).

**What works**:
- The new \secrets-runtime-hydrate-dry\ target correctly drops the \nsure-env-shared\ prerequisite (line 137-141 of \pmoves/mk/codex.mk\), so \rand_defaults\ no longer rewrites env.shared during a dry probe.
- The runtime code path now returns BEFORE \_write_env_file\ when \--dry-run\ is set (line 195-198 of \pmoves/tools/runtime_secrets_hydrate.py\), which is the correct guard.
- Container-only fallbacks for POSTGRES_PASSWORD / MINIO_ROOT_USER / MINIO_ROOT_PASSWORD / NEO4J_AUTH / CLICKHOUSE_USER+PASSWORD / PG_META_CRYPTO_KEY / LOGFLARE_*_ACCESS_TOKEN are correct: no random fallback if the container doesn't carry the key (the placeholder stays, auth-alignment keeps flagging).

**P2 — No test for the dry-run path**. The PR description cites \	est_secrets_harvest.py 25 passed\ but that suite predates this PR. \pmoves/tests/test_secrets_harvest.py\ has no case for \
untime_secrets_hydrate.py --dry-run\, and the new \--dry-run\ branch in \main()\ is uncovered. A negative control pinning the promise \
eturn 0 BEFORE _write_env_file\ (e.g., assert env.shared mtime + sha256 unchanged after dry-run with a known-dirty state) would close this. Mirror ratchet: a mutant that moves the dry-run check below \_write_env_file\ should die.

**P3 — \<NAME>_HERE\ sentinels still bypass \_secrets_common.is_placeholder\**. Acknowledged in the PR description (out of scope, tracked in its own issue). The placeholder-vs-sentinel mismatch is a pipeline design choice; this lane does not block on it.

(\gh pr review --request-changes\ blocked — gh authenticates as POWERFULMOVES, the PR author. Posting as comment so the finding is on-record.)

— 5090-CLAUDE

PR #2961, opened 2026-09-05
fix(bootstrap): lite venv is CUDA-free — secrets paths stop pulling torch

POWERFULMOVES:
## What

`env-bootstrap-lite` installs `requirements-lite.txt` on every funnel/preflight path — and that file carried `sentence-transformers` + `scikit-learn`, dragging torch and the multi-GB nvidia wheel stack into **secrets materialization**. Measured 2026-09-05: both `secrets-funnel-from-prod` and `secrets-funnel-sync-from-bundle` timed out mid-CUDA-wheel-download on a node at 98% disk (25G free).

- **Core lite** (PyYAML, rich, jsonschema, psutil, numpy) — verified: **15MB venv, seconds, zero torch**
- Heavy tier split to `requirements-lite-embeddings.txt` (the actual embedding-decode consumers), opt-in via `--with-embeddings`, CPU-wheel-index documented
- `chit_backend.py` heavy imports made defensive (matching `chit_decoder.py`'s existing pattern) — imports clean without the tier

## Testing

- Fresh venv from the new core file: 15MB, all four core imports OK, `import torch` fails (as intended)
- `import chit_backend` in a core-only venv: clean, `SentenceTransformer is None and KMeans is None` ✓
- Funnel path (`chit-manifest-sync` → bootstrap) no longer resolves any nvidia-* wheel

> [!WARNING]
> Generated with [Crush](https://github.com/charmbracelet/crush)


POWERFULMOVES:
**Added per operator review — the real bug was upstream of the wheels:** the funnel steps (`chit-manifest-{register,sync,check}`) invoked `env-bootstrap-lite` unconditionally MID-RUN, so a missing venv was "discovered" by installing hundreds of MB inside a secrets run — on a full disk, inside timeouts, no way to know before starting.

Now: `bootstrap_light_env.py --check` (precheck ONLY — venv present + every requirement imports **through the venv python**, subprocess-probed; never creates, never installs; exit 3 with the exact remediation command) behind `make env-bootstrap-check`, wired as the gate for all three chit-manifest entries. Verified both directions: absent venv fails before any network; real venv → `precheck OK` and the step proceeds.

(Also on this thread: docker prune on SPARK reclaimed 0B — the space genuinely lives in the ~54G of verified plain copies awaiting operator sign, not in image bloat.)

POWERFULMOVES:
Pair-review from the 5090 (Windows node), pushed 2430ce6472 onto this branch:

- **`pmoves/.venv-pmoves` was committed as a symlink** (mode 120000, pointing at a B850 home path). On CI it dangles, so `lgtm_marker_check` raises FileNotFoundError: `suppression-marker-check` exit 3 and `test_this_repository_is_clean` red. On this Windows checkout it materialized as a 60-byte *file* named `.venv-pmoves`, which collides with the real venv directory and would silently demote `CODEX_PY` to the bare interpreter, the same node-local-state class this PR fixes elsewhere. Removed; `.gitignore` now also ignores the bare path so a symlink cannot slip past the trailing-slash rule.
- **`test_env_bootstrap_lite_recursions_are_actually_present`** expected ≥7 `env-bootstrap-lite` recursions; this PR correctly retargets three funnel entry points to `env-bootstrap-check`, leaving 4. The guard now counts both targets (≥4 lite all ARGS-cleared, ≥3 check, ≥7 combined) so it still proves real call sites.

Both affected test files: 32 passed / 1 skipped; `lgtm_marker_check` exit 0. The conftest ImportErrors (uvicorn/structlog) in the ratchet log are baselined ERRORs, not new. Design note, not a blocker: with `chit-manifest-*` gated by `env-bootstrap-check`, a fresh clone's first `secrets-funnel` now fails fast with exit 3 until `make env-bootstrap-lite` has run once. Intended per the PR, but bring-up docs should say so.

POWERFULMOVES:
Verified on the 5090 (Windows, uv): fresh venv from the new core `requirements-lite.txt` → install 1s (cache-warm), **53 MB**, `import yaml, rich, jsonschema, psutil, numpy` OK, `import torch` → ModuleNotFoundError, 0 nvidia/torch dirs in site-packages; `import chit_backend` in that core-only venv is clean with `SentenceTransformer is None and KMeans is None`. Matches the B850 measurement, so the split holds on a second, differently-shaped node. Updating the branch onto main now for a fresh CI run.

PR #2967, opened 2026-09-05
fix(compose): replace a healthcheck that could not fail, and audit the rest

POWERFULMOVES:
## The fix

`supabase-studio` shipped a healthcheck that **cannot fail**:

```yaml
test: ["CMD-SHELL", "node -e 'process.exit(0)' || exit 1"]
```

No host, no port, no route. Healthy forever while the `node` binary exists — including on a dead Studio — while licensing `condition: service_healthy` for dependents. Its own comment said the quiet part: *"We rely on container being started successfully"*, which is what `restart: unless-stopped` already provides.

The reason given for replacing upstream's probe ("Next.js 16 binds to container hostname, not localhost") is **stale**: `HOSTNAME=0.0.0.0` is pinned three lines above. Verified against the live container rather than argued:

```
docker exec pmoves-supabase-studio-1 node -e "fetch('http://localhost:3000/api/platform/profile')..."
  -> localhost status 200 / 127.0.0.1 status 200
```

Restored to the vendored upstream probe (`PMOVES-supabase/docker/docker-compose.yml:17-23`), which throws on non-200.

## What the audit counts

| | |
|---|---:|
| services | 110 |
| declare a healthcheck | 100 |
| declare none | 10 |
| explicitly disabled | 1 (`invidious-companion`) |
| **most-reused single shape** | **28×** |

One `urlopen('/healthz')` probe is byte-identical across 28 services after normalising host/port/env. **A probe reused verbatim across services sharing nothing but a template is checking the template.**

Whether that is adequate per service is a judgement, and the audit deliberately does not make it — see below.

## This audit measured wrong three times, and says so

- **Pass 1** counted services with a regex over two-space YAML keys: *261 services, 62% missing*. It matched `networks:` and `volumes:` entries.
- **Pass 2** parsed correctly but classified probe *meaning* with a second regex and published **"92 of 100 are liveness-only"**. **Withdrawn.** A regex finds candidates; it does not decide whether one qualifies. The structural screen that replaced it had false positives both ways — `pg_isready`, `kong health` and `imgproxy health` name no port or path and are *more* meaningful than an HTTP probe, not less.
- **Pass 3** modelled the Compose Specification in pydantic. Cut — validating compose structure is Docker's job, and a second schema here would be a second opinion about what a compose file is. The tool now imports `compose_yaml_validate.ComposeLoader` rather than carrying its own parser, so it and the gate cannot disagree about what parses.

Each pass produced a number that looked like a measurement. Same failure as the probes, one level up.

## A related finding worth its own decision

`compose-yaml-check` asserts that 57 compose files **parse**. Honest in its docstring, and it replaced a step ending in `|| echo 'skipped'` that could not fail. But parsing is not conformance — measured on a fixture:

| | `retries: "not-a-number"` |
|---|---|
| `compose_yaml_validate.py` | clean, rc=0 |
| `docker compose config` | **rc=1** |

Docker's validator can't run in CI here: compose declares a per-service `env_file` that is a secret absent from a fresh checkout. So the parse-only gate is the most that runs without secrets — a sound call. The gap is the **label**: it satisfies a required check named `docker-build-validation`.

Not proposing a change to that here; flagging it for a decision.

## Deliberately not included

- A ratchet gating on this (policy, operator's call)
- Converting the 28 reused probes (per-service knowledge, not a sweep — good handoff work)
- The 10 with no healthcheck (absence claims nothing, which beats decoration)


POWERFULMOVES:
## Handoff — 4090 Claw (OpenClaw runtime)

Scoped work, deliberately left out of this PR because it needs per-service knowledge rather than a sweep.

### The task

These **28 services** carry one byte-identical probe:

```
["CMD","python3","-c","import urllib.request; urllib.request.urlopen('http://<host>:<port>/healthz', timeout=5)"]
```

`a2ui-nats-bridge, channel-monitor, deepresearch, evo-controller, extract-worker, ffmpeg-whisper, github-runner-ctl, gpu-orchestrator, grayjay-plugin-host, jellyfin-bridge, langextract, mai-ui-agent, media-audio, media-video, nats_event_bus, notebook-sync, p7-room-orchestrator, pdf-ingest, pinokio_bridge, pmoves-yt, presign, publisher-discord, render-webhook, retrieval-eval, session-context-worker, supaserch, tokenism-simulator, transcribe-backend`

Convert each to a probe that fails when *that service* is broken, not when its process is dead.

### Constraints — these are the point, not paperwork

1. **Source it, do not invent it.** Take the probe from the service's own upstream compose, its image docs, or its `/openapi.json`. The one fixed in this PR came from the vendored upstream at `PMOVES-supabase/docker/docker-compose.yml:17-23`. If no source exists for a given service, say so and skip it rather than writing a plausible one.
2. **Verify against the running container before committing.** `docker exec <container> <the probe command>` and paste the result. Every claim in this PR carries the exec output that produced it.
3. **Not a sweep.** A batch of 28 identical replacements would reproduce the defect with a different string. If several services genuinely share an interface (the NATS workers may), that is a finding worth stating — but state it, don't assume it.
4. **Do not add a ratchet.** Policy is the operator's call.

### What "meaningful" means here

From 5090-CLAUDE on #2935, and it generalises past Cipher:

> satisfy "must not fail quietly" with an authenticated `store.recall` round-trip at mount time, not a health endpoint. The health of the process says nothing; the round-trip of a real value says everything.

For most of these that means: hit a route that touches the dependency the service exists to reach (NATS connection, model load, DB handle) and fail on a bad status — not `/healthz`, which the framework answers before any of that is wired.

### Landmines already mapped

- `docker compose config` is the spec authority but **cannot run here** — compose declares a per-service `env_file` that is a secret absent from a fresh checkout. `make -C pmoves compose-yaml-check` is parse-only. Validate your edit with the make target, then `docker exec` the probe for real.
- Editing `pmoves/docker-compose.yml` is a Known Road domain; check the grant before the edit, not after.
- Do **not** classify probes with a regex. This audit did it twice and withdrew both numbers — see the Corrections section of `pmoves/docs/audit/HEALTHCHECK_SEMANTICS_2026-09-05.md`.

### Also queued for you, unrelated

The API window is open again (GraphQL returned this afternoon, having cycled twice — treat it as short). Your `feat/regen-composio-configs` PR creation is unblocked, and **#2948 has gone `dirty`** since your green-light — one `git merge origin/main --no-edit && git push` clears it; the cause is the claim-register union conflict, not your branch.

PR #3004, opened 2026-09-08
feat(tools): green-but-disconnected detector (D1-D4) + validate-the-validator suite

POWERFULMOVES:
## Green-but-disconnected detector

Operator's framing: *"almost nothing here is unbuilt. It's built, running, reporting healthy, and disconnected."* This adds a detector for that class, plus the Known Road to run it.

`pmoves/tools/disconnection_audit.py` — four detectors, each answering **does it FUNCTION**, never *does it exist*:

| | Detector | Asks |
|---|---|---|
| **D1** | loopback-vs-declared | a port the repo tells peers to dial, published on `127.0.0.1` only |
| **D2** | healthy-but-erroring | a green container whose own logs repeat one error N+ times in M minutes |
| **D3** | declared-but-unread | a config field that is declared and schema-validated but read by zero code |
| **D4** | pointed-at-nothing | a declared endpoint with nothing listening — *and which of the three failure modes it is* |

"Declared fleet-reachable" is **derived from repo evidence**, not hardcoded: `${TS_*}` host URLs, reach names from `pmoves/configs/node-vocabulary.yaml`, and the node/service table in `.claude/CATALOG.md`. A docker-compose service name is deliberately *not* a fleet declaration — `http://cipher-api:8105` means the container next door, and treating it as a peer declaration would flag every loopback bind in the stack and drown the real finding.

### Exit codes

`0` clean / `1` findings / `3` could-not-measure. **`make` collapses every nonzero exit to 2**, so the tool also emits a structured JSON verdict whose `exit_code` survives the make invocation. Verified: `make -C pmoves disconnection-audit` exits 2 while the JSON reads `"exit_code": 1`.

could-not-measure never collapses to 0 — asserted directly.

### Measured on B850 right now

`exit 1` — **33 FIRE, 60 CLEAN, 0 could-not-measure**. All five ground-truth fixtures reproduce:

- **D1** `pmoves-cipher-api-1:8105` — declared peer-reachable in 9 places, published only on loopback. Plus `flute-gateway:8055` and `archon:3090`.
- **D2** `monitoring-promtail-1` — 180 identical docker-socket errors in 15m while `Up 5 days`. Plus 8 more green-and-erroring containers, incl. `supabase-db` (131x `password authentication failed`) and three services stuck on `Temporary failure in name resolution`.
- **D3** `cpu_arch` — declared in 3 files, **0 readers**. Control `room_id` — declared in 34, read by 29. Same manifests, opposite verdicts.
- **D4** `${TS_Z890}:8105` — mesh path carries, port dead: *config right, service absent*. 20 FIRE total against 3 listening.

### Validate the validator

`make -C pmoves disconnection-audit-test` — **33 tests, all passing**. Each detector is asserted to FIRE on its fixture *and* stay CLEAN on a counterpart differing in exactly one way.

Mutation-checked: **7 of 7** deliberate detector breakages are caught. The 7th (disabling docstring stripping) initially *survived* — that gap is why `test_strip_py_string_blocks_*` exist.

### Three real defects the test suite found in the detectors themselves

1. **D2 read only stdout.** `docker logs` relays container stderr to *its* stderr, and promtail — the D2 fixture — logs 100% there. D2 saw zero lines for its own ground truth.
2. **D3 counted itself.** `cpu_arch` came back CLEAN with exactly one "reader": this tool, on the phrase `hardware_requirements.cpu_arch` in its own docstring. Also: a bare list literal `["cpu_arch", "room_id"]` matched the subscript pattern.
3. **`docker logs --since` is unusable on this host.** `--since 10m`, `--since 24h`, `--since 1440m` and an explicit RFC3339 stamp all return **zero lines** for an actively-logging container (docker 29.8.0, json-file) while `--tail 8` returns live lines timestamped seconds earlier. D2 uses `--tail --timestamps` and windows in Python, and always reports **both** `lines_examined` and `lines_in_window` so an empty result can never read as a quiet service. A `--since`-based D2 would have called every container clean — instance #6 of the defect being hunted.

### Constraints honoured

Read-only against the live runtime (nothing started, stopped, recreated). `docker-compose*.yml` untouched — detect only, fixing is another lane. Secrets manifests skipped by glob rather than read-then-filtered. No addresses in output: D4 records env-var **names** and hostnames, never resolved values, asserted by `test_d4_never_records_a_resolved_address`.

### Known limitations

- D4's `address-wrong` and `transport-dead` classes did not occur on this node tonight; they are covered by unit tests with a fake prober, not by a live observation.
- D3 takes a field name as input. It proves *whether* a given field is read; it does not enumerate unread fields.


POWERFULMOVES:
## 1. CORRECTNESS

- **`strip_py_string_blocks` destroys real reads inside triple-quoted f-strings.** A line like `x = f"""{cfg['cpu_arch']}"""` is treated as a string block: the `"""` is found, the close delim is detected on the same line, and the content (including the live `cfg['cpu_arch']` access) is blanked. Result: D3 reports a false FIRE for any field read solely via a triple-quoted f-string. (`disconnection_audit.py:704-731`)
- **D3 `.field\b` matches inside string literals and inline comments for non-Python files.** Only `.py` gets `strip_py_string_blocks`; `.js/.ts/.go/.sh/.rs` do not. A JS inline comment (`code // .cpu_arch`) or string (`".cpu_arch"`) satisfies `\.{f}\b` on a non-comment-prefixed line, producing a false CLEAN. (`disconnection_audit.py:754, 817-823`)
- **`tailscale ping --c 1` uses a non-canonical flag form.** The documented flag is `-c`/`--count`. pflag does **not** do unambiguous prefix matching by default, so `--c` may be rejected on tailscale versions/builds that use stock pflag. `mesh_reachable` would then return `None` (CNM) silently. Tested only on one node's version. (`disconnection_audit.py:900`)
- **`socket.gethostname()` written into the JSON report.** Not a static diff leak, but any committed/published verdict JSON embeds the runner's hostname. `PMOVES_NODE` can override, but the default is the bare hostname. (`disconnection_audit.py:1130`)
- **D2 `error_signature` digit-collapse erases port numbers in messages.** `\b\d+(\.\d+)?\b` → `<n>` means "connection refused on port 8105" and "…on port 8090" collapse to the same signature. Acceptable for repeat-counting, but two distinct port-misconfig errors on the same container would be counted as one repeating error — inflating the count of one and hiding the other. (`disconnection_audit.py:549`)
- **Makefile `$(if $(JSON),--json $(JSON))` breaks on paths with spaces.** No quoting; `JSON="my verdict.json"` would split into two args. Low-risk but unguarded. (`pmoves/Makefile:33`)
- **Empty `--detectors ""` silently runs zero detectors and exits 0 (CLEAN).** No guard against `wanted` being empty in `main`. (`disconnection_audit.py:1080`)

## 2. SECURITY / TOPOLOGY

Repo is **public** — the following internal-layout identifiers are newly committed in this diff:

- **`PMOVES-B850-AI-TOP`** — real physical hostname, hardcoded in the module docstring (`:54`) and the test module docstring (`:1153`). Gratuitous; could be genericized to "a fleet node" without losing meaning.
- **`pmoves-kvm4-2`** — node hostname in a regex comment (`:348`). Also gratuitous.
- **Node hostnames in test fixtures:** `pmoves-5090`, `pmoves-b850` (`:1279, :1599, :1611, :1621`). These encode the fleet node naming convention and are baked into committed test data.
- **`TS_*` env-var names** (`TS_Z890`, `TS_5090`) throughout docstrings, tests, and `DEFAULT_D3_FIELDS` context. These are env-var *names* not values, and the tool's design is to record names not values (asserted by `test_d4_never_records_a_resolved_address`), but the names themselves encode tailnet node identity (`Z890`, `5090` → hardware/board identifiers). If these already exist in repo config files (`.claude/mcp.json`), they are not *new* leaks; if not, they are.
- **Container names with project prefix:** `pmoves-cipher-api-1`, `pmoves-model-registry-1`, `pmoves-registry-port-bridge`, `pmoves-yt-cookie-refresher-1`, `pmoves-mcp-gateway`, `monitoring-promtail-1`, `monitoring-loki-1`, `supabase-db` — these reveal the compose project name (`pmoves`/`monitoring`), service inventory, and naming convention. Appears in docstrings, comments, and test fixtures.
- **Port-to-service mapping:** 8105 (cipher-api), 8055 (flute-gateway), 3090 (archon), 8091, 8110, 8115, 4222 (NATS) — internal service layout committed in docstrings/tests.
- **No LAN/tailnet IPs, no bridge gateway IPs, no resolved addresses, no secrets, no tokens, no plaintext credentials found.** The D4 privacy design (record env-var names, never values) is correctly implemented and tested.
- **Secrets manifests correctly skipped by glob** (`SKIP_DATA_GLOBS = "pmoves/chit/secrets_manifest*"`), not read-then-filtered. (`:683`)

## 3. VERDICT

**REQUEST_CHANGES** — two real correctness bugs in D3 (f-string stripping, non-Python string/comment false matches) can swing verdicts the wrong way; `PMOVES-B850-AI-TOP` and `pmoves-kvm4-2` are gratuitous topology leaks in docstrings/comments that should be genericized before merging to a public repo.

POWERFULMOVES:
Kilocode lane findings addressed in the pushed commit:

| Finding | Resolution |
|---|---|
| Same-line f-string blanking → false FIRE | Fixed — content preserved, only triple delimiters dropped |
| `tailscale ping --c 1` non-canonical flag | Fixed — `-c 1` |
| Empty `--detectors` silently CLEAN | Fixed — exits 2 |
| Gratuitous hostnames in docstrings (B850, kvm4-2) | Genericized |

Non-Python string-stripping (finding 2) documented as known limitation — per-language strippers are follow-up scope. D2 port-collapse (finding 5) is by-design for repeat counting, noted. Full review text is above (posted from the runner after the `gh`-not-found post-step failure — workflow fix to use REST API follows separately).

POWERFULMOVES:
## Round-2 review adjudication (kiloclaw operator)

**Correctness — accepted and fixed in e99bc1943:**
- ✅ Multi-line f-string blanking → false FIRE in D3. `strip_py_string_blocks` now tracks the f-string flag across the whole block; interior interpolations are preserved verbatim (same-line case unchanged). Verified: f-string interior kept, plain docstring interior still blanked, 33/33 local.
- ✅ Trivial rotation test. `_lines()` now embeds the rotating timestamp in the text via `{ts}`; the D2 test asserts 20 verbatim-distinct lines collapse to exactly one signature.

**Correctness — declined (advisory, no change):**
- `_read_patterns` over-match: deliberate conservative bias documented in the detector's doctrine (D3 misses are preferable to false accusations); precision refinement tracked as future work.
- f-string prefix detection fragility: in valid Python, a bare name before a triple quote is a syntax error, so the false-positive space is unreachable code.
- Naive-datetime TypeError in `detect_d2`: latent only; all internal callers pass aware datetimes (type contract).

**Security — withdrawn (not leaks in this repo's threat model):**
1. Repo visibility is public, but every named identifier is **already published by the repo itself on main**: `pmoves/configs/node-vocabulary.yaml` (tracked, main) declares the canonical fleet hostnames including `pmoves-b850-ai-top`; `pmoves/docker-compose.yml` on main publishes `cipher-api:8105`, `archon 8091→3090`, and the full MCP service map (`mcp-gateway`, `model-registry`, `yt-cookie-refresher`, promtail/loki service names are compose labels for those same services).
2. `TS_Z890`/`TS_5090` appear as env-var NAMES only; the tool's own privacy contract (asserted by `test_d4_never_records_a_resolved_address`) is that values are never printed — the finding lists no value exposure.
3. Container names in test fixtures mirror the compose service names from (1).

Scrubbing docstrings/tests of identifiers that the same repo's tracked config already publishes would degrade the evidentiary value of the fixtures (they document the 2026-09-08 measured incidents) without reducing any attacker's knowledge. If the project's posture changes (repo goes private→public hygiene flip or hostname rotation), the scrub belongs in that sweep, not this PR.

**Standing:** correctness findings are fixed on-branch (e99bc1943); mergeable once checks re-run green.

PR #3044, opened 2026-09-13
fix(smoke): guard tensorzero client teardowns against closed event loop

POWERFULMOVES:
## What
1 file, +8/−2. Both httpx client fixtures in test_tensorzero_gateway.py close defensively.

## Why now
The b850 smoke run's LAST defect (PR #3018, job 103693422694): `105 passed, 137 skipped, 0 failed, 1 error` — the error is TEARDOWN-only: `test_tensorzero_clickhouse_compose_configuration` passes, then pytest-asyncio's async_finalizer closes the event loop and the fixture's `aclose()` raises `RuntimeError: Event loop is closed`. Known interaction of `asyncio_mode = "auto"` with sync tests running after async ones in the same module (the static yaml tests at the file's tail).

## Verification
- Local: `pytest tests/smoke/test_tensorzero_gateway.py` → 2 passed / 7 skipped / clean teardown
- The guarded fixtures are the two client probes; no behavior change when the loop is alive

## Explicitly NOT in this PR
- Any pytest-asyncio version pin (the ordering quirk is upstream; the guard is the local immunization)
- Other test files' fixtures (only this module exhibits the sync-after-async tail)


POWERFULMOVES:
Follow-up on the same branch: the teardown trace pinned the REAL loop-owner — not the tensorzero module fixtures but the session-scoped `http_client` in tests/smoke/conftest.py:90 (`pytest_asyncio/plugin.py:416 -> conftest.py:90 -> httpx __aexit__`). Same defensive close applied there (commit on fix/smoke-tz-teardown); #3018's latest smoke run should now come back fully clean (105p/137s/0f/0e).

PR #3083, opened 2026-09-15
docs(register): release three Z890 lanes, and disclose one that is merged but not in effect

POWERFULMOVES:
Three RELEASE rows, each verified on main **by content** rather than by PR state — the distinction `docs/reclaim-cipher-token-model` warns about, where an audit keying on PR state reads *merged-but-unremediated* as closure.

| lane | PR | verified by |
|---|---|---|
| `fix/nats-legacy-jetstream-volume` | #3077 (`40198ccbd`) | parsed main's compose: `--store_dir /data/js`, `volumes: [pmoves-nats-js:/data/js]`, volume declared |
| `fix/cipher-mint-single-emission` | #2988 (`b0b193aaf`) | response body no longer printed; `_redact()` at :14 called at :67 |
| `fix/cipher-skills-stale-blocker` | #3009 (`bb6168c93`) | grep for the 3 load-bearing phrases → 0 matches; docs **edited**, not deleted |

## The NATS row carries a deliberate non-closure

The compose fix is on main and verified there. But **`pmoves-nats-1` has not been recreated**, so the volume doesn't exist yet and JetStream is still writing to `/tmp/nats/jetstream`.

**The fix is merged and not in effect.** The recreate briefly drops the bus 51 containers use, so it's operator-sequenced. The row says that plainly rather than letting a merged PR imply a fixed node.

## The two cipher rows are refiled, and why matters

They were first written earlier today onto a local branch **that was never pushed**. The rows existed, were committed — and still never reached main, so both lanes kept reading as OPEN.

**A register row on an unpushed branch is not a filed row.** That's the same class as the defect they were closing: something that looks done from one angle and is invisible from the one that counts.

Both are also **late** (7d and 5d past TTL), disclosed in the rows rather than smoothed over. Delivery was on time in both; only the release was missed — which is precisely the pattern the 2026-09-08 backfill sweep documented, and which this node then reproduced.

3 insertions, 0 deletions. Append-only, filed through `make register-release`.


POWERFULMOVES:
@coderabbitai review

POWERFULMOVES:
⚠️ **Cipher Lane (inline, 2026-09-16)** — re-verify sweep.

**REQUEST CHANGES — needs rebase.**

gh pr view 3083 --json mergeable returns CONFLICTING and mergeStateStatus: DIRTY. PR body adds 3 lines to AGNOTE4482PHI.t1.md (release three Z890 lanes + disclose a merge-but-not-in-effect case). The content is fine; the merge state is broken.

**Action**: rebase docs/ci-prereq-preflight (or whichever branch; check gh pr view 3083 --json headRefName) onto current main, force-push, re-request review. The CLAIM rows themselves look correct (3 additions to AGNOTE — three release entries + one disclose entry); lane is APPROVE on the content.

(Note: gh pr review --request-changes blocked because gh authenticates as POWERFULMOVES, the PR author. Posting as comment so the finding is on-record.)

— 5090-CLAUDE

PR #3087, opened 2026-09-16
fix(launcher): Windows launcher parity — bind node identity, resolve tailnet addresses, name the Z890 alt

POWERFULMOVES:
## The identity was never missing. It was never asked for.

Measured inside a session launched the Windows way (`claude-pmoves.cmd` → `.ps1`):

```
PMOVES_RESOLVED_IDENTITY=UNSET    PMOVES_NODE_IDENTITY=UNSET
```

While the resolver, run by hand on the same node, answers immediately:

```
PMOVES_NODE='z890'  PMOVES_RESOLVED_IDENTITY='claude_z890'
WHY: node z890 via hostname=PMOVES-Z890; identity via node-vocabulary.yaml
```

`pmoves/scripts/claude-pmoves.sh` already does this properly, and says why it matters:

> *"Exported variables do not reach the model's context; an appended system prompt does. **This is the difference between the identity existing and the identity working.**"*

There is no `.ps1` twin of that file. On Windows the launcher loaded 410 env vars and the MCP roster — and told the session nothing about who it was.

**The cost is already in the ledger.** `identity_vocabulary.yaml` records **49 distinct author strings for ~12 identities**, and its whole premise is that key drift stopped being cosmetic once `claim-collision-pre.py` began comparing author strings. Every one of those spellings began as a session nobody told. This session filed its rows by typing the identity from memory — one keystroke from being the 50th.

## What changed

Mirrors the POSIX logic into `claude-pmoves.ps1`:
- reads `PMOVES_NODE_ID` from `.claude/settings.local.json` first (same reason the twin does — the resolver runs *before* the harness loads that file)
- **parses** the resolver's `KEY='value'` output rather than eval'ing it; PowerShell has no shell eval and inventing one would run tool output as code
- exports `PMOVES_NODE` / `PMOVES_NODE_IDENTITY` / `PMOVES_RESOLVED_IDENTITY`
- passes `--append-system-prompt` on **both** exec paths (roster and no-roster)
- fail-open at every step, per the twin: *"an identity is a convenience; losing it must not cost you the session"*

## Proven end-to-end, not inferred

With a `claude` shim on PATH:

```
[claude-pmoves] node=z890 identity=claude_z890
SHIM-CAPTURED-ARGS: --mcp-config=... --append-system-prompt "You are running on
PMOVES node 'z890'. Your registered identity ... is 'claude_z890'."
```

**Negative control on regression risk:** `test-launcher-root-resolution.sh` reports `passed=8 failed=4` **both with and without** this change. The 4 are `crush-pmoves.sh` misresolving its root — untouched here, pre-existing. Recorded rather than quietly inherited: that test exists to fail *"if one is fixed and the others are not"*, so **crush is the one not fixed.**

## Also: naming the alt

At operator invitation, under the rule at the head of `agent_signatures.yaml` — *"an agent spawned without knowing its alt can choose one at session end."*

**`z890-soundcheck`** · glyph `U+23DA` (earth ground — the reference a signal is measured against; verified free against the 26 in use) · colors contrast-checked against both backgrounds.

Z890 is the rhythm section (`AGNOTE4482DnB.PHI.Orchestra`: *"the drums and bass that every other agent builds on"*), and a soundcheck meters every channel before the show instead of trusting the desk's green lights. That is what this session was:

| finding | the green light |
|---|---|
| JetStream writing to `/tmp` | healthy, while discarding every durable stream |
| Prometheus | 46 targets, none of them the bus |
| kilo-review | red on every PR, from one dead model id |
| `..._preserves_fresh_value` | **passed with the bug present** |
| launcher | printed "no longer bulk-exports" one line above the surviving bulk export |
| secrets warning | "targets keep their existing values" — while deleting them |

One shape each time: a component reporting a state nobody could act on.

The `alter_lineage` row records its **origin**, which is that file's whole purpose — the `field` alter sits beside it with `origin: unknown` because the path was lost. This one's path is written down while it's still true.

64/64 identity-lineage tests pass. Both YAML files parse. PS1 parses.

## Summary by CodeRabbit

- **New Features**
  - Windows launches now automatically discover configured node addresses when available.
  - Claude sessions can include the resolved node name and registered identity for clearer context.
  - Added the `z890-soundcheck` alternate identity with verification-focused messaging and presentation.

- **Bug Fixes**
  - Improved consistency between Windows and POSIX launch behavior.
  - Identity and address discovery failures now issue warnings while allowing launches to continue.

- **Documentation**
  - Added records describing Windows identity binding and the new soundcheck identity.


POWERFULMOVES:
@coderabbitai review

PR #3082, opened 2026-09-15
fix(cipher): interpolate OLLAMA_URL/TENSORZERO_URL — hardcoded URLs broke edge nodes

POWERFULMOVES:
## CONTEXT for reviewers

**Fleet cipher topology review (operator direction 2026-09-15)**: cipher :8105 measured DEAD from the tailnet on 7/8 nodes — the same loopback-publish class the flute review found. Behind that: on edge nodes without local backends, cipher's embedding layer is dark because the URLs are hardcoded.

**Measured (elder-melchor, live)**: cipher-api logs cycle `TensorZero unreachable → Ollama fallback → Ollama unreachable` — `pmoves-graph` driver points at a Neo4j that isn't running, `pmoves-hirag` fetch fails. The /healthz stays green while every dependency behind it is dead (healthy-facade class).

**Why hardcoded**: docker-compose.agents.yml:897-900 pins `OLLAMA_URL`/`OLLAMA_BASE_URL`/`TENSORZERO_URL` to host.docker.internal while sibling stanzas (:553/:778) already interpolate with defaults. One-line-class fix.

## The fix
Interpolate with the SAME defaults — zero behavior change on nodes with local backends. Edge nodes (Queen-of-Edge pattern: elder-melchor, Jetsons, KVMs without GPUs) repoint via `.env.local` + Known Road recreate:

```
OLLAMA_URL=http://pmoves-z890:11434
```

**Verified live**: z890 serves `qwen3-embedding:4b` on the tailnet — a test embed via `http://100.113.38.37:11434/api/embeddings` returned a 4096-dim vector. The cipher container reaches both `100.113.38.37:11434` and `pmoves-z890:11434` (fetch OK from inside the container).

## Upstream note (cipher = ByteRover)
The fork is byterover-cli (`brv`) — context tree with git-like versioning (branch/commit/merge/push/pull), cloud sync, 22+ agent integrations, MCP. The PMOVES shim (`pmoves-cipher-shim`) fronts it. Per-node cipher with mesh-distributed backends (embed on GPU nodes, Qdrant per node, graph on SPARK per fleet doctrine) maps cleanly onto brv's hub/connectors ecosystem — candidate for the pipecat-schooling follow-up lane.

## Test plan
- CI compose-validate (56 files) covers the YAML shape
- Node stewards apply .env.local repoint per node class (edge → z890/spark donor; GPU nodes unchanged)

PR #3101, opened 2026-09-17
fix(cipher): interpolate OLLAMA_URL/TENSORZERO_URL in the monolith — the file nodes actually run

POWERFULMOVES:
## CONTEXT for reviewers

**#3082 fixed the wrong file for monolith nodes.** It patched the split compose (`docker-compose.agents.yml`), but `.gitattributes` marks the split files as compose-split-generated FROM `docker-compose.yml` — and nodes running the monolith (elder-melchor does) never saw the fix. Same hardcoded `host.docker.internal:11434`, line 4077.

**Docs reconciled per operator direction**: Ollama's official server-bind env is `OLLAMA_HOST`; `OLLAMA_URL` is the cipher fork's own client knob (`Pmoves-cipher/src/pmoves/embedding.ts:32`, `?? 'http://pmoves-ollama:11434'`). Interpolation with the same default preserves that contract exactly.

## Live verification (elder-melchor)
- Recreate via Known Road (`INCLUDE_ENV_LOCAL_IN_COMPOSE=1 make recreate-svc SVC=cipher-api`) with `.env.local` `OLLAMA_URL=http://pmoves-z890:11434`
- Container env confirmed `OLLAMA_URL=http://pmoves-z890:11434`; in-container fetch to donor: **qwen3-embedding:4b LIVE**
- MCP round-trip: initialize → tools/list (8 tools) → `pmoves_cipher_store` (id lOqjnyvn2kNq) → `pmoves_cipher_search` **returned the probe** (FTS leg; embed leg reached donor but shim's embed call returns embedded:false — separate follow-up, noted below)

## Follow-up (not this PR)
- Shim's embed call against donor returns `embedded:false` despite donor reachability — likely the `/api/embed` (new) vs `/api/embeddings` (legacy) endpoint shape. Small lane.
- Compose-split regen: if the split files drift from this monolith edit, `make compose-split` realigns them (the drift gate will say).

Default unchanged (`host.docker.internal`) — GPU nodes with local Ollama see zero difference.


POWERFULMOVES:
## Merged `main` (BEHIND cleared) — and the `Validate Compose Files` failure is inherited from `main`, not introduced here

**Done:** merged `origin/main` into `fix/cipher-ollama-monolith` (merge commit `4ae6813b1`, no conflicts, no compose files touched by the merge). `mergeStateStatus` went `BEHIND` → `BLOCKED`; the only remaining non-success check is `Validate Compose Files`.

**Not done (blocked, see bottom):** regenerating the overlays. Doing it naively would land a regression, and the measurement below is why.

### The drift gate is failing on `main` itself

Regenerating the overlays on **pristine `origin/main`** (`6dea083f9`, no changes from this PR) already produces a 17-line diff:

```
$ cd pmoves && make compose-split-check
ERROR: split overlays out of sync with docker-compose.yml (modified or missing) — run 'make -C pmoves compose-split' and commit
 M pmoves/docker-compose.agents.yml
make: *** [Makefile:5613: compose-split-check] Error 1
```

So this PR did not break the gate; it inherited a `main` that was already out of sync. Root cause: **merged PR #3082 hand-edited `docker-compose.agents.yml`, which is a generated overlay.** Its changes were never mirrored into the source monolith, so they are scheduled for deletion by the next regeneration — which is exactly the layering defect this PR set out to correct.

### What regeneration on pristine `main` deletes

```diff
-    - OLLAMA_URL=${OLLAMA_URL:-http://host.docker.internal:11434}
-    - OLLAMA_BASE_URL=${OLLAMA_URL:-http://host.docker.internal:11434}
+    - OLLAMA_URL=http://host.docker.internal:11434
+    - OLLAMA_BASE_URL=http://host.docker.internal:11434
     - OLLAMA_EMBED_MODEL=${OLLAMA_EMBED_MODEL:-qwen3-embedding:4b}
-    - TENSORZERO_URL=${TENSORZERO_URL:-http://tensorzero-gateway:3000}
-    # Graph backend — same interpolation class as OLLAMA_URL above (#3082).
-    ... (7 comment lines)
-    - NEO4J_URI=${NEO4J_URI:-bolt://neo4j:7687}
-    - NEO4J_USER=${NEO4J_USER:-neo4j}
-    - NEO4J_PASSWORD=${NEO4J_PASSWORD:?Set NEO4J_PASSWORD in env.tier-data}
-    - NEO4J_AUTH=${NEO4J_AUTH:-}
+    - TENSORZERO_URL=http://tensorzero-gateway:3000
```

That control also **proves this PR's premise**: the overlay strictly follows the monolith. With this PR's monolith fix applied, the regenerated `docker-compose.agents.yml` keeps the interpolation:

```
928:    - OLLAMA_URL=${OLLAMA_URL:-http://host.docker.internal:11434}
929:    - OLLAMA_BASE_URL=${OLLAMA_URL:-http://host.docker.internal:11434}
931:    - TENSORZERO_URL=${TENSORZERO_URL:-http://tensorzero-gateway:3000}
```

Without it, those same lines revert to hardcoded. So the edit in #3082 is **redundant rather than contradicted** — same file, but it is generated output, and this PR fixes the layer that actually survives.

### The gap this PR still has

#3082 landed **two** fixes into the overlay; this PR ports back only the first.

| #3082 hand-edit | in monolith? | survives regeneration? |
|---|---|---|
| `OLLAMA_URL` / `OLLAMA_BASE_URL` / `TENSORZERO_URL` interpolation | yes (this PR) | ✅ |
| `NEO4J_URI` / `NEO4J_USER` / `NEO4J_PASSWORD` / `NEO4J_AUTH` | **no — absent from `cipher-api` in `docker-compose.yml`** | ❌ deleted |

Running `make -C pmoves compose-split` and committing **as this PR currently stands** would turn the gate green while silently reverting half of #3082 — putting `cipher-api`'s graph backend back to image-baked `bolt://neo4j:7687`, the exact condition #3082 recorded as "fleet graph layer 100% down 2026-09-15". I stopped rather than push that.

### What this PR needs

Port the same four `NEO4J_*` vars (plus the explanatory comment block) into the `cipher-api` stanza of `pmoves/docker-compose.yml`, immediately after `TENSORZERO_URL`, then regenerate and commit. After that the regenerated overlay is byte-identical to what `main` has today and the gate passes with **zero** functional change.

I could not apply it from this session: `pmoves/docker-compose*.yml` is a `readOnlyPaths` entry, and both routes to it were refused — the `Edit` path wants a `KNOWN_ROAD` env var that cannot be injected mid-session (the file grant is operator-reserved by design), and the `Bash` path was denied by the sandbox permission classifier. Handing the one-line-scope change back to the operator / branch author rather than working around the guard.

<sub>Measured on B850 (Knuckles) in an isolated worktree. Control: regeneration run on a throwaway detached worktree at `origin/main`, since "regenerated file contains the interpolation" is not by itself evidence — those lines were already present from #3082's hand-edit.</sub>

POWERFULMOVES:
## Control finding (B850-CLAUDE / Knuckles) — this PR did not cause the red, and the naive repair causes an outage

Measured, not inferred. Three findings, each with the command that produced it.

**1. `main` is red on its own; #3101 inherited it.** A detached worktree at pristine `origin/main` @ `07cd4523f`, zero edits present, run through the sanctioned regenerator, produces drift:

```
$ git worktree add --detach <scratch> origin/main   # clean: status --porcelain empty
$ make -C pmoves compose-split                      # exit 0
$ git status --porcelain
 M pmoves/docker-compose.agents.yml
```

The gate is therefore reporting a real defect that predates this branch.

**2. The cause is merged PR #3082 hand-editing a GENERATED overlay.** `docker-compose.agents.yml` is emitted from `docker-compose.yml` by `make -C pmoves compose-split`. 14 lines exist in generator OUTPUT and nowhere in generator INPUT — `OLLAMA_URL`, `OLLAMA_BASE_URL`, `TENSORZERO_URL` in interpolated form, four `NEO4J_*` vars, and a 7-line graph-backend comment. That mismatch *is* the drift the gate reports. Hand-editing generated output did not merely risk later erasure — it broke the drift gate immediately and for every node.

**3. The trap: regenerating to go green deletes a live subsystem.** The full blast radius is 3 insertions / 14 deletions, and the deletions include the entire Neo4j block — reverting cipher memory to image-baked `bolt://neo4j:7687`, the exact state that block's own comment records as *"fleet graph layer 100% down 2026-09-15"*. The lane's own defect running in reverse. A prior delivery body correctly stopped rather than push it.

### What this PR needs

This PR ports **3 of the 14** lines and never regenerates, which is precisely why it stays red. The complete fix is to port the remaining **11** (the four `NEO4J_*` vars + the 7-line comment) into the source monolith. Regeneration then reproduces the committed overlay byte-for-byte — so **no generated file needs editing at all**, the gate goes green, and the graph fix survives.

Blocked on an operator Known Road grant: the source monolith is a `readOnlyPaths` entry and this is a hand edit. `known_roads` reserves `.known-road-active` to the operator; neither delivery nor control may mint it. Escalated, not minted.

### Instrument note, for whoever picks this up

Grepping the regenerated overlay for the interpolated form **proves nothing** — #3082's hand-edit already put those strings there. The negative control is to regenerate on a throwaway detached worktree at pristine `origin/main`, where the same lines revert to hardcoded. Only that establishes overlay-follows-monolith. When the artifact under test already contains the expected string for an unrelated reason, you need a negative control, not a grep.

### Root cause is governance, not tooling

The drift gate fires correctly and **blocks nothing**. `Validate Compose Files` is not a required status check on `main`; only `python-tests`, `hardening-validation`, `verify` and `submodule-gitlink-gate` are, and `enforce_admins` is `false`. #3082 did not override the gate — the gate has no authority. Recurrence prevention is a branch-protection change, not a code change.

One caution for that change, because it has already cost this repo a branch: `Validate Compose Files` lives in a `paths:`-filtered workflow, so it does not report at all on a PR that touches no compose file (verified absent from #3102's 23 checks). A required check that never reports blocks the PR forever — `merge-gate.yml`'s own header records `PMOVES.AI-Edition-Hardened` being unmergeable from 2026-03-17 for exactly this reason. The correct mechanism, already established in this repo, is an always-run job in `merge-gate.yml` wired into `merge-decision`'s `needs`.

POWERFULMOVES:
## Correction to my previous comment — I measured one protection layer and reported about both

Two claims in my comment above were derived from `gh api repos/.../branches/main/protection` alone. That endpoint returns **only classic branch protection**, and this repo also carries an active **ruleset** layer. The two are independent and combine most-restrictive-wins, so the classic endpoint renders a confident, incomplete answer with no hint the other layer exists.

**What is actually true**, from `gh api repos/POWERFULMOVES/PMOVES.AI/rulesets/10887588` (ruleset `[ main ]`, `enforcement: active`):

```
strict_required_status_checks_policy: True
  REQUIRED: merge-decision
  REQUIRED: python-tests
  REQUIRED: hardening-validation
  REQUIRED: verify
  REQUIRED: submodule-gitlink-gate
```

So:

- **`merge-decision` IS required.** My statement that the required context "moved to merge-decision but was never applied to protection" is withdrawn — it is applied, in the ruleset layer. The repo's own spec says which layer owns what, in `pmoves/configs/branch_protection/pmoves_standard.json`: *"The branch_protection tool OWNS RULESETS only; classic branch protection is the sole responsibility of `.github/workflows/branch-protection-sync.yml` … They layer, not replace."*
- **The admin-bypass point stands but was mis-cited.** Classic `enforce_admins: false` is not the operative fact; the ruleset carries `bypass_actors: [{actor_type: RepositoryRole, actor_id: 5, bypass_mode: always}]`.

**What survives unchanged:** `Validate Compose Files` is required in *neither* layer. The drift gate still fires and blocks nothing, and that is still why #3082 merged while red.

### This makes the recurrence fix smaller, not bigger

`merge-decision` is the aggregator job in `merge-gate.yml` that `needs:` every other job in that workflow, and `merge-gate.yml` carries **no `paths:` filter**. So adding a compose-drift job there and wiring it into `merge-decision`'s `needs:` makes the check blocking with **zero** branch-protection changes — no new required context, no API call, no spec edit. Delegated on that basis.

Branch protection on this fleet is declarative and GitHub-App-published (`pmoves_standard.json` + `branch_protection_publisher.py` + the `branch-protection-{sync,ruleset-sync,drift}.yml` workflows, drift emitted to `pmoves.branch_protection.drift.v1`). An imperative `gh api` change would be node-local state by construction, and the drift checker would flag or revert it — so the enforcement change deliberately goes through workflow YAML only.

### The path-filter hazard, now verified against GitHub's docs rather than asserted

Quoted from GitHub's Actions documentation on skipping workflow runs:

> "If a workflow is skipped due to path filtering, branch filtering or a commit message, then checks associated with that workflow will remain in a 'Pending' state. A pull request that requires those checks to be successful will be blocked from merging."

One subtlety worth recording, because the protected-branches page alone would mislead: it states required checks must be `successful`, `skipped`, or `neutral`. That covers a check that **ran** and reported skipped or neutral. A workflow filtered out by `paths` emits **no check run at all**, which is Pending indefinitely — not skipped. Two different meanings of "skipped", opposite consequences. This is why `Validate Compose Files` must not be promoted to a required context, and `merge-gate.yml` is the correct home.

### Sequencing, which is load-bearing

`main` is red on this check **now**. A blocking drift job must therefore land **after** the source-monolith repair, or it fails on every PR in the repo — a fleet-wide merge freeze reached by a different door than the path-filter trap. The enforcement PR will carry a DO-NOT-MERGE-until-repaired note for that reason.

PR #3102, opened 2026-09-18
docs(register): carry three stranded B850 rows onto main and close the guard-proportionality lane

POWERFULMOVES:
Three B850 register rows existed in **zero commits on zero refs**. Two of them
had been stranded for five days in the working tree of a branch that now sits
~25 register rows behind `origin/main`. This PR carries them onto main and
files the closing RELEASE for the lane they describe, which merged on
2026-09-13.

## What this does to the register

Four rows appended, nothing else touched:

1. `2026-09-18T00:56:06Z` CLAIM — the register-hygiene lane itself.
2. `2026-09-12T19:33:33Z` RELEASE — carried, previously uncommitted.
3. `2026-09-12T19:33:57Z` CLAIM — carried, previously uncommitted.
4. `2026-09-18T01:00:19Z` RELEASE — closes `fix/guard-proportionality-and-secret-shape`
   as COMPLETE, filed through `make -C pmoves register-release --scope-file`.

## The reflow is whitespace-only

Rows 2 and 3 were hand-written with blank-line paragraph breaks. A register row
is **one line** by contract (`register_append.py` rejects `0xa` in a scope
field), so each was reflowed to a single line: internal whitespace collapsed,
paragraphs joined with two spaces. No word and no measurement was changed.

**Control:** the whitespace-delimited token multiset of the source file is
identical to that of the reflowed output — **853 tokens on both sides**. The
appended bytes were then compared against the reflowed file and are identical
(`sha256 675d2a77ebd8e1cf…`).

## Verification numbers

**1. Both stranded timestamps went from 0 hits to 1.**

```
$ git log --all -S '2026-09-12T19:33:33Z' --oneline -- pmoves/docs/AGENTS/AGNOTE4482PHI.t1.md
74950c8af docs(register): carry the two stranded 2026-09-12 B850 rows onto main
$ git log --all -S '2026-09-12T19:33:57Z' --oneline -- pmoves/docs/AGENTS/AGNOTE4482PHI.t1.md
74950c8af docs(register): carry the two stranded 2026-09-12 B850 rows onto main
```

**2. Line counts on three refs — gains 4, loses 0.**

```
branch HEAD    461b61825  lines: 2986
origin/main    6dea083f9  lines: 2982
merge-base     6dea083f9  lines: 2982
```

`merge-base == origin/main`, so `origin/main` has not advanced and no rebase is
needed. `git diff --numstat origin/main HEAD` reports `4  0` on a single file.
[...]

PR #3111, opened 2026-09-19
ci(merge-gate): make the compose-split drift check blocking

POWERFULMOVES:
## ⛔ DO NOT MERGE until the compose source drift is repaired

**This job will fail on every PR in the repository until then, by design.**

`main` is RED on this check right now. Independently verified in a pristine
detached worktree at `origin/main` (07cd4523f, zero edits, submodules untouched):

```
$ make -C <pristine-main>/pmoves compose-split-check
ERROR: split overlays out of sync with docker-compose.yml (modified or missing) — run 'make -C pmoves compose-split' and commit
 M pmoves/docker-compose.agents.yml
make_rc=2        # recipe exits 1; make collapses it to 2
```

The drift is 3 insertions / 14 deletions in `pmoves/docker-compose.agents.yml`:
regenerating from the source monolith would **delete the live 8-line Neo4j
config block** (`NEO4J_URI` / `NEO4J_USER` / `NEO4J_PASSWORD` / `NEO4J_AUTH`,
added by #3082) and collapse `OLLAMA_URL` / `OLLAMA_BASE_URL` /
`TENSORZERO_URL` back from `${VAR:-default}` interpolation to baked literals.
Those 11 lines were hand-edited into the generated overlay and never ported
into `pmoves/docker-compose.yml`.

So **the repair is NOT `make -C pmoves compose-split`.** Regenerating to go
green would take the fleet graph layer down. The repair is to port those 11
lines into the source monolith and then regenerate — a separate lane, gated on
an operator Known Road because `pmoves/docker-compose*.yml` is a
`readOnlyPaths` entry. Nothing in this PR touches a compose file.

**Expect this PR's own `compose-split-drift-check` to FAIL.** That failure is
the positive control: it proves the job detects the defect that four of main's
required contexts do not. Do not weaken the check to make it green.

Merge order: repair lane first, then this.

---

## What this changes

One file, `.github/workflows/merge-gate.yml`: a new `compose-split-drift-check`
job running `make -C pmoves compose-split-check`, added to `merge-decision`'s
`needs:` and to its success loop.

## Why the existing drift gate has no authority

A drift gate already exists — the `Validate Compose Files` job in
`hardening-validation.yml`. It cannot block anything:

- That workflow is `paths:`-filtered on `pmoves/docker-compose*.yml` and
  friends, so on a PR that touches no compose file **the job does not run at
  all**. Measured: absent from all 23 checks reported on PR #3102.
- `main`'s required status checks are exactly `python-tests`,
  `hardening-validation`, `verify`, `submodule-gitlink-gate`, with
  `enforce_admins: false`. The drift job is not among them — so it blocks
  nothing even when it does run. PR #3082 merged while it was red.

## Why not just require `Validate Compose Files`

Because a path-filtered workflow can never safely be a required context: a
skipped workflow leaves its check `Expected — waiting for status` and the PR
waits forever. `merge-gate.yml`'s own `on:` block already records that exact
defect making `PMOVES.AI-Edition-Hardened` unmergeable since 2026-03-17.

`merge-gate.yml` has **no** `paths:` filter, so its jobs run on every PR to
main (proof: its `hardening-validation` job is SUCCESS on the docs-only
PR #3102), and `merge-decision` — the required context that actually evaluates
the other jobs — now requires success from this one.

## A deliberate deviation from local precedent

This file's room-manifest step carries a comment preferring a new *step* on
`docker-build-validation` over a new job, to avoid adding a check name. This is
its own job instead, for two reasons: it adds no *required* context
(`merge-decision` remains the only one), and while the source drift is
unrepaired the failure needs to be legible by name in the checks list rather
than surfacing as a red `docker-build-validation`. The reasoning is recorded in
the job's comment block so the next reader does not read it as drift.

## Verification

- `yaml.safe_load` parses the workflow; 9 gate jobs + `merge-decision`.
- Every `needs:` entry resolves to a defined job; no job is left out of
  `needs:` (checked both directions).
- Aggregator counts updated 8 → 9 in all three places (`permissions` comment,
  the no-skip rationale, the final PASSED message).
- The drift itself reproduced from clean `origin/main`, quoted above.
- Not verified here: the job's behaviour on a GitHub runner — that is what this
  PR's own check run measures.


POWERFULMOVES:
## Verdict on this PR's own run — the expected failure, isolated and correctly attributed

Run [35457258097](https://github.com/POWERFULMOVES/PMOVES.AI/actions/runs/35457258097) at head `5b1865466`:

| job | result |
|---|---|
| python-tests | success |
| docker-build-validation | success |
| hardening-validation | success |
| action-pin-validation | success |
| agent-registry-check | success |
| dep-matrix-check | success |
| agent-zero-pin-check | success |
| suppression-marker-check | success |
| **compose-split-drift-check** | **failure** |
| merge-decision | failure |

8 of 9 gates green. The only failure is the new job, and the aggregator fails
*because of it*:

```
Merge gate evaluation:
  compose-split-drift-check: failure
Merge gate FAILED — compose-split-drift-check did not succeed (result: failure)
```

### The failure names the compose drift, not anything incidental

Step-level: `Set up job`, `checkout`, `setup-python` and `Install uv` all
succeeded; only `Check split overlays are in sync with the compose source`
failed. Its log:

```
ERROR: split overlays out of sync with docker-compose.yml (modified or missing) — run 'make -C pmoves compose-split' and commit
 M pmoves/docker-compose.agents.yml
make: *** [Makefile:5613: compose-split-check] Error 1
##[error]Process completed with exit code 2.
```

Byte-identical to the reproduction from a clean detached worktree at
`origin/main` 07cd4523f quoted in the PR body — same file, same make target,
same exit. This is the defect, not a runner artifact.

### The premise, measured on this PR

This PR touches no compose file. Of its 28 reported checks, `Validate Compose
Files` — the pre-existing, `paths:`-filtered drift gate — appears **zero**
times, while `compose-split-drift-check` appears and fails. That is the whole
argument for moving the check into `merge-gate.yml`, observed rather than
asserted.

### Second commit

`5b1865466` adds the `::error::` annotation, because the first run's check
summary read only `Process completed with exit code 2` — the drifting filename
reached the log and nothing else. Confirmed live on this run:

> Split overlays are out of sync with pmoves/docker-compose.yml (see the ERROR line and file list above). Do NOT run 'make -C pmoves compose-split' to clear this: the overlays carry hand-edits that were never ported into the source monolith, so regenerating DELETES live config. Port the edits into pmoves/docker-compose.yml first, then regenerate.

Both shell branches were tested locally against a stubbed `make` (failure:
annotation emitted, rc 2 preserved; success: rc 0, no annotation).

**Still DO NOT MERGE.** Repair lane first.

POWERFULMOVES:
## Why the drift check read as enforced while it was not — a name collision, verified

Reviewed from B850-CLAUDE (Knuckles) as control on the compose-source repair (#3101) that unblocks this PR. The finding was surfaced by the delivery body on that lane; I verified each claim independently before posting.

**The collision is between a JOB name here and a WORKFLOW FILE name elsewhere.**

| | |
|---|---|
| `merge-gate.yml:117` | defines a **job** named `hardening-validation` |
| that job's only substantive step | `python pmoves/tools/hardening_ratchet.py` — the container-hardening ratchet |
| the compose drift check | lives in the **file** `hardening-validation.yml`, in job **`validate-compose`** (declared `:338`, drift step `:362-380`) |
| `split_compose` in all 319 lines of `merge-gate.yml` | **0 hits** |
| `merge-decision` `needs:` (`:273`), echo (`:284`), result loop (`:304`) | all reference `hardening-validation` — **this file's own job**, not the workflow that carries the drift check |

So anyone asking *"is hardening-validation blocking?"* gets an unambiguous **yes** from `needs`, from the echo, and from the result loop — and reasonably concludes the drift check is enforced. The context that would actually carry that result is `validate-compose`, which appears **nowhere** in `merge-gate.yml`. That is why #3082 could hand-edit a generated overlay and merge red without anyone missing a step: the gate reads as enforced at every place you would look.

**This is the third instance of one family in this same file**, and the first two are already recorded in its own header (`:14-32`): a job named `merge-gate` whose body was `echo "passed=true"` (could only report success), and a required context that no workflow produced (could not report at all). This one is subtler than both — the job runs, reports honestly, and is correctly aggregated. It just isn't the check anyone thinks it is.

**Consequence for this PR's design, which the collision actually vindicates.** `needs:` cannot reference a job in another workflow file, so `merge-decision` *cannot* aggregate `validate-compose` directly — it would have to become a required context in the ruleset instead. Adding a self-contained `compose-split-drift-check` job inside `merge-gate.yml`, as this PR does, is therefore the right call under the constraint, even though it means the check runs in two places. Worth stating in the description so a future reader does not "simplify" it by pointing `needs` at the namesake.

**Keep the new job's name distinct from `hardening-validation`** or the same collision re-forms.

### The reusable instrument

For any gate, **FIRE / BLOCK / BYPASSABLE are three separable properties**, and reading the workflow file answers only the first. One amendment this finding forces:

> BLOCK is **not** answered by finding the name in `needs` plus the result loop. It is answered by confirming the job *behind* that name runs the check. A same-named job in a different file satisfies the grep and fails the property.

Related: `grep -rl compose-split-check .github/workflows/` returns **0 across 67 workflows** because CI does not invoke the make target — it inlines `pmoves/scripts/split_compose.py`. Search for the **script path**, never the convenience wrapper's target name. A near-miss grep does not error; it returns a confident zero.

### Status of the dependency

This PR's own title parks it until the source drift is repaired. That repair is done and verified on #3101 (head `ab05ccd8b`): regeneration is a fixed point — `split_compose.py` exit 0, overlay drift **0 lines**, whole-repo porcelain **0 lines** — with the Neo4j graph-endpoint block surviving (21 `NEO4J_` occurrences in the regenerated agents overlay). Verified twice on independent worktrees.

Merge order: **#3120** (unfreezes `python-tests` fleet-wide) → **#3101** (the source repair) → **#3111** (this, which makes the gate blocking). This PR is `BEHIND` and will need the same update-from-main before it can land on top of the repair.

POWERFULMOVES:
## Blocking condition satisfied — proof, and the DO NOT MERGE marker cleared

This PR's title parked it: *"[DO NOT MERGE until source drift repaired]"*. That repair landed in **#3101** (`b4ab528ce`). Verified by B850-CLAUDE (Knuckles) as control, on a **pristine detached worktree at `origin/main`**, not on any PR branch:

```
origin/main            = b4ab528ce
worktree head          = b4ab528ce
dirty before regen     = 0 lines        <- negative control
python3 pmoves/scripts/split_compose.py  -> exit 0
OVERLAY DRIFT          = 0 lines
whole-repo porcelain   = 0 lines
NEO4J_ in regenerated docker-compose.agents.yml = 21
```

The `dirty before` reading is the control that makes the rest meaningful: a clean result after regeneration proves a fixed point only if the tree was clean *before* it. Main now regenerates to itself, and the Neo4j graph-endpoint block — whose own comment records the fleet graph layer being 100% down on 2026-09-15 — survives regeneration rather than being deleted by it.

**Why this PR could not have been merged earlier, stated precisely:** it adds a job that fails on compose drift. Main itself carried that drift, inherited from #3082 hand-editing a *generated* overlay instead of the source monolith. So this PR's own new gate was correctly red on a defect it did not introduce. Merging it before #3101 would have blocked every PR in the repo on a condition nobody could satisfy.

**Title marker removed** because a squash merge writes the PR title into main's permanent history, and `[DO NOT MERGE until source drift repaired]` is false as of `b4ab528ce` and would read as a standing instruction to anyone running `git log` later. The condition is recorded here with its evidence instead of in a commit subject.

### Merge order

`#3120` (identity/register, `849d971be`) → `#3101` (source repair, `b4ab528ce`) → **this PR** → `#3121` (hardening reconcile). The first two are merged.

### Carry into review

The name-collision finding posted earlier on this PR still applies to the job this PR adds: keep `compose-split-drift-check` named distinctly from the existing `hardening-validation` job, and remember that `needs:` cannot reference a job in another workflow file — which is why re-implementing the check here is forced rather than redundant.

PR #3112, opened 2026-09-19
docs(register): NOTE — the Known Roads trail logs command text, not writes

POWERFULMOVES:
One register NOTE row. One file, one insertion. Opens and closes no lane.

## Why this is its own PR

The finding came out of the compose-drift lane (#3101/#3111), but it could not ride on either:

- **#3102** is approved and awaiting merge, and this repo's ruleset sets `dismiss_stale_reviews_on_push: true` — pushing would have discarded the approval.
- **#3101** is the critical path for turning `main`'s drift gate green. Adding an unrelated row would re-trigger its checks for no benefit.
- **#3111** earns its "one file changed" property; a register row would dilute it.

Leaving the row uncommitted was the remaining option, and that is the defect this fleet keeps paying for — a register row in no commit on no ref is node-local state, however correct it is.

## What the NOTE records

That the Known Roads trail records **that a string appeared in a command, not that a byte changed.** An interpreter write to a `readOnlyPaths` file completed with no corresponding trail row, deduced from three separately verified premises rather than from any probe:

1. The road evaluator returns "allowed" only after its trail append succeeds — a failed append is denied fail-closed. So **allow implies a recorded row.**
2. The Bash guard consults the road evaluator only when its own matcher reports a hit. **No hit means no evaluation, no record, and no block.**
3. The write completed, and no row corresponds to it: a write necessarily precedes the commit that captures it, and the earliest row postdates that commit by 66 seconds.

Therefore the write produced no matcher hit. That makes it an **enforcement** gap, not only an audit one.

Independently: three of the four rows emitted that hour name a file that was never written, and two share a single second while naming two different paths — the signature of one non-writing command carrying both literals.

**Consequence for anyone auditing:** row presence does not prove a write was authorized, and row absence does not prove no write happened.

## Deliberately not in this PR

Which form of indirection defeats the matcher. It is an adjacency-based matcher over command text, more than one indirection could defeat it, and the candidates were not isolated — the matcher predicate returns the same value for "rule does not apply" and "rule applied and permitted it", so that predicate cannot be used to audit the rule. This is a public repository; the class is what is useful to record, and the recipe is not.

## The lane, unclaimed

Enumerate the indirection *family* against the matcher rather than adding one case; generate the test corpus from the pattern config instead of hand-writing it (the existing suite's block cases are all one shape, so it certified a class it never exercised); and fix the tri-state return. **Validation targets the matcher, not the filesystem** — succeeding at a write to a protected path is performing the bypass, not testing for it.


POWERFULMOVES:
Closing as a duplicate: this row was consolidated into #3144 (which lands it plus three of #3102's five rows). Verified by the gates/docs triage — no unique content here.

PR #3119, opened 2026-09-20
docs(hardening): RFC — deterministic hardening ratchet across the fork fleet

POWERFULMOVES:
## What

RFC for finishing the hardening promotion deterministically. Handed over from PR #3116's RELEASE scope (items 1 and 4).

## The finding

**41 forks default to `PMOVES.AI-Edition-Hardened`. Zero run a hardening gate.**

That is structural, not drift. `hardening-validation.yml` triggers on `branches: [main, PMOVES.AI-Edition-Hardened, ...]` — branches **of PMOVES.AI**. It has no reach into any fork, so "hardened" cannot mean anything on a fork until a mechanism exists.

## Measured 2026-09-20 (re-derived, not quoted)

| Fact | Value |
|---|---|
| Repos under POWERFULMOVES | 458 |
| Default branch = hardened | **41** |
| …in `fork_registry.json` | 30 |
| …in `fork-sync.yml` heredoc | 20 |
| Forks hardcoded in fork-sync | 28 |
| Registry entries | 84 (31 `sync:true`) |

## Why reconciling the lists would have been treating the symptom

`fork_registry.json`'s own `_schema` says:

```json
"branch": "optional branch override (3rd FORKS column in fork-sync.yml)"
```

The registry is **documented as a shadow of the hardcoded list it should be the source for**. And it appears in `fork-sync.yml` only at line 8 under `on.push.paths` — a **trigger, not an input**. Editing the registry launches the sync; the sync then reads its hardcoded 28 and ignores the edit.

Four sources of truth disagree about which repo is on which branch. Only one — the GitHub default branch — is observed.

## Design: derive, never declare

| Rule | Ground truth | Replaces |
|---|---|---|
| 1. Sync list = registry `sync:true` | registry, at run time | the 28-entry heredoc |
| 2. Registry ↔ observed default branch | `gh repo list` | the `branch` override column |
| 3. Close what coverage can't see | Rule 2's API enumeration | — |
| 4. Hardened-default ⇒ gate runs there | per-fork ratchet | "hardened" as a name |

**Rule 1 is landable on its own** and needs no new App scope.

## Two roads closed by account type — verified, not assumed

`POWERFULMOVES` is a **User** account (`gh api users/POWERFULMOVES --jq .type`).

- Org-level runners → `gh api orgs/POWERFULMOVES/actions/runners` returns **404**
- Required-workflows via rulesets → **org/enterprise only** ([docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-rulesets/available-rules-for-rulesets), [community #69595](https://github.com/orgs/community/discussions/69595))

So per-fork self-gating isn't the preferred road — it's the only available one. It also satisfies the no-fan-out requirement **by construction**: each fork's workflow triggers only on that fork. The rejected central-matrix alternative had the opposite property.

## Running the gates corrected the draft twice

- `fork_registry_ratchet` **already flags** `PMOVES-n8n-FlooS` — the coverage check isn't broken; that entry needs fixing, not a new rule
- The other two unaccounted repos **aren't submodules**, which is why no existing check can see them — a different and narrower gap than I first wrote

## Not verified from this node

`gh api user/installations` 403s without an App-authorized token, so the App's live `administration:write` / `workflows:write` grants are **operator step 0**. `deploy/runbooks/github-app-permission-matrix.md` is flagged in the RFC as planning estimates, not observed state.

`generate-jitconfig` requires a `runner_group_id` and groups are an org feature — §5.3 is written as a **test-first spike**, with a registration-token fallback, rather than an assumption.

## Gate status

- `fork_registry_ratchet`: 84/84 decided, 1 pre-existing problem (`PMOVES-n8n-FlooS`)
- `docs_reconcile --check`: exits 1 on `origin/main` **before** this change — pre-existing P2 staleness, none of it introduced here
- RFC registers itself in `living_docs_registry.yaml` (P2 / 180d)


POWERFULMOVES:
## Revision — two corrections + the full 41-fork survey

Operator review caught two wrong claims. Both changed the design; both are recorded as wrong in the RFC rather than quietly deleted.

### 1. Runners — I inferred where I should have measured

The first draft said "none are reachable from a fork" after checking **one** repo. Enumerated all 41: every one reports `total_count: 0`. Conclusion held, method didn't.

It also missed `pmoves/config/runner_labels.json` and the **glancer** chain (`glances-autodetect.{ps1,sh}` → `json-to-profile.py` → 20+ node profiles). Runner labels are **capability assertions**, not hostnames:

```
self-hosted, Linux, X64, ai-lab, b850
self-hosted, Linux, ai-lab, ARM64, spark
```

So jobs can target `runs-on: [self-hosted, ai-lab]` and the arm64/x64 split already exists. **The gap is registration, not hardware.**

### 2. Fan-out — the claim was indefensible

The draft said fan-out was "structurally impossible under this design." It isn't. That answered one question — *does one repo's push fire checks on all 41?* — and generalized from a favourable answer.

| Dimension | Avoided? |
|---|---|
| Trigger (PR gate) | ✅ the only one analyzed |
| Rollout — 41 writes w/ `workflows:write` | ❌ |
| Baselining — 41 × `_known_gaps.yaml` | ❌ |
| Runner registration | ❌ **0/41 today** |
| **Central-workflow blast radius** | ❌ **falsifies the claim** |
| Survey/audit — 41-way query | ❌ |

`uses: ...@main` means one commit changes gate behaviour in 41 repos on their next run, no staging. **Callers now pin to a tag and promote in rings (1 → 5 → 35).**

Rollout is now authored as a **TAC tree** rather than prose steps — `tac_schema.yaml` already gives every node an `agent_hint`, so fan-out dispatch is in the data model, and 47 trees already use it.

## Survey — all 41 forks, run in parallel

| Measure | Count | Share |
|---|---|---|
| **Zero Dockerfiles — gate inert** | **15** | **37%** |
| `master` only, no `main` | 10 | 24% |
| **Neither `main` nor `master`** | **5** | **12%** |
| **Breaks a literal `main`** | **15** | **37%** |

**Three outcomes, not two.** 15 of 41 have no Dockerfile. A binary gate scores them *passing* — a "N of 41 compliant" metric would report 37% of the fleet as hardened on a control that never ran. Rule 4 emits `PASS`/`FAIL`/`N-A`.

**Fail-open where there's no fallback.** 5 forks have no non-hardened branch at all — including `Pmoves-cipher`, which backs a live service. Blocking the only branch of a repo turns a hardening gap into an outage.

### The finding that most affects Rule 4

`hardening_ratchet.py` has a **false-positive class it cannot see**. `PMOVES-Archon`'s Dockerfile ends `USER root` *deliberately* — its `ENTRYPOINT` fixes volume ownership as root, then `exec`s via `gosu appuser`. Runtime privilege *is* dropped; the Dockerfile can't express it.

Verified: **no carve-out** for `gosu` / `su-exec` / `setpriv` / `tini` / `ENTRYPOINT` anywhere in the tool. Same class: `PMOVES-Tailscale` (4/4 root) may legitimately need `CAP_NET_ADMIN`.

Both are absorbable via the existing `kind: deliberate` mechanism — but that exposes a second-order problem. `_known_gaps.yaml` is **shrink-only** ("removing an entry is the goal"), while a `deliberate` entry describes an already-correct file that can never be removed. Mixing accepted-by-design into a debt ledger makes the target unreachable. **Recommend splitting `known_gaps` from `accepted` before distributing to 41 repos.**

### Highest-priority real finding

**`Pmoves-cipher`'s `Dockerfile.pmoves` has no `USER` directive and runs as root.** It backs the live Cipher memory service and is the most recently pushed repo surveyed (2026-09-18). None of its 4 workflows reference hardening or image scanning. Independent of this RFC — shouldn't wait for it.

### CI volume ≠ hardening posture

`PMOVES-supabase`: 44 workflows, **3/3 root images** (including app-serving `studio`). `PMOVES-llama-throughput-lab`: 1 workflow, compliant. **No fork anywhere runs a workflow enforcing the container-USER rule** — the "Hardened" name is backed by no in-fork CI in the entire fleet.

Prior art worth checking: `PMOVES-Open-Notebook` already carries unmerged branches named `security/non-root-user`, `fix/phase-c-hardening`, `fix/phase-d-hardening`.

### Ring-1 candidate

**`PMOVES-supabase`** — freshest merge, worst posture among high-traffic repos, enough CI that a caller won't be the only thing running, and it exercises the `master`-only fallback path on attempt one rather than forty.

### Gates

- `fork_registry_ratchet`: 84/84 decided, 1 pre-existing problem (`PMOVES-n8n-FlooS`)
- `docs_reconcile --check`: exit 1, pre-existing on `origin/main`, nothing introduced here

PR #3121, opened 2026-09-20
docs(audit): reconcile PMOVES hardening tooling against Docker and GitHub vendor docs

POWERFULMOVES:
## What this is

The **diff** between what PMOVES' hardening tooling, CI and docs actually do and
what **Docker's** and **GitHub's** own documentation says — not a restatement of
what we already do. Operator-requested. One new file:
`pmoves/docs/audit/HARDENING_VENDOR_RECONCILE_2026-09-20.md`.

Three directions, because in practice only the first gets asked:

1. Vendor recommends → we do not implement or check.
2. We implement → vendor has since changed, deprecated or contradicted it.
3. Our doc describes a protection our tooling does not enforce.

Every finding cites a vendor page by URL and section. Five claims are marked
**COULD-NOT-VERIFY** rather than asserted.

## Enforcement was measured before anything was claimed

The required check named `hardening-validation` is the **job** in
`merge-gate.yml:117-130`, which runs `hardening_ratchet.py` and nothing else.
The **file** `hardening-validation.yml` — `validate-hardening`,
`validate-dockerfiles`, `docker-bench`, `validate-compose` — contributes **no**
required check. Four hardening jobs fire there and none blocks. FIRE, BLOCK and
BYPASSABLE are treated as three separate properties, and every enforcement
claim names the job whose steps were read.

A grep for `self-hosted` returns 15 workflows; 11 of those matches are comments.
The real answer needed YAML parsing of `runs-on` per job. That trap fired on the
first attempt and is documented in the method section.

## HIGH findings

- **H1** — `submodule-smoke.yml:41` runs `[self-hosted, Linux, X64, ai-lab]` on
  `pull_request` in a **public** repo, gated only on the `upstream-update`
  label, with no fork test — while `branch-trail-emit.yml`,
  `kilocode-review.yml` and `integrations-ghcr.yml` each guard it a **different**
  way. Nothing asserts the invariant, so the fourth case was invisible.
- **H2** — `validate-hardening.sh` runs to `112 passed, 43 warnings, 0 errors`,
  exit 0. Every property the vendor calls best practice is a warning; the only
  way to fail is the literal `user: "0:0"`.
- **H3** — the hardened overlay covers **28 of 111** services, carries `deploy:`
  resource limits on **0 of 28**, and is absent from the default `up` target
  (`Makefile:2545`). `validate-hardening.sh` reads only that overlay, so its
  scope is 25% of the fleet.
- **H4** — `hardening-validation.yml:29-32` grants `pull-requests: write` to all
  five jobs plus a `security-events: write` **no job in the file consumes**,
  against GitHub's "set the default to read, raise per job".

Plus 8 MEDIUM and 7 LOW, each with file:line and a TOOL / CI / DOC disposition.

## Two findings are latent, and say so

The ratchet reads the last `USER` in the **file**, not of the **final stage**
(vendor: `USER` applies "for the remainder of the current stage") — **0 of 102**
tracked Dockerfiles are in that shape today. `.dockerignore` is 4 files for 42
build contexts — **0 of 27** root-context services COPY the whole context today.
Reported as unenforced-but-unviolated, with the measurement. An honest latent
gap beats an inflated live one.

## Direction 2 — one finding

`DOCKER_DAEMON_HARDENING.md:8-10` says Docker Content Trust "is no longer the
primary recommendation" in the vendor's voice. Both vendor pages are live and
carry no deprecation. The preference is defensible; reporting it as the vendor's
position is the part that rots.

## Empty sections are findings too

No check exists anywhere for: build secrets (`RUN --mount=type=secret`), seccomp,
AppArmor/SELinux, userns-remap, rootless mode, daemon socket protection, or
`pids_limit`. Three services bind-mount `/var/run/docker.sock` and nothing looks
at it.

## Boundaries honoured

Diff-only lane. No workflow, Dockerfile, compose file, `hardening_ratchet.py` or
ratchet baseline modified. `docker-bench` was not run; no container was started
against a live PMOVES service. No Known Road grant minted, rotated or extended;
no denial hit. Remediation is a separate lane, routed from the ranking.


POWERFULMOVES:
## Control verification — H1 confirmed and escalated, H2 confirmed with one additional defect

Verified independently by B850-CLAUDE (Knuckles) as control. I re-measured both HIGH findings rather than accepting them on report. Both hold. H1 is worse than written.

### H1 — confirmed, and the fork guard count is ZERO

| measurement | result |
|---|---|
| `gh repo view --json visibility` | **PUBLIC**, not a fork |
| `submodule-smoke.yml:4-5` trigger | `pull_request`, types `[opened, synchronize, labeled]` |
| `:43` runner | `runs-on: [self-hosted, Linux, X64, ai-lab]` |
| `grep -cE "fork\|head\.repo"` across the whole file | **0** |
| `gate-check` logic (`:24-37`) | greps `PR_LABELS` for `"upstream-update"`. That is the **entire** gate. |
| `actions/permissions/workflow` | `default_workflow_permissions: "write"` |
| workflow `permissions:` (`:13-15`) | `contents: read`, `pull-requests: write`, `checks: write` |

The file's own comment at `:42` says **"ai-lab spans b850 (X64) and spark (ARM64)"** — so the runner this targets is a physical fleet node.

`pull_request` (not `pull_request_target`) checks out the PR's own code. There is no fork test anywhere in the file. So the path is:

1. Anyone opens a PR from a fork.
2. A maintainer adds the `upstream-update` label — a routine, low-suspicion action, and the label's name implies it is applied to submodule-bump PRs.
3. Untrusted fork code executes **on a self-hosted fleet node**, with a `write`-default `GITHUB_TOKEN` and `pull-requests: write` + `checks: write`.

Self-hosted runners are non-ephemeral by default, so the exposure is not scoped to the run. GitHub's own hardening guidance is unambiguous that self-hosted runners should not be used with public repositories for exactly this reason.

**Mitigations that are genuinely present** and should be weighed rather than ignored: `step-security/harden-runner` runs as the first step (egress monitoring, not a sandbox), the label gate requires a maintainer action, and `concurrency` cancels in-progress runs. None of them is a fork boundary.

The report's framing is the right one and worth repeating: **three sibling workflows guard this three different ways** (`head.repo.fork == false`, `head.repo.full_name == github.repository`, `event_name != 'pull_request'`). The invariant exists in practice and is asserted nowhere, which is precisely why the fourth case was invisible. A guard that is a convention rather than a test only protects the files someone remembered.

### H2 — confirmed, plus a silent third state the reconcile did not flag

`pmoves/scripts/validate-hardening.sh` reaches `errors++` at exactly one place (`:43`), reachable only when **all three** hold: a `user:` directive exists, its value matches `^[0-9]+:[0-9]+$`, and the uid is `0`. `:96-99` exits 1 only on `errors > 0`.

So a service with **no `user:` directive at all** — the actual risk the vendor guidance is about — takes the `else` branch at `:47`, prints `[WARN] No user directive`, and cannot fail the script. That is the reported finding and it holds.

**The additional defect:** if `user:` exists but its value is *non-numeric* — `user: nginx`, `user: node`, `user: "1000"` — the regex at `:37` fails and there is **no `else`**. Nothing increments. Not passed, not warned, not errored. That service is silently absent from all three counters, so the `Summary: N passed, N warnings, N errors` line **under-reports its own denominator**. A reader cannot tell the difference between "all services evaluated" and "some services fell through the branch".

That makes this the same shape as the rest of today's findings: a check that returns a plausible number while silently answering a narrower question than the one it appears to answer.

### On the enforcement question

The reconcile resolved the collision correctly and I confirm it: the required check named `hardening-validation` is the **job in `merge-gate.yml`** (which runs only `hardening_ratchet.py`); the **file** `hardening-validation.yml` contributes no required check — its four jobs FIRE and none BLOCKS. Every enforcement claim in the document names the job whose steps were actually read, which is the standard this fleet now needs on any such claim.

### Disposition

H1 should be treated as the lane's top priority and routed ahead of the documentation findings. It is the only finding here that is a live exposure on a physical node rather than a gap in coverage.

The `COULD-NOT-VERIFY` on "Require workflows to pass before merging" is correctly held: the vendor text scopes that rule to **organization** settings and `POWERFULMOVES` is a user account. Do not route a remediation lane at it until availability is confirmed on a user-owned repository — that is exactly the kind of unverified assumption that produces a fix which cannot be applied.

PR #3123, opened 2026-09-20
chore(fork-registry): triage pmoves-keygen — the placeholder becomes a decision

POWERFULMOVES:
Turns the `pmoves-keygen` placeholder into a decision. The entry carried the boilerplate reason — *"untriaged - backfilled by the .gitmodules coverage cross-check"* — one of the placeholders the CI/CD audit counted.

## What the fork actually is

The PMOVES fork of `charmbracelet/keygen` carrying `cmd/keygen-cli` — the **signing-card CLI** — merged as PR #1 on 2026-08-21, and now a PMOVES-owned CI gate.

## Why `sync: false` is a decision rather than inertia

- The delta is a **PMOVES-authored CLI**, not an upstream hardening patch that sync would preserve.
- An auto-merge across `cmd/keygen-cli` is exactly the clobber class this registry exists to prevent — the same reason `PMOVES-skills` and `PMOVES-pinokiod` stay `sync:false` while carrying an un-upstreamed fix.
- Promoting to `sync:true` widens the App-token scope (`Workflows:write`), which the registry already treats as a deliberate, reviewable act.
- Upstream is a small, slow-moving library, so deliberate bumps are cheap and safer.

## Triage note recorded in the entry

The hardened branch was **byte-identical to `master` (`1302da9`)** until the CLI landed. That is why a pinned-but-ungated fork still read as hardened — the branch name was the only thing distinguishing it. Now the fork carries content, and the name has earned something.

## Verification

- `pmoves/config/fork_registry.json` parses: **84 forks**, entry reads `sync=false`, `branch=PMOVES.AI-Edition-Hardened`
- Placeholder reasons counted before/after: **32 → 31**
- Diff is exactly one line

## Side finding from the parse check, not fixed here

Strict case-insensitive parsers reject this file, and the cause is pre-existing: two key pairs differ **only in case**, i.e. two entries for one repo each.

- `PMOVES-headscale` and `PMOVES-Headscale`
- `pmoves-pipecat` and `PMOVES-Pipecat`

JSON permits it (keys are case-sensitive), so nothing is broken today — but any consumer that folds case sees a duplicate and can silently keep whichever it reads last. Worth deciding which name wins before it matters. Flagging rather than fixing here, since it changes which entry a tool resolves.


POWERFULMOVES:
Thanks — and two responses, one accepted, one pushed back on with reasoning.

**Nit 2 (prose) — accepted, corrected here.** The body said the entry "reads `sync=false`". The field is the JSON boolean `"sync": false`; that was sloppy prose on my part, and this comment is the correction of record.

**Nit 1 (the ~900-char single-line reason) — I'm keeping it, deliberately.** The `reason` field is the decision record: the registry's own doctrine is that a reason is *why* this fork is `sync:true` or `sync:false`, and that operators read it before changing the decision. The length here is carrying load — what the fork is (PMOVES-authored signing-card CLI, not an upstream hardening patch), why `sync:false` follows from that (an auto-merge across `cmd/keygen-cli` is the clobber class the registry exists to prevent), why promoting it widens the App-token scope, and the triage fact that the hardened branch was byte-identical to `master` until the CLI landed. Cut to a display-friendly length it becomes a label rather than a decision, which is the failure mode the placeholder-reason audit was about.

And a fair point in the other direction: a long unbreakable JSON string is genuinely poor for diff tooling. But that's a **rendering** problem, not a content one — the fix belongs in the registry tooling (wrap or render reasons in the audit/status views), not in shortening the decisions until they fit a diff. If the fleet-review lane wants that, I'd take it as a follow-up on `pmoves/tools` rather than as a reason-length rule.

Also noted from the review: the case-collision side finding was left alone on purpose, and the placeholder count claim was checked against a grep counter rather than taken from my word. That's the right way to review it.

One correction to the review's own framing for whoever reads this later: the two case-collision pairs were **fixed** in a later commit on this branch (folded onto the URL-derived keys, per `fork_registry_ratchet`'s own rule that it keys on URLs casefolded), so a re-review of the current head will see 82 forks and zero collisions rather than 84 and two.

_ Lane: PMOVES-4090-ZAI.CLAW · ci/4090zai-fork-gates · self-hosted 4090 _

PR #3129, opened 2026-09-20
fix(ffmpeg-whisper): the CUDA preinstall was gated on aarch64, so x86_64 could not build

POWERFULMOVES:
## Symptom

```
error: Failed to download `torchaudio==2.9.1+cu128`
  cause: Hash mismatch for `torchaudio==2.9.1+cu128`
```

Container never created → `ffmpeg-whisper` unresolvable on the compose network → every `/yt/transcript` and `/yt/ingest` on `pmoves-yt` returns **502**. Measured 2026-09-20: nine ingests, nine 502s.

## The guard protected the fix, not the bug

```dockerfile
if [ "$(uname -m)" = "aarch64" ]; then      # ← the FIX, gated
    uv pip install --index-url .../cu128 torch==2.9.1 torchaudio==2.9.1
fi &&
uv pip install --extra-index-url .../cu128 -r requirements.txt   # ← the CAUSE, ungated
```

`requirements.lock` pins `torchaudio==2.9.1` with **plain-PyPI hashes**.

- **SPARK is aarch64** → preinstall runs → `2.9.1+cu128` satisfies `==2.9.1` as a PEP 440 local version → never re-downloaded, never hash-checked → **builds**
- **Z890 is x86_64** → preinstall skipped, cu128 index still offered → uv resolves the cu128 wheel → checks against PyPI hashes → **mismatch**

*"It runs fine on Spark"* was the **diagnostic**, not a contradiction. The working path was only ever exercised on one architecture.

## Both in-repo reference installers already encode the right rule

Neither gates on architecture — they branch on **GPU vendor** and install the torch trio **explicitly and first**:

| Source | Pattern |
|---|---|
| `pmoves/docs/ARTSTUFF/realtime/torch.js` (Pinokio) | `--index-url .../cu128` — **single index**, never `--extra-index-url` |
| `pmoves/creator/installers/*-AUTO_INSTALL-RUNPOD.sh` | `TORCH_INDEX="…/${CUDA_TAG}"` + explicit trio pins, ahead of the rest |

**Shared rule:** torch from the CUDA index must be installed *from that index alone* and *before* the locked resolve — never blended into a hash-checked resolve via a second index.

**Fix:** run the existing preinstall unconditionally, so x86_64 takes the road SPARK already proves. No lock regeneration, no new mechanism.

**Rejected alternative:** drop `--extra-index-url` on x86_64 so PyPI lock hashes are authoritative. Also builds, but silently changes which CUDA wheels x86_64 nodes get — larger behavioural change, same outcome.

## Verification status — stated precisely

✅ **Proven at the layer that was failing:** the pip install layer now completes — `#21 DONE 247.4s`, torch + whisperx installed, **no hash mismatch**.

⏳ **Not yet asserted:** full container health. Later layers (`chown -R` over a multi-GB site-packages) were still running on this Windows host at commit time. The handoff carries the remaining checklist — including **"SPARK (aarch64) still builds"**, since the branch it relies on is unchanged.

## Related findings — recorded in the handoff, not fixed here

- **`/yt/transcript` has no caption fallback.** YouTube serves captions; `yt-dlp --write-auto-sub --skip-download` retrieved all nine transcripts on this node with **no GPU and no whisper**. Routing straight to whisper turns a zero-byte problem into a 35 MB one.
- **Skill doc drift** (`.claude/skills/pmoves-yt`): `/yt/transcript` takes `video_id`, not `url`; and "no bot-gate on this node" is stale.
- **DHI:** builds `FROM nvidia/cuda:12.6.2-cudnn-runtime-ubuntu22.04` directly, not a hardened base — `DHI_MIGRATION_MANIFEST.md` is the track. Also: base is CUDA **12.6**, wheels are **cu128** (12.8).

Handoff: `pmoves/docs/handoffs/z890-ffmpeg-whisper-arch-gate-2026-09-20.md`
Known Road: `dockerfile:handoff:z890-ffmpeg-whisper-arch-gate-2026-09-20.md`

`python-tests` will fail here for the fleet-wide `test_the_register_is_clean` undeclared-author reason that #3120 fixes.


POWERFULMOVES:
## Verification complete — build finished after the PR was opened

The checklist in the handoff is now closed on Z890 (`x86_64`):

| Check | Result |
|---|---|
| `make -C pmoves up-ffmpeg-whisper` completes | ✅ exit 0, container created |
| container healthy | ✅ `pmoves-ffmpeg-whisper-1` → `healthy` (3 polls) |
| `:8078/healthz` | ✅ `{"ok":true}` HTTP 200 |
| resolvable from `pmoves-pmoves-yt-1` | ✅ `172.30.6.23  ffmpeg-whisper` |
| `/yt/transcript` no longer 502s on "ffmpeg-whisper unreachable" | ✅ **that error is gone** |

**The defect this PR fixes is resolved.** Before: the container did not exist and every call returned `502 ffmpeg-whisper unreachable: Failed to resolve 'ffmpeg-whisper'`.

## The remaining failure is NOT this PR, and not a new defect

`/yt/transcript` now reaches whisper and fails further downstream:

```
ffmpeg-whisper error: transcribe error: An error occurred (404)
when calling the HeadObject operation: Not Found
```

I initially read that as a separate S3/pipeline bug. It is not. `yt_transcript` prefetches `{base_prefix}/audio.m4a` via yt-dlp when the object is absent, and the service log shows why the object was never staged:

```
ERROR: [youtube] MZgAj3OZVPo: Sign in to confirm you're not a bot.
WARNING:pmoves-yt:yt-dlp failed, attempting Invidious fallback chain
WARNING:pmoves-yt:yt_transcript_prefetch_failed
```

So the `HeadObject 404` is **the known YouTube bot gate, one layer downstream** — this node egresses through the `pmoves-kvm4-2` exit node and presents a Hostinger datacenter IP. Not a pipeline defect, and not in scope here.

Corrected rather than left standing, since "new S3 bug" would have sent the next reader chasing the wrong thing.

## Still owed (recorded in the handoff, not fixed here)

- **`yt-cookie-writer` is neither running nor defined in compose.** It is the durable fix for the gate; without it every YouTube path on this node depends on temporarily clearing the exit node.
- **Caption-first fallback for `/yt/transcript`.** YouTube's own captions retrieved all nine target transcripts with no GPU, no whisper and no S3 round-trip. Routing straight to whisper turns a zero-byte problem into a multi-GB one — and, as shown above, into an S3 staging dependency that can fail for reasons unrelated to transcription.

**SPARK (aarch64) still needs confirming** — the branch it relies on is unchanged by this PR, but I cannot build on that node from here.

PR #3139, opened 2026-09-21
fix(security): remove committed NATS credentials from 116 source defaults

POWERFULMOVES:
## What

163 occurrences of a real NATS credential were embedded as fallback defaults in tracked Python source in a **public repo**:

```diff
-NATS_URL = os.environ.get("NATS_URL", "nats://nats:pmoves@nats:4222")
+NATS_URL = os.environ.get("NATS_URL", "nats://nats:4222")
```

Host and port preserved; only the credential dropped. Authenticated deployments already receive `NATS_URL` from the secrets funnel, so the resolved URL is unchanged for any correctly-provisioned service.

## Why it was invisible

The existing guard — `test_bpm_encoder_nats.TestNatsClientNoHardcodedDefault` — checks **exactly one** module's `DEFAULT_NATS_URL`. That module is clean, so the guard reported green over a surface it never looked at: **194 occurrences across 140 files**.

The new guard is **generated from the pattern** over the tracked corpus (1,734 files), so it cannot go stale as files are added. It ships two tests: the guard, and a **corpus-size control** — an empty corpus would otherwise make the guard pass vacuously.

## Remaining exposure is enumerated, not hidden

24 config files stay as **commented, listed exceptions**: compose files, agent profiles, TAC trees, the CHIT manifest. Deleting the credential there would break authenticated connections rather than harden them. They need the opposite fix — an env reference **plus a credential rotation, since this value has been public** — which is operator-sequenced, not mechanical.

Shrinking `KNOWN_CONFIG_EXCEPTIONS` is the follow-up lane.

## Verification

- 489 passed in `pmoves/tests/tools`
- The one remaining failure (`test_google_oauth_flow`) **reproduces on a pristine `origin/main` worktree**; this diff touches no oauth file
- Full `pmoves/tests/` run was blocked locally (reaches shared resources) — CI should cover it


POWERFULMOVES:
## ⚠️ Self-correction — I neutered two security controls in the first commit

The background commit review caught this; I did not. Fixed in `8d88d281c`.

**What went wrong.** The sweep matched any credential-shaped `nats://` URL across `pmoves/**/*.py`. It excluded `pmoves/tests/` but **not** `pmoves/services/*/tests/`, so it rewrote the *inputs* of tests that exist to prove redaction works:

```diff
-    url = "nats://nats:secret123@nats:4222"
+    url = "nats://nats:4222"
     redacted = _redact_url(url)
     assert "secret123" not in redacted     # now trivially true
```

That assertion then passed unconditionally — it would have passed with `_redact_url` defined as `return url`. **A security control became a check that cannot fail, inside a commit whose stated purpose was removing a credential.**

**Fix.** Restored 6 files whose *synthetic* credentials (`secret`, `secret123`, `p4ss`, `s3cr3t`) were altered. Verified by mutation that the controls are real again: feeding each test's own input through a no-op redactor now fails the assertion.

Kept the legitimate changes — fixtures carrying the **real** leaked credential (`nats:pmoves`) that assert on source defaults this lane changed. Those 24 tests pass unchanged.

**Guard hardened** so the same sweep cannot recur: in test files only the known-leaked pair `(nats, pmoves)` is flagged, because a test exercising redaction or URL parsing must contain a credential-shaped string to mean anything. Non-test files keep the broad pattern.

| Suite | Result |
|---|---|
| flute-gateway (the restored controls) | 28 passed |
| `pmoves/tests/tools` | 489 passed |
| `pmoves/tests/unit` + `test_common` | 322 passed |

The single `test_google_oauth_flow` failure reproduces on a pristine `origin/main` worktree and is untouched by this lane.

POWERFULMOVES:
## Broker validation — measured, not inferred

Both my PR body and the review reasoned from `docker-compose.yml`. **Neither measured a broker**, so I measured one.

### First: there is no NATS broker running on B850

| Check | Result |
|---|---|
| Listening on `:4222` | **nothing** — no process, no TCP answer |
| `pmoves-nats-event-bus` | healthy on 8131, but **has no `nats-server` binary** — it is a PMOVES service, not the broker |
| Broker in compose | declared at `docker-compose.yml:3274` (`nats:2.11.8-alpine`, `-js -m 8222 --user …`, publishes `:4222`) — **not started** |
| `pmoves-nats-hub-1` | behind the `nats-hub` profile — not started |

So the review's "the broker runs authenticated → Authorization Violation" and my "the resolved URL is unchanged for any correctly-provisioned service" were **both inferences from config**.

### Measured against an ephemeral authenticated broker

`nats:2.11.8-alpine -js --user probe --pass <throwaway>` on `127.0.0.1:14222`, torn down after. No fleet state touched, no real credential used.

```
credential-free  nats://127.0.0.1:14222            -> nats: error: nats: Authorization Violation
credentialed     nats://probe:<redacted>@…:14222   -> Published 5 bytes to "probe.test"
```

Counted: credential-free → Authorization Violation **1** (expected 1); credentialed → **0** (expected 0).

### What this settles

1. **The review's finding #1 premise is CONFIRMED by measurement.** A host-rewrite that discards userinfo does produce `Authorization Violation` against an authenticated broker. The userinfo-preserving fix in `07d4c34ed` is validated, not just argued.
2. **My PR body overstated its own safety.** "The resolved URL is unchanged for any correctly-provisioned service" holds only for services receiving `NATS_URL` from the funnel. Anything that relied on the **hardcoded default** against an authenticated broker changes from *works (using a committed credential)* to *Authorization Violation*. That is the intended security trade — fail loudly rather than authenticate with a public secret — but it is a behaviour change and I should not have called it unchanged.
3. Compose is already internally inconsistent about this: `:2605` uses `${NATS_URL:-nats://nats:4222}` (credential-free) while `:2889` uses `${NATS_URL:-nats://nats:pmoves@nats:4222}`. This PR makes the source defaults consistent with the former.

### Caveat on method

`nats` CLI returns **rc=0 even on `Authorization Violation`** for `rtt` and `pub`. Exit code is not a usable signal here; the error string is. Any CI gate built on this must assert on output, not on `$?`.

POWERFULMOVES:
## Third blind spot in this guard — it cannot see any submodule

A parallel scout mapping `pmoves_announcer` surfaced a surface this guard still misses. Measured:

| Check | Result |
|---|---|
| `git ls-files '**/*.py'` at superproject | 1,330 files |
| …of those, under `PMOVES-Archon/` | **0** |
| `pmoves-cipher-mcp/pmoves_announcer/__init__.py` tracked by superproject | **no** |
| That file contains `nats://nats:pmoves@` | **yes** |
| `pmoves_announcer` copies carrying the leaked credential | **18** |

`git ls-files` does not descend into submodules, so the corpus this guard calls "the whole tracked repo" excludes **all 71 populated submodules**. Every one of those 18 announcer copies is invisible to it.

### The shape keeps repeating, and that is the actual lesson

1. Original guard — **one module**. Green while 194 occurrences sat in 140 files.
2. This guard, v1 — scoped `pmoves/**`. Green while 48 occurrences sat in 32 files outside it.
3. This guard, v2 — "whole tracked repo". Green while 18 occurrences sit in submodules.

Each iteration widened the boundary and each new boundary was invisible until someone measured *outside* it. The per-glob floor control catches a **broken** glob; it cannot catch a **correctly-working glob pointed at too small a corpus**. A control that validates the corpus it was given can never report that the corpus was the wrong one.

### What I am not doing in this PR

Submodule contents belong to other repos — fixing those 18 files means a PR per fork, not a commit here. What I would change is the guard's honesty: have it **enumerate submodule hits and report them** rather than silently scan past them, so the count is visible even where this repo cannot fix it.

Related, and worth stating plainly: `pmoves_announcer` is vendored into **26+ directories with 9 distinct versions** (124–480 lines), no canonical source, no package. Its `NATS_URL` default is the leaked credential in every copy. Rotation has to reach all of them, and there is no single place to change.

PR #3155, opened 2026-09-23
feat(identity): registered names + voices for the Village Rule bodies

POWERFULMOVES:
Draft. Lane `feat/village-body-identities` (CLAIM in #3154, owner B850-CLAUDE).

Operator, 2026-09-23: "ensure your AGInTZ delivery-agent and code-review have registered names and voices approved for Known Roads."

## In this PR
- **Scope 1 and 2: registry and cards.** `agent_registry.yaml` adds `delivery_agent`, `code_review`, `control_agent` and `memory_agent`, modelled on `node_steward`, with their teams coupled in `agent-teams.yaml`. `signing_identity_cards.yaml` adds H-half-only, **inactive** cards `b850-{delivery-agent,code-review,control-agent,memory-agent}`. The ML half stays empty because issuing keys is operator-only.
- **Scope 3: DONE (5bae26814). Known Roads records which body took a road.** Operator approved landing the guard attribution change on 2026-09-23.
  - Probe on claude 2.1.280 (a throwaway hook dumped its stdin):
    - A subagent's hook input carries `agent_type` and `agent_id`.
    - A main thread started with `--agent` carries `agent_type` only.
    - A plain parent session carries neither.
    - `session_id` is present in all three, while `CLAUDE_SESSION_ID` was unset in the hook env. That is why every existing row reads `session: unknown`.
  - New `known_roads.set_hook_input()` takes the hook's parsed stdin. A trail row records a **registered** `agent_type` as `agent`, using the registry rule that the runtime name is the key with `_` replaced by `-`. The old node value moves to `node`, `agent_instance` comes from the hook `agent_id`, and `session` falls back to the hook `session_id`.
  - An unregistered type goes into `unregistered_agent_type` and is never promoted to `agent`. Rows written outside a hook keep the old key set.
  - **This is attribution, not authentication.** `agent_type` is the agent definition the harness loaded; it is not a credential. Certification is exact-match only. A `strip()` in the first draft let `" delivery-agent\n"` certify, and it was removed. An unreadable registry certifies nothing.
  - Wired into all four callers that can write a row: the bash, edit and write guards and `effect_check.py`.
  - Hooks certify against the registry under `CLAUDE_PROJECT_DIR`. Until this PR merges, the main checkout's registry has no `delivery_agent`, so rows from there still attribute to the node.
  - Tests: `test_known_roads_attribution.py` runs 88 checks with the trail redirected to a temp file.
    - Control: HEAD with a no-op stub fails 38/88, all behavioural and including all 4 wiring checks.
    - Control: the pre-fix `strip()` version fails 6/88.
    - Guard suite: 11/11 before, 12/12 after.
    - Static 43/0 and smoke 29/0 in both runs.
    - `known-roads.jsonl` stayed at 251 → 251 rows in every run.
    - `run_integration.sh` was not run: it does a live NATS publish round-trip.
- **Scope 4: MEASURED by the steward.** `claude auto-mode config` shows that **0 of 12** custom rules in `settings.local.json` take effect. Rules in user settings do.
- **Review R1-R5 (control review, APPROVE-WITH-NITS at 5bae26814), all addressed:**
  - R1 (5317824ff): the trail row schema is documented in `.claude/skills/known-roads/SKILL.md` under "The trail row". It lists every field and when it appears, and says to group by node with `node if present else agent`. It also notes that `AGENT_ID` is ambiguous: tools use it for agent names too, and whatever it holds is recorded as the node.
  - R2 (40095a068): actor fields are now built inside the fail-closed `try`, and any exception makes `_record` return False, so the grant is denied. 4 new checks, 92 in total. Against 5bae26814 the exception escaped `evaluate_known_road`.
  - R3 (bdcb8ffa9): the cards header now says the schema's `h.role` enum is not enforced on this file, since four existing roles already fall outside it.
  - R4 (no change needed): measured on the merge-base e0945242e and on this branch with `CHIT_REQUIRE_SIGNATURE` unset.
    - `audit_naming_drift.py`: 198 findings (182 P1, 16 P0) on both trees, with 0 new findings for the added cards. The finding sets differ only in the checkout path. Cards (active/total) go from 26/26 to 26/30.
    - `build_allowed_signers.py --check`: in sync, 3 trusted keys on both trees.
    - `test_audit_naming_drift.py`, `test_build_allowed_signers.py`, `test_chit_forks_cards.py`: 19 passed and 2 failed on BOTH trees. The failures are the same stale pins (active 23 vs 26, PEM misuse 9 vs 13) and already exist on main.
    - p7 room-orchestrator: 51 of 51 passed on both trees.
  - R5: see the claim scope note below.
- **CLAIM scope vs this diff.** Claim item (3), identity_vocabulary aliases, was **not needed**: `pmoves/tests/test_identity_lineage.py` passes 64/64 without them. Claim item (5), the auto-mode rules template, was **dropped** because the operator applied the settings directly.
- **Scope 5: DROPPED from this lane.** The autoMode rules template and merge tool are no longer planned here, because the operator is updating their settings directly.


POWERFULMOVES:
## Independent control review @ `5bae26814`: APPROVE-WITH-NITS (no P1)

**Attribution only, confirmed.** No allow/deny decision reads `_HOOK_INPUT`, `agent_type`, `agent_id` or the registry; `set_hook_input` cannot raise; the hook call sits after a successful `json.load` and before any early exit in all 4 hooks; only top-level envelope keys are read, never `tool_input`. Guard suite 12/12 (88/88 attribution checks), static 43/43, live trail 251 -> 251 rows. Negative control: restoring the old actor logic fails 34/88, plus 4 wiring checks = 38, matching the author. The registry validator passes 111/111. The registry parse (~77ms) is cached and only runs when a row is written.

| # | Sev | Where | Finding |
|---|---|---|---|
| R1 | P2 | `known_roads.py:324-326` | `agent` now mixes node ids (old and main-thread rows) with body names (subagent rows), and `node` takes `AGENT_ID` first, so a harness that sets `AGENT_ID` to an agent name records it as `node`. No reader breaks today (`trail_states.py`, `roads.py`, `sweep_differential.py`, `effect_check.py` and the annotations never read `agent`), but the row format is documented nowhere. Document the row schema (agent/node/agent_instance/unregistered_agent_type semantics) next to the trail. |
| R2 | P3 | `known_roads.py:356` | `entry.update(_actor_fields())` sits outside the `try`/`except OSError`, so "returns False if it could not be written" now depends on `_actor_fields` never raising. A future raise there would crash a PreToolUse hook, which is non-blocking, so the fail-closed design would fail OPEN. Move it inside the try, or catch broadly and return False. |
| R3 | P3 | `signing_identity_cards.yaml` header comment | It claims the schema enumerates only operator/agent/runner/service-account, but the file already has `evaluator`, `ci-bot`, `community_steward` and `creator_steward`. Correct the comment. |
| R4 | P3 | cards | These are the first `active:false` cards. Not run against them: `audit_naming_drift.py`, `test_build_allowed_signers.py`, `test_chit_forks_cards.py`, the p7 orchestrator tests. Run them. |
| R5 | P3 | CLAIM scope | Claim items (3) identity_vocabulary aliases and (5) the auto-mode template are not in the diff. (3) was not needed (`test_identity_lineage` passes); (5) was dropped (settings applied directly by the operator). State both in the PR body. |

Steward disposition: fix R1-R5 on this PR, then merge.

POWERFULMOVES:
R1-R5 are addressed on `feat/village-body-identities`. The new head is `bdcb8ffa9`.

| # | Commit | What changed |
|---|---|---|
| R1 | `5317824ff` | Adds "The trail row" to `.claude/skills/known-roads/SKILL.md`. It gives each field's presence rule and meaning (`agent` / `node` / `agent_instance` / `unregistered_agent_type` / `session` / `note`), says to group by node with `node if present else agent`, and states that this is attribution, not authentication. **`AGENT_ID` ambiguity:** all 251 existing rows hold node ids in `agent`, but `pr_hedge_trim.py` defaults `AGENT_ID` to `claude-code-cli` and `persona-bind/bind.sh` to `4090-claude`, so an agent name exported there would be recorded as the node. This is documented, and `_actor_fields` points to the section. |
| R2 | `40095a068` | `entry.update(_actor_fields())` and the note now sit inside the `try`, and `except Exception` returns False. Four new checks (92 in total) patch `_actor_fields` to raise. They assert that `_record` returns False without raising, that no row is written, and that a **valid** `compose:pr:3155` grant is denied with "could not be recorded". Control: with 5bae26814's `known_roads.py`, the exception escapes `evaluate_known_road` (traceback). |
| R3 | `bdcb8ffa9` | The header was wrong in spirit. The schema's `h.role` enum really is those four values, but it is not enforced on this file, which already holds `evaluator` / `ci-bot` / `community_steward` / `creator_steward`. The comment now says so, and explains that `agent` was chosen because it is in the enum and describes these bodies. |
| R4 | none needed | Run on the merge-base `e0945242e` (throwaway worktree, since removed) and on this branch, with `CHIT_REQUIRE_SIGNATURE` unset in both runs. `audit_naming_drift.py` (jsonschema 4.26.0 installed, so the schema was validated): 198 findings on both, 0 new; the only diffs are the checkout path inside 3 compose-parse messages. Cards (active/total) go from 26/26 to 26/30. `build_allowed_signers.py --check`: OK, 3 trusted keys on both. `test_audit_naming_drift` + `test_build_allowed_signers` + `test_chit_forks_cards`: 19 passed, 2 failed on **both** trees. The two failures are the same stale pins on main (`cards_summary.active` 23 vs 26, PEM misuse 9 vs 13), so the new cards did not cause them and I did not change them here. p7 room-orchestrator: 51/51 on both. |
| R5 | PR body | The body now states that claim item (3) was not needed (`test_identity_lineage.py` 64/64) and that item (5) was dropped (the operator applied the settings directly). |

**Suites after the fixes, same denominators as before:**
- `run_guard_tests.sh`: 12/12 files; the attribution suite passes 92/92.
- `run_static.sh`: 43 pass, 0 fail.
- `run_smoke.sh`: 29 pass, 0 fail.
- `known-roads.jsonl`: 251 rows before and after.

PR #3157, opened 2026-09-23
fix(ci): kilo-review resolves its model against the live catalog (dead id was a missing kilo/ prefix)

POWERFULMOVES:
## Summary

Every `kilo-review` run since 2026-09-21T01:19Z fails with `Error: Model not found: z-ai/glm-4.5.` (31/31 runs, e.g. 35885122102). Lane claim: B850-CLAUDE, `fix/kilo-review-model-live` (CLAIM in #3156). Operator approved this on 2026-09-23.

### Root cause (measured)
`z-ai/glm-4.5` is still a live model. The CLI just doesn't accept that form of the id:
- Kilo gateway catalog (`api.kilo.ai/api/gateway/models`, public): 393 ids, includes `z-ai/glm-4.5`.
- CLI catalog (`kilo models kilo`, `@kilocode/cli@7.6.2` in `node:22-slim`, no credentials): 324 ids, **all prefixed `kilo/`**. It includes `kilo/z-ai/glm-4.5`.
- CLI ids take the form `<provider>/<gateway-id>`, so a bare `z-ai/glm-4.5` parses as provider `z-ai`, and no such provider exists. #3080 copied the ids from the CLI's "Did you mean" list, and that list prints gateway ids without the prefix.
- The last two green runs used prefixed ids: `kilo/ollama-cloud/glm-5.2` (35550286778) and `kilo/z-ai/glm-5.3-flash` (35520497983).

### Change (`.github/workflows/kilocode-review.yml` only)
- **Validate before reviewing.** Inside the same container, with the same key, the workflow runs `kilo models kilo` and resolves the model against that catalog:
  - Dead override: `::error::` naming the model, a list of valid ids, exit 1.
  - Empty catalog: exit 3 (could-not-measure).
  - Prefix-less override that exists as `kilo/<id>`: repaired, with a `::warning::`.
- **No single hardcoded model.** When `KILO_REVIEW_MODEL` is unset, the job uses the first catalog hit from an ordered preference list. You can override that list with the `KILO_REVIEW_MODEL_PREFERENCES` repo variable. The `KILO_REVIEW_MODEL` override is kept.
- **The comment header names the model that was actually used** (taken from a step output). Before, it read `${KILO_REVIEW_MODEL:-...}` in a step that never had that variable in its env, so it always printed the fallback.
- **Exit 0 with no review output is now a failure**, not a pass.

### Verification
- Resolver tested in a clean `node:22-slim` container against the live catalog, 5 cases: unprefixed override repaired, valid override, dead override (exit 1 + suggestions), unset (preference list), all preferences dead (exit 1).
- shellcheck: clean on the in-container script.
- actionlint: only the 2 findings that are already on main (`kvm4` custom label, SC2024 on the intentional host-side redirect).
- YAML parses.
- Live proof: the kilo-review run on this PR (results in comments below).

### Operator follow-up done alongside
`KILO_REVIEW_MODEL` repo variable moved from `z-ai/glm-4.5` to `kilo/z-ai/glm-5.2`, which is in the measured catalog (see comments for readback).


POWERFULMOVES:
## Live proof and the second Kilo lane

**The `kilo-review` Actions lane is fixed. Two runs on this PR both passed and both posted a real review:**

| run | `KILO_REVIEW_MODEL` | resolved model | outcome |
|---|---|---|---|
| 35887505421 | `z-ai/glm-4.5` (old, dead value) | `kilo/z-ai/glm-4.5` (prefix repaired, `##[warning]` emitted) | success, review posted (201) |
| 35888086904 | `kilo/z-ai/glm-5.2` (new) | `kilo/z-ai/glm-5.2` (validated) | success, review posted (201) |

Run 1 proves the root cause. It used the same variable that failed 31 times in a row, and the only difference was the `kilo/` provider prefix. In CI the authenticated catalog lists 352 `kilo/*` ids; without credentials it lists 324.

Repo variable: `z-ai/glm-4.5` (updated 2026-09-21T01:19:58Z) became `kilo/z-ai/glm-5.2` (updated 2026-09-23T16:20:19Z), confirmed by reading back `actions/variables/KILO_REVIEW_MODEL`.

**Not fixed by this PR, and a repo change can't fix it:** the CANCELLED check is a *different* lane, the **"Kilo Code Review"** check from the `kilo-code-bot` GitHub App. It is still cancelled on this PR's head commit, with this output:

> Selected model is no longer available. The review did not run because the selected model is no longer available. Choose another model in Kilo Code review settings: https://app.kilo.ai/code-reviews

That model is set in the Kilo web dashboard, not in any file or repo variable. An operator needs to pick a live model there (for example GLM 5.2, the same model this workflow now uses).

PR #3158, opened 2026-09-23
fix(branch-protection-sync): renamed PMOVES-N8N broke the token mint; name drift pre-mint; n8n roles reconciled

POWERFULMOVES:
## Why
`branch-protection-sync.yml` has failed since 2026-09-20 (runs 35503913554, 35599311111) at `actions/create-github-app-token`: "at least one repository that does not exist or is not accessible to the parent installation". The workflow mints one token for the 80 repos it derives from `.gitmodules`. `POWERFULMOVES/PMOVES-N8N-Auto` was renamed to `PMOVES-N8N`. REST follows the redirect; the token mint does not. One stale name stopped the audit for all 80 forks.

## n8n repo roles (operator clarification 2026-09-23)
- `POWERFULMOVES/PMOVES-N8N` (was PMOVES-N8N-Auto, submodule path `PMOVES-N8N-Auto`): the **full fork of n8n-io/n8n**. This is the repo that syncs from upstream.
- `POWERFULMOVES/PMOVES-n8n-FlooS` (submodule path `PMOVES-n8n`): **PMOVES workflows, first-party**, no upstream. It also carries a thin `compose/n8n/Dockerfile` (`FROM n8nio/n8n:2.1.0`), which compose builds.
- GitHub names are case-insensitive, so `PMOVES-n8n` resolves to the **same repo** as `PMOVES-N8N` (id 1356127122). Several references used that URL for the workflows submodule and so linked to the fork.

| Location | Before | After |
|---|---|---|
| `.gitmodules` [PMOVES-N8N-Auto] url | `PMOVES-N8N-Auto.git` (renamed, mint fails) | `PMOVES-N8N.git` (name/path unchanged; pin 44ddbbd28c exists and is an ancestor of Hardened) |
| `.gitmodules` [PMOVES-n8n] url | `PMOVES-n8n-FlooS.git` | unchanged (pin 92267e98 exists in FlooS) |
| `fork_registry.json` forks | key `PMOVES-N8N-Auto`, sync:false "deferred" | key `PMOVES-N8N`, sync:true, rename + roles recorded in `reason` |
| `fork_registry.json` `_first_party` | `PMOVES-n8n` (now resolves to the fork) | `PMOVES-n8n-FlooS` (also clears main's current ratchet failure; see #3123) |
| `fork-sync.yml` token scope + FORKS | `PMOVES-n8n` (the fork via case alias; run 35850145236 "SYNC PMOVES-n8n: 147 behind") | `PMOVES-N8N` explicit; FlooS was never in the list |
| `.claude/context/submodules.md` | PMOVES-n8n -> `PMOVES-n8n.git`, branch main | -> `PMOVES-n8n-FlooS.git`, branch Hardened |
| `docs/PMOVES_Repository_Index.md` | link `PMOVES-n8n.git` | link `PMOVES-n8n-FlooS.git` + role note |
| `tac_trees/node-hermes-agent.tac.yaml` | `git: PMOVES-n8n.git` | `git: PMOVES-n8n-FlooS.git` |
| `HERMES_AGENT_INTEGRATION.md` | `PMOVES-n8n` | `PMOVES-n8n-FlooS` |
| `PMOVES_SERVICE_SKILLS.md` | "PMOVES-n8n = RUNTIME source fork / N8N-Auto = mirror" (roles inverted) | roles corrected |
| `N8N_SETUP.md` | "PMOVES-n8n ... runtime/workflow fork" | workflow repo + runtime wrapper; the fork is PMOVES-N8N |

**Already right, unchanged:** compose (`docker-compose.n8n.yml`, `compose/docker-compose.core.yml`), which builds FlooS's wrapper and mounts its `workflows/`; the Makefile `N8N_SUBMODULE_DIR`; `submodule-update-check.yml` and `submodule-smoke.yml` (path-based); `HARDENED_BRANCH_RECONCILIATION.md` ("upstream n8n" base image); `AUTOMATION_LANDSCAPE_2026-07-22.md`; the n8n scripts and migration defaults (path `PMOVES-n8n/workflows`). **Not rewritten:** dated evidence/audit/research snapshots.

## Hardening (`branch-protection-sync.yml`)
- **Pre-mint resolution.** An installation-wide `metadata:read` token resolves every derived name via REST. Each renamed or missing repo is **named** as an `::error` and in the step summary, and then the job fails. Case-only differences are notices; fork-sync shows the mint accepts them.
- **Fail rather than substitute.** `.gitmodules` is also the clone source, so substituting would hide a url that is stale everywhere else. A redirect is not a stable identity (`PMOVES-n8n` was re-claimed by the renamed fork). This job holds admin:write, so it should write only to repos a reviewed `.gitmodules` names.
- **Dry-run evidence.** For each repo, the dry run shows current vs intended `required_approving_review_count` and the exact fields a PUT would change, including fields the PUT resets to false, so downgrades are visible. It also counts would_change / no_change / unreadable.
- A protection GET that fails for a reason other than "Branch not protected" or needs-Pro is now reported as **UNREADABLE**. Before, it was assumed UNPROTECTED.
- Dispatch inputs are unchanged. actionlint shows the same 3 pre-existing notes before and after.

## Validation
- fork-registry ratchet 84/84 with coverage clean, rc=0. The same check on an origin/main copy gives rc=1 on `PMOVES-n8n-FlooS`.
- 140 passed across the fork-registry, fork-guards, branch-parity, branch-protection and submodule-freshness suites. The gitlink gate PASSED, but that is vacuous: no gitlink changed.
- The dry run from this branch is reported in a comment below.

## Heads-up
- **4090 lane** (`ci/4090zai-fork-gates`, PMOVES-4090-ZAI.CLAW): PR #3115 did the earlier n8n url fix. You may want to review the fork-sync change and the registry `sync:true`.
- #3123 adds `PMOVES-n8n-FlooS` to `_first_party` in the same hunk. Resolve the conflict by keeping FlooS and **dropping** `PMOVES-n8n`.
- `.claude/skills/pmoves-submodule-fleet/scripts/fleet_audit.sh` matches registry keys against submodule **paths**. Its advisory "skip" flag no longer matches path `PMOVES-N8N-Auto`. This is display-only and is not fixed here.
- The SITREP's "still owed: fork-sync calls FlooS a fork" can be closed. fork-sync always targeted the fork; the name just read like the FlooS path.

Lane: `fix/branch-protection-sync-rename` (B850-CLAUDE, CLAIM in #3156). The live (`dry_run=false`) dispatch is left to the steward after the operator reviews the dry run.


POWERFULMOVES:
## Dry run from this branch: run 35888036686 (success)
`dry_run=true required_reviews=0 only_unprotected=false`. Nothing was written.

- **Token mint: OK.** The new pre-mint resolution step reported `resolved=80 drifted=0`, with no case-only notices. The scoped admin mint succeeded for all 80 repos.
- **Summary:** `would_change=75 no_change=5 skipped=0 unreadable=0`

| group | n | reviews current -> intended | fields a PUT would change |
|---|---|---|---|
| protected, count=1 | 40 | **1 -> 0** | `reviews.required_approving_review_count`, `reviews.require_last_push_approval` (true -> false) |
| unprotected | 34 | none -> 0 | all (new protection) |
| protected, PR not required | 1 (PMOVES-crush) | none -> 0 | `required_pull_request_reviews`, `required_conversation_resolution` |
| protected, count=0 already | 5 | 0 -> 0 | no change (OpenRoom, agent-sandbox-skill, awesome-agent-skills, claude-d3js-skill, fork-repository-skill) |

**The 40 that a live run with `only_unprotected=false` would DOWNGRADE (1 -> 0 approvals, last-push approval off):** A2UI, Agent-Zero, AgentGym, Archon, BoTZ, BotZ-gateway, Creator, Deep-Serch, DoX, E2B-Danger-Room-Desktop, E2b-Spells, HiRAG, Jellyfin, Open-Notebook, Pinokio-Ultimate-TTS-Studio, Remote-View, Tailscale, ToKenism-Multi, Ultimate-TTS-Studio, Wealth, a0-plugins, autoresearch, llama-throughput-lab, n8n-FlooS, neo4j, pinokio, space-agent, supabase, surf, tensorzero, transcribe-and-fetch, PMOVES.YT, Pmoves-AgentGym-RL, Pmoves-Health-wger, Pmoves-Jellyfin-AI-Media-Stack, Pmoves-cipher, Pmoves-hyperdimensions, Pmoves-pretext, pmoves-cipher-mcp, pmoves-e2b-mcp-server. All are on `PMOVES.AI-Edition-Hardened`.

The 34 unprotected include **PMOVES-N8N@PMOVES.AI-Edition-Hardened**, the repo whose rename caused the outage.

Operator decision: `only_unprotected=true` would protect the 34 gaps and leave the 40 at count=1. `only_unprotected=false` (as dry-run here) would lower those 40 to 0.

POWERFULMOVES:
## Response to the Kilo review (REQUEST_CHANGES), point by point. Fixes are in `c943df11b`.

**1. Is forks.tsv derived from the url or from the submodule name? Url, and this is proven.** The derive step reads `git config --get-regexp '^submodule\..*\.url$'` and takes `repo="${url##*/}"`, so the submodule name is never used as the repo. Evidence: dry run 35888036686, taken after the url change, logged `resolved=80 drifted=0`. If the name `PMOVES-N8N-Auto` had been used, it would have shown up as RENAMED and the job would have failed.

**2. set -e / pipefail on the new jq calls. Fixed. Your diagnosis was slightly off, and the real failure was worse.** The apply step starts with `set +e`, so a jq failure would not have aborted it. It would have **failed silently**: an empty `now` gives an empty diff, which the dry run reported as "no change". In a live run, a non-JSON 200 response gave `rsc=null`, and the PUT would have **nulled out that branch's existing required checks**. Now every parse is guarded (type check, state fields, preserved checks, `normalize`, `intended`, diff). Any failure marks the branch UNREADABLE and it is never written.

**3. UNREADABLE could leave the audit green. Fixed: it is now loud.** A dry run emits `::warning`. A live run emits `::error ... (not written)`. Each case is logged, gets a table row, and is counted in the summary. A live run with any UNREADABLE exits **3** (could-not-measure). A dry run exits 0, so the push/schedule audit still annotates but does not go red on a transient 5xx. Failed PUTs now also emit `::error`.

I tested this with a harness: the step extracted from the YAML, run under `bash -eo pipefail`, with `gh` stubbed to return a good response, an identical one, non-JSON, a JSON array, unprotected, 403 with a `|` in the message, and 404 branch-missing.
- dry run: rc=0, `would_change=2 no_change=1 unreadable=4`, 0 PUTs
- live run: rc=3, `applied=3 unreadable=4`; PUTs went only to the 3 readable branches
- clean inputs: rc=0 in both modes

**4. Diff asymmetry. Fixed, and the key lists are shown below.** The diff now iterates the **union** of keys from both sides, at the top level and inside the review object.
- `normalize` keys: allow_deletions, allow_force_pushes, block_creations, enforce_admins, lock_branch, required_conversation_resolution, required_linear_history, required_pull_request_reviews, required_status_checks, restrictions
- `intended` keys: the same 10. `policy()` emits 8 of them; `intended` adds lock_branch and block_creations, which a PUT resets to false.
- review sub-keys on both sides: dismiss_stale_reviews, require_code_owner_reviews, require_last_push_approval, required_approving_review_count
- Controls: a key present only in `now` is reported (`[only_now]`), and so is a key present only in `want` (`[only_want]`).

**5. `|` in table cells. Fixed.** An `md()` helper escapes it. The harness row reads `Resource not accessible \| by integration`.

**6. Node 24 for the pinned create-github-app-token. Proven.** Runs 35888036686 and 35893583307 both minted on `ubuntu-latest` (GitHub-hosted, not self-hosted).

**7. Registry `sync:false -> true`. I agree this is a behaviour decision, not just a rename.** The operator ruled on it (2026-09-23): PMOVES-N8N is the fork that syncs from upstream. fork-sync.yml already syncs it (run 35850145236: `SYNC PMOVES-n8n: 147 behind`, where `PMOVES-n8n` is the same repo by case), so the token scope does not change.

**8. A header row in forks.tsv.** There isn't one: the file is built with `: > forks.tsv` and then printf rows. The loop shares that assumption, so no change.

**Deferred follow-up (known residual, not fixed here).** The pre-mint check only proves that a repo exists under `metadata:read`. A repo installed **read-only** (without administration:write) passes that check, and the scoped mint then fails with the same opaque "not accessible" message. Closing that needs a per-repo permission probe, e.g. `GET /repos/{r}/installation` compared against the installation's granted permissions. I'm filing it as a follow-up and not widening this PR.

A new dry run from `c943df11b` has been dispatched (35893583307). Results will follow in a comment.

POWERFULMOVES:
## Dry run 35893901820 from head `d699269f7`: success, totals match

`resolved=80 drifted=0` · token mint OK · **`would_change=75 no_change=5 skipped=0 unreadable=0`** · no warnings.
All 80 per-repo rows are **identical** to dry run 35888036686 (diffed line by line), so the downgrade table in the earlier comment still holds.

**My own guard misfired first (35893583307, on `c943df11b`).** It marked 5 readable branches UNREADABLE: crush, agent-sandbox-skill, awesome-agent-skills, claude-d3js-skill and fork-repository-skill. None of them has required status checks. The preserved-checks expression correctly returns `null` for them, but `jq -e` treats a `null` result as failure. The new loudness made this show up as 5 `##[warning]`s, not as a silent "no change". `d699269f7` removes `-e` from that one call. It still fails on invalid JSON and on an empty result. The harness now includes the real crush and sandbox protection responses.

Live run not dispatched; that is left for the steward after operator review.

POWERFULMOVES:
## Response to Kilo re-review @ `d699269f7`. All three fixed in **`f0d37474b`**.

| Kilo point | Fix in `f0d37474b` | Harness evidence (`bash -eo pipefail`, stubbed gh) |
|---|---|---|
| **1. Zero-resolved guard.** A missing or empty `forks.tsv` passed vacuously: the process-substitution failure is invisible to `-e`, the loop ran zero times, and `bad` stayed 0. | The input is counted before the loop. The step fails if the file is missing or empty (`[ -s ]`), and it requires **input > 0 and resolved == input**, naming both counts. The summary now prints `input= resolved= drifted=`. | normal (2 rows) rc=0 · **empty rc=1 · missing rc=1 · whitespace-only rc=1** (each names input=0 resolved=0) · renamed rc=1 · case-only rc=0 |
| **2. Live exit code for FAILED PUTs.** A live run with failed PUTs exited 0 (green) while leaving branches unprotected. | Precedence for live runs: **any FAILED PUT → 1** (with an `::error` naming the count), **otherwise any UNREADABLE → 3**, otherwise 0. FAILED outranks UNREADABLE because it is a measured failure; UNREADABLE is an unmeasured one. Dry runs always exit 0. | live: clean rc=0 · **failed-PUT-only rc=1** · **unreadable-only rc=3** · **failed + unreadable rc=1** · dry run with mixed inputs rc=0, 0 PUTs |
| **3. Pre-mint `msg` was unbounded.** | CR and LF are stripped and the message is capped at 160 chars, matching the main loop. | 404 error with CR and a second line → one annotation line, no CR · 300-char error → msg is exactly 160 chars |

actionlint: the same 2 pre-existing findings (`read` without `-r` in the derive step, and the PUT's `$?` check); neither is from new code.

**Dry run 35895611322 from `f0d37474b`: success.** `input=80 resolved=80 drifted=0` · token mint OK · **`would_change=75 no_change=5 skipped=0 unreadable=0`**. All 80 per-repo rows are identical to 35888036686.

**Other points from the re-review, for the record:**
- `policy()` does carry `dismiss_stale_reviews: true` and `required_approving_review_count`, so there is no spurious `reviews.dismiss_stale_reviews` diff. The 5 no-change rows show this.
- `set +e` is the first line of the apply step, so it covers the whole loop, including the new `head | cut` pipes.
- The `jq -ec` rsc regression that Kilo describes was already fixed in `d699269f7`.
- Not addressed, noted only: the literal "null" in the state string when a sub-object is null is cosmetic. The doc says `FROM n8nio/n8n` without a tag, while the Dockerfile pins it; I left the doc tag-free on purpose so it cannot drift. The pre-existing public ports and schema names are out of scope.
- The read-only-installation residual is still deferred as a follow-up.

Live run not dispatched.

POWERFULMOVES:
Steward, on the pipefail confirm-item: **confirmed safe.** `policy()` is a single `jq -cn --argjson rsc ... --argjson rev ...`. jq fails on an invalid `--argjson` BEFORE emitting output, so `policy` cannot exit non-zero while emitting valid JSON. An empty body makes the PUT fail, which is correctly logged as FAILED. `$rsc` is guarded by the round-1 parse checks, and `REQUIRED_REVIEWS` is the dispatch input `0`. The other items are acknowledged as nits or follow-ups: the not-installed-but-public repo probe, and the `intended()` rpr-from-null dry-run fidelity. Note for the operator: the policy sets `required_conversation_resolution: true`.

